PR #1, opened 2015-03-29
Realtime client

yavor87:
Realtime client initial version. Implemented features:
* Websocket communication with json
* Attach, detach, publish and subscribe for channels


yavor87:
Indeed, removing channels before detached sight is an oversight. I'll fix it.

yavor87:
About the callbacks, I wanted to implement an api that is natural to .net developers. Receiving result (and acting upon it) is implemented in .net 4.0+ with Tasks API. If this is an important feature I can add an override with a predicate of type Action<T>. This way we can have an api like:

```csharp
channel.Publish("event", null);

// With callback
channel.Publish("event", null, (result) => { Console.Writeline("Message received" });

// Or (.net40 +)
Task<PublishResult> result = channel.PublishAsync("event", null);
result.ContinueWith((antecedent) => { Console.Writeline("Message received" } })
```

yavor87:
The point I wanted to make is, that in .NET there are currently [3 patterns](https://msdn.microsoft.com/en-us/library/jj152938(v=vs.110).aspx) for async methods and I wanted to conform to these. I feel that this style of callbacks is more like javascript programming, which is not a bad thing, but a little bit alien for .net. There are some examples of such an api in C# [here](https://socketio4net.codeplex.com/).

Martin, what do you think? @marto83

mattheworiordan:
Wow, @yavor87, this is awesome work!  I am very impressed.

marto83:
The thrift references are because the github repo is way behind bitbucket. I will push my code to github this evening which will fix some of the issues.

marto83:
Regarding the the callbacks I think we should have both versions - the one that takes an Action and the Async one that returns a Task. 
This way people should be able to use it both ways.

mattheworiordan:
@paddybyers can you review commit 6991f7c as we need to decide which approach we will be taking for the .NET library.

PR #16, opened 2015-09-25
Realtime connection management redone

yavor87:
I have completely redone the realtime connection management. Connection state machine is implemented using the state design pattern. Every connection state is a separate class and the connection manager coordinates the state, the transport and stuff like message acknowledgement.

Please check the affected files in the final version, because things can vary between commits.


nuget/ably.io.0.7.0.nuspec
```diff
@@ -0,0 +1,34 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+    <metadata>
+        <id>ably.io</id>
+        <version>0.7.0</version>
+        <title>ably.io</title>
+        <authors>Yavor Ivanov,Martin Georgiev</authors>
+        <owners>Matthew O'Riordan</owners>
```
mattheworiordan: Maybe make he owners `Ably Real-time Ltd`

mattheworiordan:
Wow, this looks really good.  You have taken an interesting approach with your Connection state classes having the logic for each state in there, it's a nice.  Good work @yavor87

mattheworiordan:
:+1:

PR #24, opened 2015-12-03
Fixed #21. Forced TokenDetails dates to be serialized as milliseconds…

### LINKED ISSUE #21: TokenDetails serialisation
@yavor87 a customer has today been using the .NET library and used it to generate a token that is sent to the client.  However, we have a serialisation issue and a clientId issue.

Here is a sample of the serialised TokenDetails object:

```
JsonConvert.SerializeObject(tokenDetails) => 
{"token":"REMOVED","keyName":"REMOVED","expires":1449148030,"issued":1449148026,"capability":"{ \"*\": [ \"*\" ] }","clientId":"undefined"}
```

So the obvious problem here is that the serialised expires & issued values should be in milliseconds not seconds since the epoch (unix time).

Also, "clientId" is "unidentified" because sending an empty string clientId failed.  I expect that is due to the fact that a clientId cannot contain an empty string, and must support null values, which when serialised should be omitted.

See https://github.com/ably/docs/commit/145a0f7abe845b43b1bb37d8af891ead39142102 and https://github.com/ably/docs/commit/1ca0fe7504e37590ac689a5b4311e36cf138bb09.

Actions needed ASAP;

- [x] Ensure a serialised TokenDetails and TokenRequest object has the correct timestamp value in milliseconds, plus test coverage
- [x] Ensure a serialised TokenDetails and TokenRequest object supports a null value clientId, which is the default value.  When the value is null, the key and value should be omitted from the serialised object i.e. `{ "token": "x", "expires": 999, "issued:" 999 }`, plus test coverage
- [x] Ensure a serialised TokenDetails and TokenRequest object supports a string value clientId, plus test coverage

marto83:
… instead of seconds. Removed cliendId from json if empty


src/Ably.Tests/TokenDetailsTests.cs
```diff
@@ -44,5 +46,28 @@ public void FromJson_ParsesTokenCorrectly()
...
+            var details = new TokenDetails()
+            {
+                Expires = DateTimeOffset.Now,
+                Issued = DateTimeOffset.Now.AddSeconds(1),
+            };
+
+            var json = JsonConvert.SerializeObject(details);
+
+            var jobject = JObject.Parse(json);
+            ((string) jobject["expires"]).Should().Be(details.Expires.ToUnixTimeInMilliseconds().ToString());
```
mattheworiordan: Should you not test issued as well?

src/Ably/Auth/TokenDetails.cs
```diff
@@ -48,7 +51,8 @@ public sealed class TokenDetails
         /// <summary>
         /// The clientId associated with the token
         /// </summary>
-        [JsonProperty("clientId")]
+        [JsonProperty("clientId", NullValueHandling = NullValueHandling.Ignore)]
+
```
mattheworiordan: Looks good, do you have any idea how MsgPack will handle this? Is it worth having a test to confirm?

mattheworiordan:
:+1:

src/Ably.Tests/TokenDetailsTests.cs
marto83: Good point. I'll update the test.

src/Ably/Auth/TokenDetails.cs
marto83: I need to investigate how the messagepack serialiser works

marto83:
Now all conversions from Dates to unix time will be done in milliseconds.

mattheworiordan:
Nice work mate! :+1: 
Thanks so much @marto83

PR #30, opened 2015-12-31
Xamarin

yavor87:
Adding support for Xamarin.IOS, Xamarin.Android.


mattheworiordan:
LGTM, thanks :+1:

yavor87:
I revised the work that I have done previously on the Xamarin support. Last time I made a lot of changes to the Ably project. This time I only added Xamain support without modifying the Ably project. So everything that is working in master branch should be working here as well.

mattheworiordan:
> I revised the work that I have done previously on the Xamarin support. Last time I made a lot of changes to the Ably project.

Ok, makes sense for this PR.  What should we do with the bunch of commits you had though as it seems like you fixed / improved things quite a bit and it would be a waste to lose that work.

yavor87:
In the previous PR I started with the idea to replace the current method for REST communications with RestSharp. The idea was that RestSharp is cross-platform and supports wide range of technologies. But while doing this refactoring a lot of things went ugly. Finally I come to the conclusion, that for Xamarin support no major changes to the Ably project are needed, because Xamarin supports all the things we need. So RestSharp is not needed and all the things I have done in the previous branch were fruitless. I started from scratch here and as you see, the results are much better.

mattheworiordan:
> I started from scratch here and as you see, the results are much better.

Ah, great, I did not realise.  Thanks @yavor87

mattheworiordan:
@yavor87 I merged this in, but I see the builds are breaking long before they even get to run, see https://travis-ci.org/ably/ably-dotnet/builds/99849811.  Any idea what's changed that is causing the builds not to run now?

yavor87:
As far as I can see from the log, travis can't find targets for Xamarin IOS
and Android. Its highly possible that travis doesn't support building
xamarin. More research is needed.

mattheworiordan:
Ok, I have raised an issue for it at https://github.com/ably/ably-dotnet/issues/32.  It seems it is possible, but I am not sure

PR #31, opened 2016-01-02
RSC7a and RSC3

Const-me:


Const-me:
About the test — I’m testing the logger. When I debug the test, visual studio shows me those messages.
Unlike algorithms that compute something, I/O code (logging, network calls, draw calls, etc) is nearly impossible to unit-test. I can remove that test, if you want to.

mattheworiordan:
> About the test — I’m testing the logger. When I debug the test, visual studio shows me those messages.

These are automated tests, there is no value in having an automated test that requires a human to confirm if it's working.

> Unlike algorithms that compute something, I/O code (logging, network calls, draw calls, etc) is nearly impossible to unit-test

We have managed it in most other libraries.  You could potentially capture STDOUT from a process you spin up, or you could try stubbing out methods to make sure the underlying logger is called. As a custom logger is a requirement of the spec, see http://docs.ably.io/client-lib-development-guide/features/#RSC4, you could instead set up a custom logger and ensure it is receiving the calls.

Const-me:
“You could potentially capture STDOUT from a process you spin up” — generally, there’s no STDOUT in .NET DLL. Besides, visual studio (that runs unit tests here) spins no processes, instead it loads those DLLs right into the visual studio’s process (in different app.domain, I suppose)

“you could try stubbing out methods” — unfortunately, there’s no way in .NET to dynamically replace a base class with another one. And in your current codebase, the logging is implemented through internal class ExtendedSource : TraceSource where TraceSource is a framework-provided class that does the actual I/O.

“you could instead set up a custom logger and ensure it is receiving the calls” — yep, that method indeed should work.

Const-me:
Unfortunately, the changes weren't local.
But now it should compile on windows phones, and the tests actually verify some stuff.

marto83:
I was wondering why you moved to static log methods instead of what was there. I initially added the interface to let others easily plugin the Ably logs to whatever is currently used. This way if people are using log4net or NLog that can have the logs piped through.

Const-me:
“why you moved to static log methods instead of what was there” — because in your whole codebase, there was no other loggers called other than Logger.Current. This means that the “Current” is redundant. And because logging is going to be used a lot across the app, it’s critical to make logging easy to use. So, in this case, the less typing, the better.

“This way if people are using log4net or NLog that can have the logs piped through” — with my current implementation that’s much simpler then with your one. Only one simple interface method to implement, versus dozen abstract methods before that.
Documentation for Logger.SetDestination method:
        /// <summary>Override logging destination.</summary>
        /// <param name="i">The new destination where to log those messages. Pass null to disable logging completely.</param>
        /// <remarks>This method is mainly for unit tests. However, if in your product you're using some logging infrastructure like nlog or custom logging,
        /// you may want to provide your own <see cref="ILoggerSink"/> implementation to have messages from Ably logged along with your own ones.</remarks>

In fact, my unit test works by providing a custom ILoggerSink implementation.

marto83:
Brilliant. I should have checked all the commits. Ignore me.

mattheworiordan:
@Const-me well interestingly we have moved away from a static class logger and instead opted for an instance bound logger. See https://github.com/ably/ably-js/issues/156 where this has presented a problem when sharing a logger between all instances of the Ably library.  I personally think we should prefer an instance bound logger to a global static class logger.

Const-me:
@mattheworiordan “we have moved away from a static class logger” — in your .NET library, you haven’t moved anywhere. Before my changes, Logger.Current was still a static property, and all logger calls were calling that Logger.Current. If you want to verify that, see my commit 0409f712bff1ce5959af6ff8f0fddaa90041c655.

“where this has presented a problem when sharing a logger between all instances of the Ably library” —in JavaScript it could be. C# however is strongly-typed OO language.
People just don’t write such lambdas in C#.
Lambdas have their disadvantages. (1) Performance cost for creating (2) Parameters are unnamed and undocumented so they’re harder to use. When calling a method or implementing an interface method, you have parameter names + XML documentation for them. (3) They’re harder to read: each variable captured by a lambda is essentially an implicit input/output parameter.

If you want to support more than one key/session simultaneously + distinguish them in logs, one way would be add some context information to ILoggerSink.LogEvent arguments.
Another way — pass ILoggerSink instance when constructing.
However, I don’t expect many users will use the feature, so that change request is IMO low priority.

mattheworiordan:
> C# however is strongly-typed OO language

I am not following why this is relevant?  If a developer customised the log level or logger for one library, and didn't for another, surely they would expect each library to respect the log configuration for that instance. I don't see how strongly-typed languages make any difference.

> People just don’t write such lambdas in C#.

That's fine, a class with a Logger interface is not a problem.  In Ruby that is what we do, see `logger` in http://www.rubydoc.info/gems/ably/Ably%2FRest%2FClient%3Ainitialize

> If you want to support more than one key/session simultaneously + distinguish them in logs, one way would be add some context information to ILoggerSink.LogEvent arguments.

Sure, but I think simply having a logger instance for the client library instance is the most flexible and resilient approach long term.

> However, I don’t expect many users will use the feature, so that change request is IMO low priority.

I agree, happy to raise an issue with low priority to fix later.  Can you do that and we can then merge this PR in?

Const-me:
> I am not following why this is relevant?

That was about using those lambdas in the constructor argument.

> the most flexible and resilient approach long term

IMO exposing logger as the public API ruins flexibility.
Imagine we’ll want to add one more method for logging something else, like to log some custom status object.
If Logger interface is part of the public API (i.e. we’re accepting user-provided loggers), client’s code will stop building saying “you haven’t implemented an interface method”. In Ruby’s case, will probably crash in runtime saying, “Your logger object doesn’t have a method I’m calling”.
If Logger interface is not part of the public API (i.e. we’re accepting a single-method user-provided ILoggerSink object, but keep logger itself our internal API), we’re fine to do such changes without affecting our users. We’re free to change the logging API however we like: move from static classes to non-static or back, add new formatting methods to keep calling code simple — and our users won’t even notice.
That’s what I call flexibility :-)

> Can you do that

Like this? https://github.com/ably/ably-dotnet/issues/33

mattheworiordan:
> Imagine we’ll want to add one more method for logging something else, like to log some custom status object

Fine, but imagine we make any breaking change. This is part of the contract we provide to our developers with our API, so if we want to change anything in future that is breaking, we want the static analysers to report that.

> If Logger interface is part of the public API (i.e. we’re accepting user-provided loggers), client’s code will stop building saying “you haven’t implemented an interface method”

Great

> If Logger interface is not part of the public API (i.e. we’re accepting a single-method user-provided ILoggerSink object, but keep logger itself our internal API), we’re fine to do such changes without affecting our users

Sure, I am happy with that approach. The point was simply that we need a user to be able to customise the log level or log outputter at an instance level.

> Like this? #33

Sure

PR #45, opened 2016-03-09
Async refactor on master

mattheworiordan:
@Const-me I believe this is the PR you wanted to issue?  Please confirm


mattheworiordan:
So as far as I can tell, it looks like all of the tests still use the old API. I cannot accept a PR that does not fix the tests and make sure they pass.  Can you address that?  I appreciate you like to use the `ConsoleTest` as a quick test to make sure things work, however I don't understand why you don't just run one of the acceptance tests individually to do the same thing? We have test coverage for all the actions you are testing in the console test.  We need to have the tests updated and passing before we can merge this into master

mattheworiordan:
> > why you don't just run one of the acceptance tests individually to do the same thing?

> Before I started to change the API heavily, that’s exactly what I did.
> Only created a command-line wrapper over all your tests to be able to read output and use the debugger.
> But now when I’m changing the API, maintaining all your unit tests would be too slow.

Ok, but that means you are introducing a problem that someone else needs to fix and at the same time you are not sure you are not breaking the intended functionality.

> Even if by “maintaining” I only mean “make them compile”.

Sure, but if they don't compile I believe we have a problem too that needs to be fixed.
 
> Another reason is, I want to have a complete integration test in one piece, not scattered in dozen of files.
 
I have no idea what you mean by this.  Our integration tests are huge and would be impractical in one file.  For example, see https://github.com/ably/ably-ruby/blob/master/SPEC.md and all tests that are located in `acceptance` which means they are integration tests.  You want to move all the tests from their topic such as Channel / Messages / Connection into one file so that it's simpler?  I am not following how that makes sense.

> Another reason is, I don’t like the quality of your tests.

Well you're here to fix things, not make things worse by breaking existing tests.

> No status output, lots of copy-pasted code, no comments except copy-pasted Arrange/Act/Assert.

I am not sure I follow you, the tests themselves form the documentation. See https://github.com/ably/ably-dotnet/blob/master/src/Ably.AcceptanceTests/RealtimeAcceptanceTests.cs#L145-L169 for example.  It's named `TestRealtimeClient_Time` so it's clearly testing the Time functionality of a realtime client.  It then defines the `Arrange` part which sets up the test, it calls the method in `Act`, and then in `Assert` it checks the result is correct.  There is nothing wrong with this approach, it follows clear logic.

> Also I don’t understand the authors of those frameworks you use. How’s “ShouldBeEquivalentTo()” is better than “==”?

It's an assertion.  If you say `x == y` it will return false, however if you say `x.ShouldBeEquivalentTo(y)` it will raise an exception if it is not.

Const-me:
> that means you are introducing a problem that someone else needs to fix and at the same time you are not sure you are not breaking the intended functionality.

Maintaining your test takes more time then working on the library.
And IMHO this brings questionable value to the quality of the product.
If you have different view on the value of unit tests, I can certainly fix that but it’ll cost you more (i.e. within the hourly contract).

> I have no idea what you mean by this

I mean what happens in runtime.
In runtime, I want a single application that I can run (and debug!), that connects to the service simulating a customer app using the library, then performs some non-trivial interaction, again simulating a customer app.
Your tests are hundreds of small functions each testing their own miniscule piece of the functionality. They don’t even compile to an .exe. They require complex infrastructure to run. They require complex setup that has nothing common to what happens in the customer’s app. They call private APIs not accessible to the users of the app (see that `[assembly: InternalsVisibleTo("Ably.Tests")]` and `[assembly: InternalsVisibleTo("Ably.AcceptanceTests")]` in AsemblyInfo.cs of the IO.Ably.dll)
That’s why I don’t see much value in them in terms of quality.

> It's an assertion.

I know that, however “==” is much shorter and much easier to read.

mattheworiordan:
> Maintaining your test takes more time then working on the library.

Yes, that is the way of a well tested library.

> And IMHO this brings questionable value to the quality of the product

Unfortunately we disagree, and so do our entire team and all of our other client library devs who believe in tests.

> If you have different view on the value of unit tests, I can certainly fix that but it’ll cost you more

I am quite surprised by this, you are committing code that you know breaks tests, yet you consider the effects of your changes to be outside the scope of the work.  

> In runtime, I want a single application that I can run (and debug!), that connects to the service simulating a customer app using the library, then performs some non-trivial interaction, again simulating a customer app.

That is no reflection on the client library doing it's job, it's simply testing one path.  If you believe you can simulate all scenarios like that and that is more comprehensive than an array of acceptance tests that test individual and group functionality, then I don't know where to being.  For example, say you wanted to test a connection failure during a presence sync, once with token auth, once with key auth, once over JSON and once over MsgPack, how are you going to do that with your console app test given that's already 4 permutations for just one test.  We have hundreds of tests, given the possible permutations there's probably hundreds of thousands of possible scenarios one app could follow.

> Your tests are hundreds of small functions each testing their own miniscule piece of the functionality. They don’t even compile to an .exe.

And why should they compile to an exe, we only need to run them as a test suite.

> They require complex infrastructure to run

They require no infrastructure, they run against sandbox

> They call private APIs not accessible to the users of the app (see that [assembly: InternalsVisibleTo("Ably.Tests")] and [assembly: InternalsVisibleTo("Ably.AcceptanceTests")] in AsemblyInfo.cs of the IO.Ably.dll)

Sounds like an elegant way of exposing inner workings of the Realtime lib to the test suite so that you can inspect message flows between the sandbox env and the client library. We don't get that luxury in other languages so typically have to build in our own hooks, stub, or simulate traffic.

> That’s why I don’t see much value in them in terms of quality.

I don't see any problems in above

Const-me:
> you are committing code that you know breaks tests, yet you consider the effects of your changes to be outside the scope of the work

Yes, I consider unit tests out of scope.

> say you wanted to test a connection failure during a presence sync, once with token auth, once with key auth, once over JSON and once over MsgPack

I don’t see why can’t I test all 4 in a console app.
Also, if you’ll use some clever magic to break DLL encapsulation, replace some components with mockups, and separately test that scenario with different auth, then with different encoding — that proves nothing, because then any of that 4 cases may break in reality on some lower levels of abstractions like network stack or your server.

> They require no infrastructure

By infrastructure I mean megabytes of third-party DLLs not found on the clients systems: nunit.\*.dll, xunit.\*.dll, and also some visual studio IDE components

PR #46, opened 2016-04-07
Fix tests and build

marto83:
I had to go back a bit with portability to get the library stable and get the CI build working again. 

Summary: 
Reverted back from a portable library (once all the features are implemented I'll change it back). 

Fixed the rest unit and integration tests. 

Cleaned up any unused code and started tiding names of classes and methods so we have a single naming convention.


tools/nunit-runners/CHANGES.txt
```diff
@@ -0,0 +1,986 @@
+﻿NUnit 3.2 - March 5, 2016
```
mattheworiordan: Is it necessary to commit this type of stuff?

mattheworiordan:
@marto83 looks good. Not much for me to comment on really given the changes are not specific to the logic or the API, so I'll go with your judgement on whether we should merge this or not.

tools/nunit-runners/CHANGES.txt
marto83: Probably not. I just committed the whole nuget package.

marto83:
This is more for visibility. I'll merge it in the morning.

PR #48, opened 2016-04-13
Specs/rest client

marto83:
Implemented or tagged all the RestClient tests except RSC14d. 

Each test that corresponds with a spec item has an attlibute like this
`[Trait("spec", "RSC15")]`. 

Sometimes an attibute will be applied to a whole test class if all the tests apply to the same spec item.


src/IO.Ably.Tests/Rest/RestSandBoxSpecs.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSC16")]
+        public async Task Time_ShouldReturnAValidDateTimeOffset(Protocol protocol)
+        {
+            var client = await GetRestClient(protocol);
+
+            var now = await client.Time();
+
+            now.Should().BeCloseTo(Config.Now(), (int)TimeSpan.FromHours(1).TotalMilliseconds);
```
mattheworiordan: Do you not need to quantify how close?

src/IO.Ably.Tests/Rest/RestSpecs.cs
```diff
@@ -6,88 +6,10 @@
 using FluentAssertions;
 using IO.Ably.AcceptanceTests;
 using IO.Ably.Auth;
-using IO.Ably.Transport;
 using Xunit;
 
 namespace IO.Ably.Tests
 {
-    public class RestInitSpecs : AblySpecs
```
mattheworiordan: This is removed but `Rest\RestInitiSpecs.cs` was not commited in this commit?

src/IO.Ably.Tests/Rest/RestSpecs.cs
```diff
@@ -233,7 +233,7 @@ public void Init_WithTokenId_SetsTokenRenewableToFalse()
         {
             var rest = new AblyRest(new ClientOptions() { Token = "token_id" });
 
-            rest.TokenRenewable.Should().BeFalse();
+            rest.AblyAuth.TokenRenewable.Should().BeFalse();
```
mattheworiordan: What is `rest.AblyAuth` as oppose do `rest.Auth`?

src/IO.Ably.Tests/Rest/RestSpecs.cs
```diff
@@ -273,6 +289,114 @@ private AblyRest CreateClient(Action<ClientOptions> optionsClient)
...
+                _response.StatusCode = HttpStatusCode.BadGateway;
+                var client = CreateClient(null);
+
+                var ex = await Assert.ThrowsAsync<AblyException>(() => MakeAnyRequest(client));
+                var firstAttemptHosts = _handler.Requests.Select(x => x.RequestUri.Host).ToList();
+                _handler.Requests.Clear();
+                ex = await Assert.ThrowsAsync<AblyException>(() => MakeAnyRequest(client));
+                var secondAttemptHosts = _handler.Requests.Select(x => x.RequestUri.Host).ToList();
+
+                //TODO: Find if there is a better way to assert differnt order
```
mattheworiordan: I doubt it, that's reasonably good though, only a 1 in 14,400 chance of this test failing :)

src/IO.Ably.Tests/Rest/RestSandBoxSpecs.cs
```diff
@@ -54,27 +56,29 @@ public RestSandBoxSpecs(AblySandboxFixture fixture)
...
-            public WithTokenAuthAndInvalidToken(AblySandboxFixture fixture) : base(fixture) { }
+            public WithTokenAuthAndInvalidToken(AblySandboxFixture fixture, ITestOutputHelper output) : base(fixture, output) { }
 
             [Theory]
             [ProtocolData]
             [Trait("spec", "RSC10")]
             public async Task WhenTokenIsRenewable_ShouldRenewToken(Protocol protocol)
             {
                 var authClient = await GetRestClient(protocol);
+                Output.WriteLine("Getting Token to expire in 1 second");
```
mattheworiordan: Should this be committed? I suppose it's only tests

mattheworiordan:
Looking good @marto83

src/IO.Ably.Tests/Rest/RestSandBoxSpecs.cs
marto83: Currently it is set if the time is within the hour. I found an issue in Json.net where the timezone is not correctly deserialized. I've got a test in JsonDeserializationBugTests which is currently skipped. Once I find a way to fix it I can lower the threshold to seconds.

src/IO.Ably.Tests/Rest/RestSpecs.cs
marto83: Oops. That's probably why this commit failed the build :)

src/IO.Ably.Tests/Rest/RestSpecs.cs
marto83: AblyAuth is a way for me to expose the internal Auth object so I can validate things. Users will not see it and will just use the IAuth interface.

mattheworiordan: Ok, thanks for clarifying

src/IO.Ably.Tests/Rest/RestSandBoxSpecs.cs
mattheworiordan: Ok

PR #49, opened 2016-04-18
Specs/auth

marto83:
Finished all the Auth specs except a few that need to have the Realtime client finished. 

Still outstanding
RSA12a - Realtime
RSA7b3 - Realtime 

I still need to do a few integration tests expecially around AuthUrls. There will be one more commit before this is ready to be merged.


src/IO.Ably.Tests/AuthorisationTests.cs
```diff
@@ -7,37 +7,30 @@
...
+        
+        AblyResponse _dummyTokenResponse = new AblyResponse() { TextResponse = "{ \"access_token\": {}}" };
 
-        static Task<AblyResponse> dummyTokenResponse { get { return _dummyTokenResponse.ToAblyResponse(); } }
+        private static string KeyId => ValidKey.Split(':')[0];
 
-        private AblyRest GetRestClient()
+        [Fact]
+        [Trait("spec", "RSA1")]
+        public void WithTlsFalseAndBasicAuth_Throws()
```
mattheworiordan: Do we have a `WithTlsTrueAndBasicAuth_Succeeds`?  Seems we're testing the failure only, but I may have missed another tests

src/IO.Ably/AblyTokenAuth.cs
```diff
@@ -24,6 +24,8 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
         private TokenParams _lastTokenRequest;
         private readonly AblyRest _rest;
 
+        //public string CurrentClientId { get; set; }
```
mattheworiordan: Do we need this if commented out?

src/IO.Ably.Tests/AuthSandboxSpecs.cs
```diff
@@ -0,0 +1,200 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace IO.Ably.Tests
+{
+    [Collection("AblyRest SandBox Collection")]
+    [Trait("requires", "sandbox")]
+    public class AuthSandboxSpecs : SandboxSpecs
+    {
+        public AuthSandboxSpecs(AblySandboxFixture fixture, ITestOutputHelper output) : base(fixture, output)
```
mattheworiordan: Out of interest, why are these called `AuthSandboxSpecs` as opposed to just `AuthSpecs`?

mattheworiordan:
This looks great @marto83, nice work.

One comment, when you rearrange/move code as opposed to add/remove/changing code, it would be nice to have that as a separate PR.  For example, in https://github.com/ably/ably-dotnet/pull/49/commits/29d8bc096007affc55ff8247b84534e359923acd, you have moved a lot of tests such as `WithDefaultParamsAndNothingSpecifiedInMethod_UsesDefaultParams` to a new file.  As such, I don't know if I should reread the entire set of changes, or assume that you simply moved it.  If you keep code moving to a separate commit, I can just ignore the change and assume you've done your job OK if everything compiles. WDYT?

marto83:
Matt. Sounds like a good plan. I'll keep it in mind. I want to make it as easy for you to read through the code. If you have any other suggestions especially around naming of the tests let me know.

src/IO.Ably.Tests/AuthorisationTests.cs
marto83: I'll add another one to validate it works.

src/IO.Ably.Tests/AuthSandboxSpecs.cs
marto83: It is because only these tests actually hit the sandbox environment. I also have a tool that constantly runs all tests on every code change so I exclude these and mainly run them before I commit and push.

src/IO.Ably/AblyTokenAuth.cs
marto83: This has been removed in a further commit.

PR #51, opened 2016-04-22
Specs/rest channel

marto83:
This includes all you can do with a rest channel.


src/IO.Ably.Tests/MessageEncodes/MessageEncodersAcceptanceTests.cs
```diff
@@ -0,0 +1,324 @@
...
+                public string Name { get; set; }
+                public int Age { get; set; }
+                public DateTime DoB { get; set; }
+            }
+
+            public static IEnumerable<object[]> SupportedMessages
+            {  
+                get
+                {
+                    yield return new object[] {new Message("int", 1), "json"};
```
mattheworiordan: Ints are not supported, should we not be checking for failure here, it states that at http://localhost:4000/client-lib-development-guide/features/#RSL4a

mattheworiordan:
Looks good.  I've not read through all the code in detail though as I'm a bit strapped for time at present.  I would recommend you merge away!

PR #52, opened 2016-04-22
Specs/rest presence

marto83:
I've set the base temporarily to specs/RestChannel but once I merge the other one I'll change it to the proper one. 

This includes all the presence tests.


mattheworiordan:
👍

PR #53, opened 2016-04-22
Specs/crypto

marto83:
Cpypto specs

### DIFF
--- a/src/IO.Ably.Tests/EncryptionSpecs.cs
+++ b/src/IO.Ably.Tests/EncryptionSpecs.cs
@@ -1,8 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http.Headers;
+using System.Security.Permissions;
 using System.Text;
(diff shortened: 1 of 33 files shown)


src/IO.Ably.Tests/EncryptionSpecs.cs
```diff
@@ -13,13 +16,80 @@ public class EncryptionSpecs : AblySpecs
...
+
             [Fact]
             [Trait("spec", "RSE1a")]
             public void ShouldReturnCompleteCipherParamsInstance()
             {
-                //Crypto.
+                var result = Crypto.GetDefaultParams();
+                result.Algorithm.Should().Be("AES");
+                result.CipherType.Should().Be("AES-128-CBC");
+                result.KeyLength.Should().Be(128);
```
mattheworiordan: Is this not 256 now?

mattheworiordan:
Ah, ignore previous comment about 128 as you fixed with 9abac8f.  LGTM

PR #57, opened 2016-05-04
Specs/realtime connection

marto83:
This started getting big so I think it will be better to merge it in two or more parts. 

Includes the first 14 spec items from connection specs. 
The few other refactorings. 
- I created an internal thread that manages. 
- Simplified about half of the existing realtime tests to make them easier to maintain
- Got the ConnectionAwaiter to work properly and wait for an specific connection state without blocking. 
- Added more logging


src/IO.Ably/AblyRealtime.cs
```diff
@@ -13,23 +11,29 @@ public class AblyRealtime : IRealtimeClient
...
+        public AblyRealtime(ClientOptions options) : 
+            this(options, clientOptions => new AblyRest(clientOptions))
+        {
+            
+        }
+
+        internal AblyRealtime(ClientOptions options, Func<ClientOptions, AblyRest> createRestFunc)
         {
-            RestClient = new AblyRest(options);
+            RestClient = createRestFunc(options);
```
mattheworiordan: Not a great name `createRestFunc`!  Perhaps simple `createRestClient`?

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -49,6 +49,16 @@ public void WithAutoConnect_CallsConnectOnTransport()
                 LastCreatedTransport.ConnectCalled.Should().BeTrue();
             }
 
+            [Fact]
+            [Trait("spec", "RTN3")]
+            public void WithAutoConnectFalse_LeavesStateAsInitialized()
+            {
+                var client = GetClientWithFakeTransport(opts => opts.AutoConnect = false);
```
mattheworiordan: Why are you using a fake transport, is that necessary?  Where possible, the less stubbing the better.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -197,6 +198,24 @@ public void EmittedEventTypesShouldBe()
                 });
             }
 
+            [Fact]
+            [Trait("spec", "RTN4b")]
+            [Trait("sandboxTest", "needed")]
```
mattheworiordan: What does `sandboxTest` needed mean?  An integration test?

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -197,6 +198,24 @@ public void EmittedEventTypesShouldBe()
...
+                var expected = new[] {ConnectionStateType.Connecting, ConnectionStateType.Connected,};
+                var states = new List<ConnectionStateType>();
+                client.Connection.ConnectionStateChanged += (sender, args) =>
+                {
+                    states.Add(args.CurrentState);
+                };
+
+                client.Connect();
+
+                states.Should().BeEquivalentTo(expected);
```
mattheworiordan: I must admit that I am not sure of the value of these stubbed tests if you are only going to add a Sandbox test anyway.  Personally I use unit tests where you are testing something different to what an integration test would be testing.  Given our client lib is all about how it behaves with the realtime service, I would err on the side of using an integration test over a unit test with components stubbed out.

src/IO.Ably.Tests/Rest/RestSpecs.cs
```diff
@@ -417,6 +417,7 @@ private AblyRest CreateClient(Action<ClientOptions> optionsClient)
 
             [Fact]
             [Trait("spec", "RSC15a")]
+            [Trait("intermitten", "true")]
```
mattheworiordan: Typo

src/IO.Ably.Tests/Infrastructure/AblySpecs.cs
```diff
@@ -9,12 +9,14 @@ public abstract class AblyRealtimeSpecs : MockHttpRestSpecs
         internal virtual AblyRealtime GetRealtimeClient(ClientOptions options = null, Func<AblyRequest, Task<AblyResponse>> handleRequestFunc = null)
         {
             var clientOptions = options ?? new ClientOptions(ValidKey);
+            clientOptions.SkipInternetCheck = true; //This is for the Unit tests
```
mattheworiordan: Is there not a better way to hide this option?  I assume this new option will come up in people's IDEs now yet it is confusing and will be without documentation

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -69,20 +69,52 @@ public void WithAutoConnectFalse_LeavesStateAsInitialized()
             }
 
             [Fact]
+            [Trait("spec", "RTN19")]
             public void WhenConnectedMessageReceived_ConnectionShouldBeInConnectedStateAndConnectionDetailsAreUpdated()
             {
-                var client = GetClientWithFakeTransport(opts => opts.AutoConnect = false);
+                var client = GetClientWithFakeTransport();
```
mattheworiordan: Why a fake, sandbox will return a valid `connectionDetails` attribute in all cases?

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -335,7 +337,310 @@ public EventEmitterSpecs(ITestOutputHelper output) : base(output)
             }
         }
 
+        [Trait("spec", "RTN7")]
```
mattheworiordan: This entire section is very well tested, but should also have integration tests that at least ensure headline requirements are met.  I am a bit concerned by the level of stubbing which is not advisable given the client is intrinsically linked with the service to work.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -382,5 +382,31 @@ public AckNackSpecs(ITestOutputHelper output) : base(output)
...
         }
+
+        [Trait("spec", "RTN8")]
+        public class ConnectionIdSpecs : ConnectionSpecs
+        {
+            [Fact]
+            [Trait("spec", "RTN8a")]
+            public void ConnectionIdIsNull_WhenClientIsNotConnected()
+            {
+                var client = GetClientWithFakeTransport(opts => opts.AutoConnect = false);
```
mattheworiordan: As with other comments, why are we stubbing this, this would work exactly the same without stubbing.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -426,9 +427,75 @@ public void OnceConnected_ShouldUseKeyFromConnectedMessage()
...
+                var transportParams =await client.ConnectionManager.CreateTransportParameters();
+
+                transportParams.ConnectionSerial.Should().Be(targetSerial);
+            }
+
+            [Fact]
+            public void ConnectionSerialUpdated_WhenProtocolMessageReceived()
+            {
+                // Arrange
+                var client = GetClientWithFakeTransport();
```
mattheworiordan: Why fake it, why not just send a few messages, receive them, check that the serial number is accurate.  This test and the previous one, IMHO, are not really reflecting what `RTN10c` is trying to achieve as there are too many fake things going on

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -489,6 +490,27 @@ public void ConnectionSerialUpdated_WhenProtocolMessageReceived()
...
+                FakeMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Connected));
+                var initialSerial = client.Connection.Serial;
+
+                FakeMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Message));
+
+                // Act
+                client.Connection.Serial.Should().Be(initialSerial);
```
mattheworiordan: Fine, but this should really just be tested with some valid `ProtocolMessage`s that don't contain serials as you're then getting two for one, i.e. better test coverage overall, and a real test against Sandbox.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
```diff
@@ -521,13 +529,75 @@ public ConnectionSerialSpecs(ITestOutputHelper output) : base(output)
...
+
+                result.IsSuccess.Should().BeTrue();
+                result.Value.Value.Should().Be(Now);
+            }
+
+            [Fact]
+            [Trait("spec", "RTN13b")]
+            public async Task WithClosedOrFailedConnectionStates_ShouldReturnError()
+            {
+                var client = GetClientWithFakeTransport();
```
mattheworiordan: Why fake it, just open & close a connectiojn

mattheworiordan:
Overall this looks good @marto83. I do think there is an unnecessary amount of fakes and stubs where they are not needed.  I am a big advocate of unit testing, but it feels as though you are not unit testing, but instead replacing components of the system with fakes and then testing the whole system.  I feel unit tests are better suited to test a single component instead and fake all of its dependencies, whereas you seem to be doing the reverse. Then, where tests can be run against sandbox, in almost all cases we should do that because it is more likely to flush out issues early on against a real host.

src/IO.Ably/ClientOptions.cs
```diff
@@ -164,6 +165,8 @@ internal AuthMethod Method
             }
         }
 
+        internal bool SkipInternetCheck { get; set; } = false;
```
marto83: Yes. That will not appear to users.

marto83:
Hi Matt. Fair point about the faking. I've been depending on the fake transport for several reasons. 
1. To make the tests smaller so they test the bits from the spec
2. To make them run quicker. I run the whole test suite continuously and it catches logic issues as they happen. There are currently about 9 to 1 unit tests to integration tests and the unit tests run very quickly
3. It helps to exactly test the interaction between the components. I can tweak the message and fake some of the stuff. 


You are right that most issues are discovered when using the integration tests. I manage to get it 90% right with unit tests and then do some bigger integration tests to iron out the production issues. For places that I'm not 100% confident I mark it with sandboxtest needed to make sure I don't forget to add it.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
marto83: True but this way I can run it much quicker and more often. For the sandbox tests I need to setup an app every time.

mattheworiordan: > True but this way I can run it much quicker and more often.

I'd take accuracy over frequency any day!

> For the sandbox tests I need to setup an app every time

This should only be needed once per test suite run.  However, if you want quicker app setup, why not simply add a cache for the credentials whilst you are working i.e. if a static file exists, then use those credentials for example, if not, then create a new app?

mattheworiordan:
> You are right that most issues are discovered when using the integration tests. I manage to get it 90% right with unit tests and then do some bigger integration tests to iron out the production issues. For places that I'm not 100% confident I mark it with sandboxtest needed to make sure I don't forget to add it.

I appreciate unit tests are faster, but I don't think that in itself is a good enough reason.  With fakes, unfortunately you don't get latency and pauses that will occur in reality.  So race conditions rarely happen in these types of envs, but will happen in reality.  I believe we should use unit tests to test single classes only, and use integration tests in most cases.  Right now, IMHO, having a spec item as completed using fakes does not prove it works unfortunately.  If perhaps you want a set of fake tests that you can run for your continuous tests to help iron out issues quickly you should do that, but that is not a substitute for acceptance tests that prove this works in the real world.

src/IO.Ably.Tests/Realtime/ConnectionSpecs.cs
marto83: That will work. Currenty I setup one per test run and usually works quite well.

Issue #33, opened 2016-01-05
Per-instance loggers


Const-me:
It's desirable to be able to have different loggers for different ably instances.

See also (RSC4) A custom logger can be provided in the constructor

marto83:
I think this is a minor nice to have and in most cases you won't have multiple instances running in the same environment. 
Closing for now. If requested we can reopen it later.

mattheworiordan:
Ok, but can you set the log level in ClientOptions when instanced like all other client libraries?  If not, we need to raise an issue for that.

withakay:
Whilst the OPs requests is now technically possible, the feature is not exposed publicly and based on internal discussions some time ago, it seems unlikely that this will happen, so I am going to close this for now. Feel free to re-open if you feel this is an important feature.

PR #127, opened 2017-09-05
Isolated logger for reliable testing

withakay:
I have pushed down the static Logger code into an internal class (InternalLogger) that can be instanced from the test suite. This prevents interactions when running tests in parallel. The Logger interface (small 'i') is unchanged and uses a singleton instance of InternalLogger to do its work.

### DIFF
--- a/src/IO.Ably.Shared/Logger.cs
+++ b/src/IO.Ably.Shared/Logger.cs
@@ -21,14 +21,14 @@ namespace IO.Ably
         void LogEvent(LogLevel level, string message);
     }
 
+    /// <inheritdoc />
     /// <summary>The default logger implementation, that writes to debug output.</summary>
     internal class DefaultLoggerSink : ILoggerSink
     {
-        readonly object syncRoot = new object();
-
+        private readonly object _syncRoot = new object();
         public void LogEvent(LogLevel level, string message)
         {
-            lock (syncRoot)
+            lock (_syncRoot)
             {
                 Debug.WriteLine("Ably: [{0}] {1}", level, message);
             }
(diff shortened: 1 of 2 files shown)


mattheworiordan:
LGTM. To be honest I am not sure why the logger is static in the first place.  I believe in all other client libs it's instanced with the library.  I assume changing it now though would cause a world of pain for anyone using the static Logger right?  Perhaps something we should defer until a 2.0 release?

withakay:
@mattheworiordan The Logger is (erm, was) static, but an ILoggerSink implementation can be passed in (although the property it is assigned to is static). As @marto83 has mentioned in his notes using something like [LibLog](https://github.com/damianh/LibLog) would have been preferable, but it is a lot of work to retro fit and would almost certainly break the contract ILoggerSink has created, so I think on balance probably not worth pursuing. Down the line we could add ILoggerSink implementations for popular logging libraries, or perhaps just include some examples in the docs.

mattheworiordan:
> The Logger is (erm, was) static, but an ILoggerSink implementation can be passed in (although the property it is assigned to is static)

Ok, so this changes that to use an instance variable now right?

Out of interest, if a dev sets the log level, is that done at an class instance level then or at the static level?  I believe that is a far more common use case (well at least in when it's being used server-side).

> Down the line we could add ILoggerSink implementations for popular logging libraries, or perhaps just include some examples in the docs.

Yup, I think a example would suffice to be honest.

withakay:
>> The Logger is (erm, was) static, but an ILoggerSink implementation can be passed in (although the property it is assigned to is static)

> Ok, so this changes that to use an instance variable now right?

Yes an instance variable is now used, but perhaps not as you might be thinking... The InternalLogger instance is essentially a private singleton* on the static Logger class, the static Logger class then delegates to the InternalLogger instance. So all the code that accesses Logger does so in the same way as before, the contract has not been broken. This allows me to create an instance of the InternalLogger that is detached from the main lib to run tests against in isolation.

*I know singletons are seen as a bit of an anti pattern these days (and generally not without reason) but I couldn't see a better alternative given the circumstances.

> Out of interest, if a dev sets the log level, is that done at an class instance level then or at the static level?  I believe that is a far more common use case (well at least in when it's being used server-side).

It will still effectively be static, so if used in and ASP.NET WebAPI project, the log level would be global for that project and not per client, for example. This is similar to there libraries I have worked with ([PayPal](https://github.com/paypal/PayPal-NET-SDK) is the only one that comes to mind right now). I am happy to reevaluate if you have an alternative?

>> Down the line we could add ILoggerSink implementations for popular logging libraries, or perhaps just include some examples in the docs.

> Yup, I think a example would suffice to be honest.

marto83:
@mattheworiordan the logger started as static as this was how it was implemented in the java lib and that's what I followed as an example. Since then it has never been a priority to change it as you would rarely init two versions of the library with different log levels and logger sinks. However, it's been a minor pain in the tests so I think this is making it better. :)

withakay:
I have had no objects, so I am going to close this.

withakay:
I have had no objections, so I am going to merge this and close.

mattheworiordan:
>  I am happy to reevaluate if you have an alternative?

No, no objection.  I will create an issue to later review this though as it's not consistent with the other libraries.

mattheworiordan:
In fact, seems we have a very old issue https://github.com/ably/ably-dotnet/issues/33 for this already

marto83:
In fact in the Java lib logging is still done with statics: https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/util/Log.java

mattheworiordan:
> In fact in the Java lib logging is still done with statics: https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/util/Log.java

Bugger, it is :(

PR #133, opened 2017-09-11
A fix to comply with RTL4b 

withakay:
Using the existing IsTerminalConnectionState (which returns true if the connection is Closed, Closing, Suspended Or Failed and false otherwise) to check if an exception should be thrown when attaching a channel as per RTL4b.

This feature is covered by [this test](https://github.com/ably/ably-dotnet/blob/025cf1e2eccfd3e544e64d0f0da98ca75ac7e343/src/IO.Ably.Tests/Realtime/ChannelSpecs.cs#L268), which now passes.

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Channel.cs
+++ b/src/IO.Ably.Shared/Realtime/Channel.cs
@@ -140,9 +140,9 @@ namespace IO.Ably.Realtime
                 return;
             }
 
-            if (ConnectionState == ConnectionState.Failed)
+            if (IsTerminalConnectionState)
             {
-                throw new AblyException("Cannot attach when connection is in Failed state");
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Shared/Realtime/Channel.cs
```diff
@@ -140,9 +140,9 @@ public void Attach(Action<bool, ErrorInfo> callback = null)
                 return;
             }
 
-            if (ConnectionState == ConnectionState.Failed)
+            if (IsTerminalConnectionState)
             {
-                throw new AblyException("Cannot attach when connection is in Failed state");
+                throw new AblyException($"Cannot attach when connection is in {ConnectionState.ToString()} state");
```
marto83: Just a note. You don't need to call `ToString` on ConnectionState

withakay:
removed the ToString() call :)

PR #137, opened 2017-09-15
Injectable INowProvider

withakay:
To enabled more reliable testing I have replaced the usage of the static Config.Now() with a local Now() method for each class that needs it.

The Now() method implementation is passed in to the constructor as an implementation of an INowProvider. 

An INowProvider implementation can be passed in externally via ClientOptions, if not specified then the existing default behaviour is used (DateTimeOffset.UtcNow)

For classes that need access to Now(), I have added a class called NowProviderConcern. This is designed to be inherited (or on classes that are already inheriting, instanced and composed implementing the INowProvider interface). I am not entirely convinced by the name 'NowProviderConcern' but the idea is that is is used by classed that are concerned with providing Now(), I am open to alternate suggestions.

For a specific examples of why this is useful see the tests 'ShouldSuspend_WhenFirstAttemptEqualOrGreaterThanConnectionStateTtl_ShouldReturnTrue' or 'WhenInSuspendedState_ShouldTryAndReconnectAfterSuspendRetryTimeoutIsReached'

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -5,13 +5,15 @@ using System.Net.Http;
 using Newtonsoft.Json.Linq;
 using Newtonsoft.Json;
 using System.Threading.Tasks;
+using IO.Ably.Shared;
 
 namespace IO.Ably
 {
-    internal class AblyAuth : IAblyAuth
+    internal class AblyAuth : NowProviderConcern, IAblyAuth
     {
         internal AblyAuth(ClientOptions options, AblyRest rest)
         {
+            NowProvider = options.NowProvider;
             Options = options;
             _rest = rest;
 
@@ -65,7 +67,7 @@ namespace IO.Ably
             }
             else if (Options.Token.IsNotEmpty())
             {
-                CurrentToken = new TokenDetails(Options.Token);
+                CurrentToken = new TokenDetails(Options.Token, Options.NowProvider);
             }
             LogCurrentAuthenticationMethod();
         }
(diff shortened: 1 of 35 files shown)


marto83:
Looks good. I can see how it is and will be useful in tests that deal with timing. 
I'm not sure about the name of `NowProviderConcern` as well but can't suggest anything better. So it might just grow on me. 

Make sure you don't expose the NowProvider to the clients. Make it all internal. We don't want to be providing an option for them to supply a NowProvider and we don't want them to be wondering about it either. Also if you make it internal you can refactor it without breaking anyone that tried to use it.

withakay:
It could just be NowProvider or NowConcern perhaps. It is not super important and can easily be refactored later, but the name does bug me a little. 

I will go over the code again and make sure I am not exposing anything I shouldn't before merging

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -5,13 +5,15 @@
 using Newtonsoft.Json.Linq;
 using Newtonsoft.Json;
 using System.Threading.Tasks;
+using IO.Ably.Shared;
 
 namespace IO.Ably
 {
-    internal class AblyAuth : IAblyAuth
+    internal class AblyAuth : NowProviderConcern, IAblyAuth
```
mattheworiordan: Do we really need to pollute the AblyAuth class with a new interface / class that it inherits from.  Given all it does is simply map `Now() -> NowProvider.Now()` why not just replace all instances of `Now()` with `NowProvider.Now()` and remove this abstraction?  I don't really follow what the `NowProviderConcern` really adds other than obfuscating how `Now()` works?

withakay: I think you are absolutely right.
A better solution is just to add in internal INowProvider property to each class that needs it.

PR #139, opened 2017-09-18
Injectable logger

withakay:
Replace the use of static logger with instance that is configurable.


src/IO.Ably.Shared/ILogger.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace IO.Ably
+{
+    public interface ILogger
```
mattheworiordan: Out of interest, does this interface match any common Logger classes / systems out there, or is this Ably specific?  It would be nice (but not necessary, and especially not now) if we did use generic Logger interfaces.

withakay: There is not really such a thing for .Net, the best we could do is provide adapters to common logging libraries, either in this package or (probably better) as an add on.

PR #140, opened 2017-09-22
Simplify now provider

withakay:
This PR simplifies the internal resolution of Now(), removing the ill conceived NowProviderConcern and replacing the use of the INowProvider interface with a simple delegate (Func<DateTimeOffset>) that is passed into the appropriate constructors.


withakay:
Am I OK to merge this @mattheworiordan ?

PR #142, opened 2017-09-28
simplified the message queue logic in ConnectionManager

withakay:
Previously messages would be added to a list if the Connection Manager was transitioning to a new state, but processed immediately if not. 
If found this logic quite hard to follow when looking for possible race conditions but I could see the intent, so I have changed it so all messages are queued using a ConcurrentQueue and if the state is not transitioning when the message is received, the queue is processed. 

Additionally I have changed the way OnTransportDataReceived is executed, previously this was wrapped in Task that had task.ConfigureAwait(false) called on it, this has the effect of running the task on a new thread (from the thread pool) and returning immediately (essentially fire and forget), which has the potential to jumble the message ordering. 
I could not see any benefit to this being async/parallel so I change the task to use WaitAndUnwrappedExceptions().

These changes have made the echo message ordering problem go away, without any noticeable side effects.

### DIFF
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -36,7 +36,6 @@ namespace IO.Ably.Transport
         public Connection Connection { get; }
         public ConnectionState ConnectionState => Connection.State;
         private readonly object _stateSyncLock = new object();
-        private readonly object _transportQueueLock = new object();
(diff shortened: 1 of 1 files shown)


withakay:
No further feedback or objections, so I am going to merge this.

mattheworiordan:
Nice, thanks @withakay

mattheworiordan:
Sorry for the delay. 👍

PR #145, opened 2017-10-09
Make MsgPack optional and disabled by default

withakay:
This PR add a compile time constant `MSGPACK` that is now required to build against MsgPack.cli. By default this constant is not present and as such MsgPack support will not be built in to the library by default anymore.

To achieve this MsgPack specific code has been removed using compiler preprocessor statements (`#if MSGPACK`) and by moving some specific classes to a separated shared project and using a conditional on the import statement `<Import Project="..\IO.Ably.Shared.MsgPack\IO.Ably.Shared.MsgPack.projitems" Label="Shared" Condition="$(DefineConstants.Contains(MSGPACK))" />`

Additionally I have modified the build and package scripts to take a `-msgpack` switch, this will set the MSGPACK constant and, in the case of the package script, direct it to use a MsgPack specific nuspec file that declares the dependency.

There are no public facing code changes, the `useBinaryProtocol` Client Option is still present but has no effect when MsgPack support has been disabled.


functions.ps1
```diff
@@ -1,20 +1,23 @@
...
+	Write-Output "Constants: $constants"
 
 	try {
 		switch($configuration)
 		{
-			"package" { exec { & $msbuild $solutionPath "/t:clean;build" "/p:Configuration=release;Platform=Any CPU" "/p:AssemblyOriginatorKeyFile=$signKeyPath" "/p:SignAssembly=true" "/p:DefineConstants=PACKAGE" } }
-			"release" { exec { & $msbuild $solutionPath "/t:clean;build" "/p:Configuration=$configuration;Platform=Any CPU"} }
-			default { exec { & $msbuild $solutionPath "/t:clean;build" "/p:Configuration=$configuration;Platform=Any CPU"  } }
+			"package" { exec { & $msbuild /nologo $solutionPath "/t:clean;build" "/p:Configuration=release;Platform=Any CPU" "/p:AssemblyOriginatorKeyFile=$signKeyPath" "/p:SignAssembly=true" /p:DefineConstants="$constants" } }
+			"release" { exec { & $msbuild /nologo $solutionPath "/t:clean;build" "/p:Configuration=$configuration;Platform=Any CPU" "/p:DefineConstants=$constants"} }
```
marto83: Nice. I didn't know you can pass constants like that 👍

nuget/io.ably_msgpack.nuspec
```diff
@@ -0,0 +1,72 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+    <metadata>
+        <id>ably.io</id>
+        <version>0.8.5</version>
+        <title>ably.io</title>
+        <authors>Martin Georgiev, Yavor Ivanov</authors>
```
marto83: Jack, make sure you add your name to the list here.

src/IO.Ably.Android/WebsocketTransport.cs
```diff
@@ -188,8 +188,10 @@ private void socket_DataReceived(object sender, DataReceivedEventArgs e)
             {
                 try
                 {
+#if MSGPACK
```
marto83: If MSGPACK is disabled add a log message because that shouldn't be happening. It's always good to know.

src/IO.Ably.Net45/WebsocketTransport.cs
```diff
@@ -189,8 +189,10 @@ private void socket_DataReceived(object sender, DataReceivedEventArgs e)
             {
                 try
                 {
+#if MSGPACK
                     var message = MsgPackHelper.DeserialiseMsgPackObject(e.Data).ToString();
                     Logger.Debug("Websocket data message received. Raw: " + message);
+#endif
```
marto83: same as above. ☝️ Just add a log message to say that's unexpected when the binary protocol is off.

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -53,26 +52,33 @@ public IEnumerable<PresenceMessage> ParsePresenceMessages(AblyResponse response,
                 ProcessMessages(messages, options);
                 return messages;
             }
-
+#if MSGPACK
             var payloads = MsgPackHelper.Deserialise(response.Body, typeof(List<PresenceMessage>)) as List<PresenceMessage>;
             ProcessMessages(payloads, options);
             return payloads;
+#else
+            throw new AblyException($"Response of type '{response.Type}' is not enabled for this build. Use ResponseType.Json");
```
marto83: ResponseType.Json will not make sense to a customer as it's an internal time. Maybe just remove the last bit.

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -53,26 +52,33 @@ public IEnumerable<PresenceMessage> ParsePresenceMessages(AblyResponse response,
...
                 ProcessMessages(messages, options);
                 return messages;
             }
 
+#if MSGPACK
             var payloads = MsgPackHelper.Deserialise(response.Body, typeof(List<Message>)) as List<Message>;
             ProcessMessages(payloads, options);
             return payloads;
+#else
+            throw new AblyException($"Response of type '{response.Type}' is not enabled for this build. Use ResponseType.Json");
```
marto83: same here Re: ResponseType

marto83:
It looks good. Only a few tiny comments 👍

src/IO.Ably.Net45/WebsocketTransport.cs
withakay: If MSGPACK is disabled then e.Data should be JSON, so could I do something like below?

Logger.Debug("Websocket data message received. Raw: " + Encoding.UTF8.GetString(e.Data));

examples/AndroidSample/AndroidSample.csproj.bak
```diff
@@ -0,0 +1,192 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mattheworiordan: Why are you committing a bak file?

nuget/io.ably.nuspec
```diff
@@ -15,38 +15,44 @@
...
         <language />
         <dependencies>
+          
           <group>
             
           </group>
+          
           <group targetFramework="net46">
-            <dependency id="MsgPack.Cli" version="[0.9.1]" />
+            <!-- Some dependecies are merged using ILRepack, see csproj -->
```
mattheworiordan: Typo, `dependencies`

src/IO.Ably.Android/WebsocketTransport.cs
mattheworiordan: 👍

mattheworiordan: Actually, we should throw an exception as MsgPack is not supported surely?

src/IO.Ably.Net45/WebsocketTransport.cs
mattheworiordan: I think we should raise an exception.  This code path should never happen when MsgPack is disabled.

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -93,9 +99,12 @@ private void LogRequestBody(byte[] requestBody)
         {
             try
             {
+#if MSGPACK
                 var body = MsgPackHelper.DeserialiseMsgPackObject(requestBody)?.ToString();
-                
                 Logger.Debug("RequestBody: " + (body ?? "No body present"));
+#else
+                Logger.Debug("RequestBody: MsgPack disabled, cannot log request");
```
mattheworiordan: Throw exception too?

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -113,12 +122,17 @@ public byte[] GetRequestBody(AblyRequest request)
                     request.ChannelOptions);
 
             byte[] result;
-            if (_protocol == Protocol.Json)
+            if (_protocol == Protocol.Json || !Config.MsgPackEnabled)
```
mattheworiordan: This feels wrong.  Shouldn't the build ensure Protocol is always JSON?  If not, we have a problem in the build and should raise an exception no?

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -113,12 +122,17 @@ public byte[] GetRequestBody(AblyRequest request)
...
+            {
                 result = JsonHelper.Serialize(request.PostData).GetBytes();
+            }
             else
             {
+#if MSGPACK
                 result = MsgPackHelper.Serialise(request.PostData);
+#endif
```
mattheworiordan: As above, and i'll stop commenting for all others and assume you check all if you agree

mattheworiordan:
@withakay this looks good, nice work.

Few comments:

- I think code paths for builds w/o MsgPack that should never reach MsgPack code should always raise an exception.  
- How will this now impact our builds in terms of customers using the library?  Will the default build now not include MsgPack?  I assume so.  If so, do we even want to document how people can use the MsgPack version, or leave it off completely until we are confident it's stable?

examples/AndroidSample/AndroidSample.csproj.bak
withakay: An error, I will remove it.

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
withakay: This *should* not be needed, I was being extra cautious.

src/IO.Ably.Android/WebsocketTransport.cs
withakay: I will fix up a version that throws an exception, for review.

nuget/io.ably.nuspec
withakay: Will fix

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
withakay: I put a check in the constructor and throw an exception, something like

`public MessageHandler(ILogger logger, Protocol protocol)
        {
            if(!Config.MsgPackEnabled && protocol == Protocol.MsgPack)
                throw new AblyException("Invalid Protocol 'MsgPack'. MsgPack is disabled in this build.");
...
}`

withakay:
@mattheworiordan 

Regarding your first point, I can have another pass over the code and see where I can add some guards that raise exceptions, but I am thinking maybe this is the wrong approach. ClientOptions.UseBinaryProtocol can only return false if MSGPACK is not defined and the only time the protocol gets set it is based on the value of UseBinaryProtocol. 
Although adding checks that throw feels safer ('it can't hurt' etc.) but actually it is probably just adding noise to the code. What do you think?

Regarding builds, the default build will not include MsgPack in anyway, there is no dependency declared and I have double checked that it is not being inlined. (Note, the projects still have a reference to the MsgPack package, this makes it easy to enable with the -msgpack switch on the build script, I am assuming at some point we will come back and give it another try). 
When you say 'do we even want to document how people can use the MsgPack version...' are you alluding to something I have done in this PR or that just a general question? If the latter, then my feeling is that it is an implementation detail and we don't really need draw to much attention to it.

src/IO.Ably.Android/WebsocketTransport.cs
withakay: actually not sure about this, see comment elsewhere

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
withakay: I went back and looked at why I did this this way. The reason is that not having the !Config.MsgPackEnabled check causes the build to fail with 'Error CS0165: Use of unassigned local variable 'result' (136, 65)'. This is because, although we know that _protocol == Protocol.Json will always be true when MsgPack is not enabled, the compiler does not.
The obvious solution to that is to conditionally build in the Protocol.MsgPack enum, so when MsgPack is not enabled Protocol.MsgPack would not exist. I expect doing that will have some side effects and requires changes elsewhere, but it would make reasoning about the protocol easier. I will investigate.

withakay: I have done the work of removing Protocol.MsgPack altogether, on another branch, diff here https://github.com/ably/ably-dotnet/compare/remove-msgpack...remove-msgpack-alt?expand=1

src/IO.Ably.Net45/WebsocketTransport.cs
marto83: @withakay if you have useBinary false you will never receive binary data from the socket. You can safely throw an exception

mrueedi:
So 0.8.6-beta3 does not support the expected default protocol "useBinaryProtocol"?  If yes, this limitation should be mentioned in the release notes because the JSON based protocol is advertised as [inferior vs. the binary protocol](https://support.ably.io/support/solutions/articles/3000047365-do-you-binary-encode-your-messages-for-greater-efficiency).

mattheworiordan:
@mrueedi you are correct, MsgPack has been removed, an dis detailed in https://github.com/ably/ably-dotnet/releases/tag/0.8.6-beta3.  You are right from a purely technical perspective a binary protocol is typically better than a text protocol, however in reality the difference is often small (a few bytes) and we definitely think customers would prefer stability.  Once the upstream MsgPack is stable across platforms, we'll be reinstituting it like all other libraries that support MsgPack.

Issue #141, opened 2017-09-28
XML documentation / types where necessary


mattheworiordan:
The AuthCallback of ClientOptions is defined with an Object type.  This is because it can return a TokenDetails or TokenRequest object.  Documentation as follows:

> A function which is called when a new token is required. The role of the callback is to either generate a signed TokenRequest which may then be submitted automatically by the library to the Ably REST API requestToken; or to provide a valid token in as a TokenDetails object. See an authentication callback example or our authentication documentation for details of the token request format and associated API calls.

In some libraries a JSON object is supported to.

A customer has suggested the following:

> Either add XML comments so that the dev knows what he can use, either change the syntax so that it can be strongly typed. Maybe a where type constraint?

Where we have these loose object types that make it difficult in the IDE to see what return type is expected, we should try and help developers.  Our plan is not to move our documentation into the libraries, but the libraries should be intuitive enough to ensure developers can work in their IDEs and use our docs where appropriate.

tcard:
In iOS, we define [a TokenDetailsCompatible](https://github.com/ably/ably-ios/blob/85fd3b53f84a5dd7ba6ea7e9d2f09df897397008/Source/ARTAuthOptions.h#L15) protocol (interface, in C# parlance), make TokenDetails and TokenRequest implement it (and String, but I believe in C# you can't implement interfaces for types already defined) and then use TokenDetailsCompatible in the auth callback.

tcard:
PS. In case it isn't clear, a TokenDetailsCompatible is something that can be turned into a TokenDetails and hence has a single method: `toTokenDetails`. For TokenDetails, it just returns `this`. For TokenRequest, it executes itself (ie. posts it to Ably) and returns the resulting TokenDetails.

withakay:
A `where` type constraint is not possible on a Func/delegate unfortunately. @tcard's suggestion is good, but in the c# lib TokenRequest is not responsible for posting itself, it is effectively a DTO with some initialisation methods, the current design doesn't make it easy to change that, best I can tell. There does some to be some intersection between TokenDetails and TokenRequest (namely ClientId and Capability) so an interface could be extracted, but without a common ToTokenDetails() method I don't think there would be much benefit.

So that leaves improving the inline docs, which I have taken the liberty of doing.

![screen shot 2017-10-09 at 20 24 13](https://user-images.githubusercontent.com/249723/31354935-2bbbb4c0-ad30-11e7-9ce0-d277a60f7b0e.png)

[...]

PR #146, opened 2017-10-09
Improve the XML Docs for AuthCallback

withakay:
Improve XML Doc for AuthCallBack, see https://github.com/ably/ably-dotnet/issues/141 for details


src/IO.Ably.Shared/AuthOptions.cs
```diff
@@ -72,6 +73,8 @@ public AuthOptions Merge(AuthOptions defaults)
             if (Key.IsEmpty()) Key = defaults.Key;
             if (UseTokenAuth.HasValue == false) UseTokenAuth = defaults.UseTokenAuth;
             if(QueryTime.HasValue == false) QueryTime = defaults.QueryTime;
+
```
mattheworiordan: Why the extra space?

Issue #114, opened 2017-06-27
RealtimePresence possibly broken


SimonWoolf:
Customer reports that with .NET, using the AblyRealtime client, they only ever see one member present (presumably themselves) even when there are definitely multiple on the channel (as seen from other client libs or the AblyRest client).

Looking at the lib implementation, I thought at first it was because [`waitForSync` isn't implemented yet](https://github.com/ably/ably-dotnet/blob/e76f6c94df952171a38bc84a2f5cfd1ac1a42e6f/src/IO.Ably.Shared/Realtime/Presence.cs#L51), but customer reports that they still only see one member if they add a delay before getting presence.

It's still possible that it's user error of some kind. But hard to be sure, because looking at the specs, there's only one test for GetAsync with the realtime client, and that only checks for a single member (itself), which isn't inconsistent with the report.

So would be nice if we could implement `waitForSync`, and then have a proper presence test where a couple of clients enters, then a different client attaches, gets presence, and sees the first two clients, so we can know for sure if the reported behaviour is a client lib bug or user error.

marto83:
Would you be able to get the customer to enable the debug logging. We can see what happens to the presence messages when they arrive. We've had a few people using it before without issues. I'll double check it and let you know.

If waitForSync is needed I can prioritize it and get it implemented in the next couple of weeks.

SimonWoolf:
Fair enough. I've asked him for a debug log and to post it directly into this issue.

jeraldAR:
Hi @marto83,

This is what I get when trying to get all presence members in our channel. I am developing an application in WPF using Ably.io Latest prerelease 0.8.5-beta7. A workaround to get this working for me is to subscribe to the presence and handle each member that way, but I would rather use the GET method of the presence as expected by getting all users. 

Thanks for the help guys!

```
GetOptions options = new GetOptions();
options.WaitForSync = true;

var presence = await AblyHelper.Instance.channelTeam.Presence.GetAsync(options);
List<PresenceMessage> members = presence.ToList();
```

+		this	{Sneek_Desktop.MainWindow}	Sneek_Desktop.MainWindow
+		options	{IO.Ably.Realtime.GetOptions}	IO.Ably.Realtime.GetOptions
[...]

PR #147, opened 2017-10-11
Issue #114 - Should be removing Absent members instead of Present members

jeddawson:
This is to fix an underlying problem with #114. Currently, members that are `Present` get removed from the members dictionary. I believe the correct behavior is to remove `Absent` members.

One concern I have with this is I don't see when/where a member becomes `Absent`. I've yet to receive an `Absent` action on a channel so it's possible that we should be removing members with the last action of `Leave` or possibly (`Leave` or `Absent`) if `Absent` is used in situations that I've yet to observe.

It looks like there are some tests that might cover this, but they are currently set to `Skip`. If you want me to enable them and update this PR, just say the word 👍

### DIFF
--- a/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
+++ b/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
@@ -9,7 +9,24 @@
 
         public static bool IsNewerThan(this PresenceMessage oldMessage, PresenceMessage newMessage)
         {
-            return oldMessage.CompareTo(newMessage) > 0;
+            if (oldMessage.IsSynthesized() || newMessage.IsSynthesized())
+            {
+                if (oldMessage.Timestamp > newMessage.Timestamp) return true;
+            }
+            
+            var thisValues = oldMessage.Id.Split(':');
+            var otherValues = newMessage.Id.Split(':');
+            var msgSerialThis = int.Parse(thisValues[1]);
+            var msgSerialOther = int.Parse(otherValues[1]);
+            var indexThis = int.Parse(thisValues[2]);
+            var indexOther = int.Parse(otherValues[2]);
+
+            if (msgSerialThis == msgSerialOther)
+            {
+                if (indexThis > indexOther) return true;
+            }
+            
+            if (msgSerialThis > msgSerialOther) return true;
         }
     }
 }
--- a/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
+++ b/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
@@ -7,26 +7,26 @@
             return !msg.Id.StartsWith(msg.ConnectionId);
         }
 
-        public static bool IsNewerThan(this PresenceMessage oldMessage, PresenceMessage newMessage)
+        public static bool IsNewerThan(this PresenceMessage thisMessage, PresenceMessage thatMessage)
         {
-            if (oldMessage.IsSynthesized() || newMessage.IsSynthesized())
+            if (thisMessage.IsSynthesized() || thatMessage.IsSynthesized())
             {
-                if (oldMessage.Timestamp > newMessage.Timestamp) return true;
+                return thisMessage.Timestamp > thatMessage.Timestamp;
             }
             
-            var thisValues = oldMessage.Id.Split(':');
-            var otherValues = newMessage.Id.Split(':');
+            var thisValues = thisMessage.Id.Split(':');
+            var thatValues = thatMessage.Id.Split(':');
             var msgSerialThis = int.Parse(thisValues[1]);
-            var msgSerialOther = int.Parse(otherValues[1]);
+            var msgSerialThat = int.Parse(thatValues[1]);
             var indexThis = int.Parse(thisValues[2]);
-            var indexOther = int.Parse(otherValues[2]);
+            var indexThat = int.Parse(thatValues[2]);
 
-            if (msgSerialThis == msgSerialOther)
+            if (msgSerialThis == msgSerialThat)
             {
-                if (indexThis > indexOther) return true;
+                return indexThis > indexThat;
             }
             
-            if (msgSerialThis > msgSerialOther) return true;
+            return msgSerialThis > msgSerialThat;
         }
     }
 }
--- a/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
+++ b/src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
@@ -1,0 +1,15 @@
+﻿namespace IO.Ably
+{
+    internal static class PresenceExtensions
+    {
+        public static bool IsSynthesized(this PresenceMessage msg)
+        {
+            return !msg.Id.StartsWith(msg.ConnectionId);
+        }
+
+        public static bool IsNewerThan(this PresenceMessage oldMessage, PresenceMessage newMessage)
+        {
+            return oldMessage.CompareTo(newMessage) > 0;
+        }
+    }
+}
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -315,10 +315,12 @@ namespace IO.Ably.Realtime
                     residualMembers?.Remove(item.MemberKey);
                 }
 
-                // compare the timestamp of the new item with any existing member (or ABSENT witness)
+                // Compare for newness (RTP2b)
                 PresenceMessage existingItem;
-                if (members.TryGetValue(item.MemberKey, out existingItem) && (item.Timestamp < existingItem.Timestamp))
+                if (members.TryGetValue(item.MemberKey, out existingItem) && CompareForNewness(existingItem, item))
+                {
                     return false;
+                }
 
                 members[item.MemberKey] = item;
 
@@ -356,14 +358,18 @@ namespace IO.Ably.Realtime
                 bool result = true;
 
                 PresenceMessage existingItem;
-                if (members.TryGetValue(item.MemberKey, out existingItem) &&
-                    existingItem.Action == PresenceAction.Absent)
+                if (members.TryGetValue(item.MemberKey, out existingItem) && CompareForNewness(existingItem, item))
                 {
-                    result = false;
+                    //RTP2c
+                    return false;
                 }
 
                 members.TryRemove(item.MemberKey, out PresenceMessage _);
 
+                if (existingItem?.Action == PresenceAction.Absent)
+                {
+                    result = false;
+                }
                 return result;
             }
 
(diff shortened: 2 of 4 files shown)


SimonWoolf:
Thanks @jeddawson , nice spot!

The `absent` action isn't used by the server, only locally: it keeps things in order if a leave event is received in the middle of a sync. See [https://docs.ably.io/client-lib-development-guide/features/#RTP2f](RTP2f) for more information (that's our client lib spec). That item -- and several other realtimepresence spec items -- don't seem to have been implemented yet in ably-dotnet, so the presence behaviour will still be somewhat broken, even with this fix. So for now, if you want to be able to trust the result completely I'd still suggest using `AblyRest`, I'm afraid.

(Of course, if you do feel like diving in and helping out implementing some more presence spec items, you'd be very welcome :smiley: I'd be happy to answer any questions about the spec, and @marto83 can answer any library-specific questions. But obviously no obligation -- either way, hopefully we'll have a version with fully-working presence out for you soon).

jeddawson:
@SimonWoolf that spec doesn't look too overwhelming :D

If I get a chance later this week, I'll dive in and see if I can help make some progress!

jeddawson:
@SimonWoolf I had a little free time and decided to give it an initial stab. If it appears that I'm on the right track, I'll add some tests to ensure things are behaving correctly!

From the spec that you linked me to above, I believe I took care of:
- [RTP2b](https://docs.ably.io/client-lib-development-guide/features/#RTP2b) 
- [RTP2c](https://docs.ably.io/client-lib-development-guide/features/#RTP2c)
- [RTP2d](https://docs.ably.io/client-lib-development-guide/features/#RTP2d)
- [RTP2e](https://docs.ably.io/client-lib-development-guide/features/#RTP2e)
- [RTP2f](https://docs.ably.io/client-lib-development-guide/features/#RTP2f)

RTP2a & RTP2g were already good, so upon merging the entire [RTP2](https://docs.ably.io/client-lib-development-guide/features/#RTP2) spec will be implemented in ably-dotnet 😄

mattheworiordan:
@jeddawson you are a legend, thank you!

@withakay would be good for you to review this PR and comment.

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -315,29 +315,74 @@ public bool Put(PresenceMessage item)
...
 
                 return true;
             }
 
+            private bool CompareForNewness(PresenceMessage oldMsg, PresenceMessage newMsg)
+            {
+                try 
+                {
+                    if(oldMsg.Id.StartsWith(oldMsg.ConnectionId)
+                        && newMsg.Id.StartsWith(newMsg.ConnectionId))
```
SimonWoolf: The sense of this conditional is the wrong way round -- it's only if either of the messages does *not* start with the connection id -- i.e. if either is a synthesized message -- that you fall back to comparing by timestamp. See e.g. https://github.com/ably/ably-js/blob/master/common/lib/client/realtimepresence.js#L463-L477 as another implementation

(The reason is that with a synthesized message you can't compare it properly, by message serial, since synthesized messages have a message serial of 0)

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -315,29 +315,74 @@ public bool Put(PresenceMessage item)
...
+                        item.Action = PresenceAction.Present;
+                        break;
+                }
 
                 members[item.MemberKey] = item;
 
                 return true;
             }
 
+            private bool CompareForNewness(PresenceMessage oldMsg, PresenceMessage newMsg)
```
SimonWoolf: Could the name of this method make it clearer what the sense of the check is? i.e. it that returns true if `newMsg` is actually older than `oldMsg`. Maybe `isOlderThan(item, existing)`?

SimonWoolf:
@jeddawson Very nice, thanks! I can't fully review as I'm not a .NET dev, but made a couple comments re the logic.

withakay:
@jeddawson thanks for the submission, I will take a closer look when I am at my desk later today.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: Nit picking here, but I think it would align better with the wording of the spec if the method name were `IsNewerThan`
And I think using an extension method would also make the intent a bit clearer `item.IsNewThan(existingItem)`

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: Perhaps IsSynthesized() could be added as an extension method, or added to PresenceMessage itself (as the JavaScript library does), it conveys just a little more meaning.

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -315,29 +315,74 @@ public bool Put(PresenceMessage item)
...
+                        var msgSerialOld = int.Parse(oldValues[1]);
+                        var msgSerialNew = int.Parse(newValues[1]);
+                        var indexOld = int.Parse(oldValues[2]);
+                        var indexNew = int.Parse(newValues[2]);
+
+                        return (msgSerialOld == msgSerialNew && indexNew < indexOld)
+                            || msgSerialNew < msgSerialOld;
+                    }
+                } catch {
+                    return false;
```
withakay: It would be good to log this exception when debugging in enabled
`if(Logger.IsDebug) Logger.Debug(ex)`

src/IO.Ably.Shared/Realtime/Presence.cs
jeddawson: Ah yes, that would certainly be more obvious and making it an extension would indeed be a lot cleaner 👍

src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace IO.Ably
+{
+    internal static class PresenceExtensions
+    {
+        public static bool IsSynthesized(this PresenceMessage msg)
+        {
+            return !msg.Id.StartsWith(msg.ConnectionId);
```
jeddawson: @SimonWoolf I think I have this correct now, thanks for the explanation. When the message does not start with the connection id, it is a synthesized message. Correct, or did I get fully confused?

jeddawson:
@SimonWoolf & @withakay thanks for the feedback and pointers and no worries about being picky, that kind of consistency is a good thing 👍 

I *think* the logic is correct now and the naming is much more obvious. I've got some other work that needs my attention, but will definitely get some tests in place as time allows.

src/IO.Ably.Shared/Extensions/PresenceExtensions.cs
withakay: This makes sense to me (and it matches the logic in the JS lib) 👍

SimonWoolf: > When the message does not start with the connection id, it is a synthesized message. Correct

Yup, exactly right.

(What's going on behind the scenes is that when a connection disconnects without detaching from any channels, the server keeps it in the presence set for 15 seconds, so that if it just reconnects again immediately, it can just carry on as before, and other people on the channel aren't deluged with a `leave` followed immediately by an `enter` just because someone's internet cut out for a couple seconds. If after 15 seconds it hasn't reconnecting, the server synthesizes a fake `leave`, pretending from that connection. That fake `leave` has its `connectionId` set to that of the connection (else the memberId wouldn't be right), but just makes up a messageId -- it couldn't generate a real-looking one even if it wanted to, since the serial part of the msgId is set by the client on publishing, the server doesn't keep track of it. So that's why a connectionId not matching the first bit of the msgId means it's synthesized, which also means you can't compare by serial)

jeddawson: @withakay & @SimonWoolf excellent! It's all making sense now 😄 

I've been cranking on a project that is due today, so I haven't returned to get this tested but I'll likely have time either over the weekend or Monday evening. Thanks again for the great feedback!

withakay:
I am thinking we should merge this PR and raise a new issue to add more tests around it, which @jeddawson or myself can pick at some point. Any objections @mattheworiordan?

mattheworiordan:
@withakay well I would recommend we add some more tests now to confirm this does what it should do before we change behaviour and potentially introduce a different type of regression. Can you review what tests exist, are inadequate and are missing perhaps and then propose what needs doing to confidently merge this in as soon as possible?

jeddawson:
Hi guys - sorry I've been MIA! I'm on site with a client this week and things got a lot more involved than anticipated 💀 

I started working on tests last weekend and never finished them up since I think we need to have the `getAsync()` properly handle `waitForSync`. Otherwise this change only partially improves the situation so I started on implementing the `waitForSync` pieces, but haven't finished that up yet.

Also, when I enabled `WhenAClientAttachedToPresenceChannel_ShouldEmitPresentForEachMember` it was very intermittent (sometimes both `JSON` and `MsgPack` would pass, sometimes neither, sometimes one). It seemed to me like there's perhaps an issue with the cleanup between the two test runs, but I need to spend some more time to understand the tests a bit more before my opinion there should count for much!

I'm headed back to the west coast tomorrow, so I'll have some more time to help next week.

mattheworiordan:
@jeddawson if you have the WIP tests perhaps we can pick up from there?  Otherwise, it's something we should be doing so we can make progress on this front.  WDYT @withakay?

withakay:
@mattheworiordan 

> Can you review what tests exist, are inadequate and are missing perhaps and then propose what needs doing to confidently merge this in as soon as possible?

Presence (as noted by @SimonWoolf earlier) is largely not implemented in the dotnet lib, there are only 6 tests around presence (covering parts of RTP1, RTP2 and RTP4) and 2 of those are skipped because they fail (although I now have fix for RTP4 on another branch) so if anyone is using this part of the library they are very unlikely to be getting the behaviour detailed in the spec, as it is.

The changes @jeddawson has made are all covered in the spec by RTP2 and its subsections. The only test tagged as RTP2 currently fails, so I think adding test to cover each section of RTP2 would be the next step.

As a side note, I am not sure what the workflow would look like for this, perhaps we should merge @jeddawson's fork to a feature branch, otherwise I think we will have to submit PRs to his fork until it is merged? I have not been in the situation before so maybe there is a better way, any thoughts?

withakay:
@jeddawson FYI, I began working on tests to cover all of RTP2

https://github.com/ably/ably-dotnet/tree/issue-114

mattheworiordan:
Thanks @withakay. I think your approach of a new branch is good, and once ready, you can do a PR and close this PR in favour of that PR.  That makes the most sense.

jeddawson:
@withakay what do think about me changing the base for this PR over to your new branch? I **think** that would allow this to get merged to your branch, then you can open the PR into master when all the tests pass 👍

withakay:
@jeddawson it is worth a try I think, if it doesn't work we can just go back to creating new PR

withakay:
I am closing in favour of https://github.com/ably/ably-dotnet/pull/157

PR #148, opened 2017-10-12
Use MSGPACK compiler flag to add/remove Protocol.MsgPack

withakay:


mattheworiordan:
@withakay unfortunately a customer Casey has had an issue with this build.  He has an issue connecting, whereas the reconnect works.  I will send you the confidential info showing the problem now for you to review.

withakay:
@mattheworiordan are you sure it is this branch and not what is in master?

mattheworiordan:
@withakay given the customer has not been forthcoming with logs after almost two weeks now, I think we should aim to release this.  WDYT @withakay?

withakay:
@mattheworiordan agreed, but perhaps lead with a beta release and give that a week or 2 before pushing a stable package?

mattheworiordan:
> but perhaps lead with a beta release and give that a week or 2 before pushing a stable package?

Ok, perhaps once we have stabilised presence we can do a stable release. WDYT?

PR #152, opened 2017-10-16
Update release process

withakay:
I have updated the way that versioning is managed.
There is now a GitVersion.yml file that works with GitVersion.exe.

The nuspec files now use a $version$ variable that is passed in from the build script, the build script picks up the versions from GitVersion. 

Now, when creating a package from the master branch the version string will be the plain SerVer 'Maj.Min.Patch' as normal, but on feature branches the version string will contain the branch name and subversion e.g. '0.8.5-feature-branch-a0001', this would become a nuget package called 'ably.io.0.8.5-feature-branch-a0001.nupkg'. 

The reason this is useful is because I wanted to be able to test prerelease builds without polluting the main nuget repository, so I have set up a seperate package repository just for this on myget.org (https://www.myget.org/feed/ably-dotnet-dev/package/nuget/ably.io/). This naming convention makes it easy to keep track of what release version is being used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -195,3 +195,4 @@ FakesAssemblies/
 
 # Visual Studio 6 workspace options file
 *.opt
+secrets/*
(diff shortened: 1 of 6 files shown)


mattheworiordan:
Looks good, 👍

PR #154, opened 2017-10-18
Update to allow building and packaging with Appveyor

withakay:
Updates to the build and package scripts and associated proj files to work with appveyor (and more flexible in general should we want to move to another CI provider).

I have removed the the 'Package' and 'Travis' specific solution files, they are no longer required. There are now 'ci_release' and 'package' build configurations in the main solution which achieve the same thing.

Continuous integration is configurable and is enabled, each build is pushed to myget. This could easily be updated to push to nuget, but until this setup is battle tested I am happy to keep that as a manual step.

Currently the test step is set to skip as there is always one or two failures (still) :(
Everything is setup so the test can be run once the test suite is stable.


mattheworiordan:
Nice 👍

PR #157, opened 2017-11-01
Added presence features covering RTP2 and others

withakay:
This is a continuation of the work done on PR #147 by @jeddawson in response to issue #114, the crux of which was that the library was removing Absent members instead of Present members. Fixing that lead on to implementing some of the missing Presence features.

Using the other library implementations as a guide, in particular ably-java, I have added tests to cover the parts of the spec affected by @jeddawsons original work and fixed some issues that were raised by those tests.

The new parts of the spec covered are:

- All of RTP2a/b/c/d/f/g (note RTP2e is missing at the moment)
- RTP6a/b
- RTP18a/b/c

Additionally WaitOnSync is now implemented. 

I have also fixed up the tests for:

- RTP1
- RTP4

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -315,10 +315,12 @@ namespace IO.Ably.Realtime
                     residualMembers?.Remove(item.MemberKey);
                 }
 
-                // compare the timestamp of the new item with any existing member (or ABSENT witness)
+                // Compare for newness (RTP2b)
                 PresenceMessage existingItem;
-                if (members.TryGetValue(item.MemberKey, out existingItem) && (item.Timestamp < existingItem.Timestamp))
+                if (members.TryGetValue(item.MemberKey, out existingItem) && CompareForNewness(existingItem, item))
+                {
                     return false;
+                }
 
                 members[item.MemberKey] = item;
 
@@ -356,14 +358,18 @@ namespace IO.Ably.Realtime
                 bool result = true;
 
                 PresenceMessage existingItem;
-                if (members.TryGetValue(item.MemberKey, out existingItem) &&
-                    existingItem.Action == PresenceAction.Absent)
+                if (members.TryGetValue(item.MemberKey, out existingItem) && CompareForNewness(existingItem, item))
                 {
-                    result = false;
+                    //RTP2c
+                    return false;
                 }
 
                 members.TryRemove(item.MemberKey, out PresenceMessage _);
 
+                if (existingItem?.Action == PresenceAction.Absent)
+                {
+                    result = false;
+                }
                 return result;
             }
 
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -256,59 +256,68 @@ namespace IO.Ably.Realtime
 
         internal void OnPresence(PresenceMessage[] messages, string syncChannelSerial)
         {
-            string syncCursor = null;
-            if (syncChannelSerial != null)
+            try
             {
-                syncCursor = syncChannelSerial.Substring(syncChannelSerial.IndexOf(':'));
-                if (syncCursor.Length > 1)
-                    Map.StartSync();
-            }
+                string syncCursor = null;
+                if (syncChannelSerial != null)
+                {
+                    syncCursor = syncChannelSerial.Substring(syncChannelSerial.IndexOf(':'));
+                    if (syncCursor.Length > 1)
+                        Map.StartSync();
+                }
             
-            if (syncChannelSerial != null)
-            {
-                int colonPos = syncChannelSerial.IndexOf(':');
-                string serial = colonPos >= 0 ? syncChannelSerial.Substring(0, colonPos) : syncChannelSerial;
-                /* Discard incomplete sync if serial has changed */
-                if (Map.IsSyncInProgress && _currentSyncChannelSerial != null && _currentSyncChannelSerial != serial)
+                if (syncChannelSerial != null)
                 {
-                    /* TODO: For v1.0 we should emit leave messages here. See https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L219 for example. */
+                    int colonPos = syncChannelSerial.IndexOf(':');
+                    string serial = colonPos >= 0 ? syncChannelSerial.Substring(0, colonPos) : syncChannelSerial;
+                    /* Discard incomplete sync if serial has changed */
+                    if (Map.IsSyncInProgress && _currentSyncChannelSerial != null && _currentSyncChannelSerial != serial)
+                    {
+                        /* TODO: For v1.0 we should emit leave messages here. See https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L219 for example. */
+                        EndSync();
+                    }
+
+                    syncCursor = syncChannelSerial.Substring(colonPos);
+                    if (syncCursor.Length > 1)
+                    {
+                        Map.StartSync();
+                        _currentSyncChannelSerial = serial;
+                    }
+                }
+
+                foreach (var message in messages)
+                {
+                    bool updateInternalPresence = message.ConnectionId == _channel.RealtimeClient.Connection.Id;
+                    var broadcast = true;
+                    switch (message.Action)
+                    {
+                        case PresenceAction.Enter:
+                        case PresenceAction.Update:
+                        case PresenceAction.Present:
+                            broadcast &= Map.Put(message);
+                            if (updateInternalPresence)
+                                InternalMap.Put(message);
+                            break;
+                        case PresenceAction.Leave:
+                            broadcast &= Map.Remove(message);
+                            if (updateInternalPresence)
+                                InternalMap.Remove(message);
+                            break;
+                    }
+                    if (broadcast)
+                        Publish(message);
+                }
+                // if this is the last message in a sequence of sync updates, end the sync
+                if ((syncChannelSerial == null) || (syncCursor.Length <= 1))
                     EndSync();
-                }
 
-                syncCursor = syncChannelSerial.Substring(colonPos);
-                if (syncCursor.Length > 1)
-                {
-                    Map.StartSync();
-                    _currentSyncChannelSerial = serial;
-                }
             }
-
-
-            foreach (var message in messages)
+            catch (Exception ex)
             {
-                bool updateInternalPresence = message.ConnectionId == _channel.RealtimeClient.Connection.Id;
-                var broadcast = true;
-                switch (message.Action)
-                {
-                    case PresenceAction.Enter:
-                    case PresenceAction.Update:
-                    case PresenceAction.Present:
-                        broadcast &= Map.Put(message);
-                        if (updateInternalPresence)
-                            InternalMap.Put(message);
-                        break;
-                    case PresenceAction.Leave:
-                        broadcast &= Map.Remove(message);
-                        if (updateInternalPresence)
-                            InternalMap.Remove(message);
-                        break;
-                }
-                if (broadcast)
-                    Publish(message);
+                Logger.Error($"An error occurred processing Presence Messages for channel '{_channel.Name}'. Error: {ex.Message}");
+                throw new AblyException(
+                    new ErrorInfo($"An error occurred processing Presence Messages for channel '{_channel.Name}'. See the InnerException for more details."), ex);
             }
-            // if this is the last message in a sequence of sync updates, end the sync
-            if ((syncChannelSerial == null) || (syncCursor.Length <= 1))
-                EndSync();
         }
 
         internal void EndSync()
@@ -527,7 +536,8 @@ namespace IO.Ably.Realtime
                 }
                 catch (Exception ex)
                 {
-                    if (Logger.IsDebug) Logger.Debug(ex.Message);
+                    Logger.Error($"PresenceMap.Put | Channel: {_channelName}, Error: {ex.Message}");
+                    throw ex;
                 }
 
                 switch (item.Action)
@@ -608,10 +618,8 @@ namespace IO.Ably.Realtime
                 }
                 catch (Exception ex)
                 {
-                    if (Logger.IsDebug)
-                        Logger.Debug($"EndSync | Channel: {_channelName}, Error: {ex.Message}");
-
-                    
+                    Logger.Error($"PresenceMap.EndSync | Channel: {_channelName}, Error: {ex.Message}");
+                    throw ex;
                 }
                 finally
                 {
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -59,9 +59,7 @@ namespace IO.Ably.Realtime
             _channel.InternalStateChanged += OnChannelStateChanged;
             _clientId = cliendId;
         }
-
-        //TODO: Validate the logic is correct
-
+        
         internal bool InternalSyncComplete => (!Map.IsSyncInProgress && SyncComplete);
         internal PresenceMap Map { get; }
         internal PresenceMap InternalMap { get; }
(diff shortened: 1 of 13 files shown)


src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -325,6 +325,32 @@ public bool Put(PresenceMessage item)
...
+                        var msgSerialOld = int.Parse(oldValues[1]);
+                        var msgSerialNew = int.Parse(newValues[1]);
+                        var indexOld = int.Parse(oldValues[2]);
+                        var indexNew = int.Parse(newValues[2]);
+
+                        return (msgSerialOld == msgSerialNew && indexNew < indexOld)
+                            || msgSerialNew < msgSerialOld;
+                    }
+                } catch {
+                    return false;
```
mattheworiordan: We should log the failure here as this is not expected behaviour.

src/IO.Ably.Shared/Types/PresenceMessage.cs
```diff
@@ -51,5 +51,30 @@ public PresenceMessage(PresenceAction action, string clientId, object data)
 
         [JsonIgnore]
         public string MemberKey => $"{ClientId}:{ConnectionId}";
+
+        public int CompareTo(PresenceMessage other)
```
mattheworiordan: Should we not remove the [Presence `Newness` method](https://github.com/ably/ably-dotnet/pull/157/commits/1b32df9b11f7a8875f57bb0f778f835b1d6713fc#diff-927b012d999551505dd3e7c328da24e6R328) completely if this exists?  It's also a lot more logical to have something like `if (presenceMessage.Newer(existingPresenceMessages))`, instead of `CompareTo` no?  That name is not all that logical, unless that's an idiomatic method used by iComparable?  However, even if it is, is one message really greater than another unless we're specific about in what way?  I am concerned making them comparable works now, but later we may want to consider messages being comparable in other ways and this could confuse things.

mattheworiordan: Ah, I see a later commit https://github.com/ably/ably-dotnet/pull/157/commits/c6b7776e200a45b0c0c0fe0bc2bdb767187eb38d largely address this :)

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -61,13 +83,60 @@ public Task<IEnumerable<PresenceMessage>> GetAsync(GetOptions options = null)
...
-                                                (x.ConnectionId == getOptions.ConnectionId)));
+            if (getOptions.WaitForSync)
+                WaitForSync();
+
+            var result = Map.Values.Where( x => (getOptions.ClientId.IsEmpty() || x.ClientId == getOptions.ClientId)
+                                               && (getOptions.ConnectionId.IsEmpty() || x.ConnectionId == getOptions.ConnectionId) );
             return Task.FromResult(result);
         }
 
+        internal Task<IEnumerable<PresenceMessage>> GetAsync(string clientId, bool waitForSync = false)
```
mattheworiordan: This should default to true, both in 0.8 https://docs.ably.io/client-lib-development-guide/versions/v0.8/features/#RTP11c1, and 1.0 https://docs.ably.io/client-lib-development-guide/features/#RTP11c1

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -61,13 +83,60 @@ public Task<IEnumerable<PresenceMessage>> GetAsync(GetOptions options = null)
...
+                    /* assignment (=) is intended */
+                   !(syncIsComplete = InternalSyncComplete))
+            {
+                Task.Delay(10);
+            }
+
+            // to be completed when extra states in 1.0 spec are implemented 
+            if (!syncIsComplete)
+            {
+      //          /* invalid channel state */
```
mattheworiordan: Why are we committing commented code, and what's the thinking here anyway?

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -61,13 +83,60 @@ public Task<IEnumerable<PresenceMessage>> GetAsync(GetOptions options = null)
...
+        }
+
+        private void WaitForSync()
+        {
+            bool syncIsComplete = false;
+            while ((_channel.State == ChannelState.Attached || _channel.State == ChannelState.Attaching) &&
+                    /* assignment (=) is intended */
+                   !(syncIsComplete = InternalSyncComplete))
+            {
+                Task.Delay(10);
```
mattheworiordan: Is that 10 seconds or 10 milliseconds?

mattheworiordan: Is there not a better way to handle this without a loop i.e. can you not simply subscribe once for Attached or Attaching states?

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -155,32 +224,101 @@ internal Task UpdatePresenceAsync(PresenceMessage msg)
...
             }
+            
+            if (syncChannelSerial != null)
+            {
+                int colonPos = syncChannelSerial.IndexOf(':');
+                string serial = colonPos >= 0 ? syncChannelSerial.Substring(0, colonPos) : syncChannelSerial;
+                /* Discard incomplete sync if serial has changed */
+                if (Map.IsSyncInProgress && _currentSyncChannelSerial != null && _currentSyncChannelSerial != serial)
+                {
+                    /* TODO: should emit leave messages here? */
```
mattheworiordan: I don't think you'd emit anything until the new sync has completed surely, and then work out which members have gone or entered no?

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -212,10 +350,25 @@ private void OnChannelStateChanged(object sender, ChannelStateChange e)
         {
             if (e.Current == ChannelState.Attached)
             {
+                /* Start sync, if hasPresence is not set end sync immediately dropping all the current presence members */
+                Map.StartSync();
+                _syncAsResultOfAttach = true;
+                //if (!hasPresence)
+                //{
```
mattheworiordan: Not sure why we're committing commented code again?  Why is this commented out?

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -312,79 +467,115 @@ public bool Put(PresenceMessage item)
...
+                try
+                {
+                    if (_members.TryGetValue(item.MemberKey, out var existingItem) && existingItem.IsNewerThan(item))
+                    {
+                        return false;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (Logger.IsDebug) Logger.Debug(ex.Message);
```
mattheworiordan: Should this not be error level at least given this shouldn't happen right?  As far as I am concerned, a failure like this should probably even result in the channel becoming detached because we have an unknown fault and as such can no longer guarantee the state of the channel is correct.  Do you not think this is a channel failure?

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -312,79 +467,115 @@ public bool Put(PresenceMessage item)
...
+                            _residualMembers = null;
                         }
+                        IsSyncInProgress = false;
                     }
+                    
+                }
+                catch (Exception ex)
+                {
+                    if (Logger.IsDebug)
+                        Logger.Debug($"EndSync | Channel: {_channelName}, Error: {ex.Message}");
```
mattheworiordan: Again, this is an unexpected fault, we shouldn't be hiding this in debug messages, and we should be considering this a significant enough problem that we should fail the channel with an error

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: I didn't realise that it was in the spec, I will fix this.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: @mattheworiordan That is 10 milliseconds. 
The [Java version](https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L789) uses a while loop, which I based this on. 
I will look at alternatives

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: I added this when porting from [Java](https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L794) but realised that ChannelState.Suspended was not implemented yet.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: The same excuse as before, this is functionality that will need to be brought over from 1.0. Would you like me to strip them out?

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: I will remove this comment, it is not relevant for 0.8, but just to give you some background, I left that there to remind myself to double check because in the Java code there is a method called [endSyncAndEmitLeaves](https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L219) that gets called.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: I didn't actually add this code, it comes from the original PR. 
That said, I am not convinced there should be a try/catch block here at all, the use of TryGetValue means only place where an exception can bubble up here is from the call to existingItem.IsNewerThan which parses the presence message id.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: As before I didn't add this and can't be sure of the motivation behind it (useful to the developer when initially writing the code perhaps?), but I don't think the try/catch is required here at all. WDYT?

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: @withakay not a biggie, just seems unnecessary to effectively poll for a result when we can subscribe to state changes.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: Ah, maybe add a TODO comment then explaining.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: Yup, makes sense, perhaps TODO it.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: Ok, TODO it to address when working on 1.0

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: Well I think if there is a risk that we can crash people's apps, then it would make sense to catch it, change state to failed, and make sure we log that error.  Generally that would be my preference.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: Well not sure if it's needed, but so long as we don't gloss over errors, show the error, and fail the channel

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: I think the approach in the [Java version](https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L901) make sense, the responsibility is placed with `hasNewerItem` method. The [calling code](https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java#L853) has no exception handling. I do think we should add some more test coverage to check the handling of malformed id strings though.

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: @mattheworiordan I have updated WaitForSync to be async (and renamed to WaitForSyncAsync as is idiomatic in .net) and internally run off events rather than the ugly while loop.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: @withakay I spoke with @SimonWoolf and we both agree that the Java implementation is certainly wrong.  Silently failing when doing a compare means we have a malformed message and thus the entire protocol at that point is compromised.  The only thing at that point we should do is raise an exception, log the problem, and move the channel to the failed state.  I am going to raise an issue in Java to correct this.   Silently catching exceptions in almost all cases is wrong.

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: See https://github.com/ably/ably-java/issues/357

src/IO.Ably.Shared/Realtime/Presence.cs
mattheworiordan: See https://github.com/ably/ably-java/issues/357

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -76,39 +83,56 @@ public void Dispose()
...
-            // TODO: This code was added when porting from the Java lib but can't completed be until the extra states added in 1.0 spec (specifically ChannelState.Suspended) are implemented 
+            // if the channel state changes and is not Attached or Attaching then we should exit
+            _channel.StateChanged += (sender, args) =>
+            {
+                if (_channel.State != ChannelState.Attached && _channel.State != ChannelState.Attaching)
+                    tsc.SetResult(false);
+            };
+
+            InitialSyncCompleted += (sender, args) => CheckAndSet();
+            Map.SyncNoLongerInProgress += (sender, args) => CheckAndSet();
```
mattheworiordan: @withakay not sure how `TaskCompletionSource` behaves, but as far as I can tell, `CheckAndSet` could in theory be called twice, and as a result, trigger `tsc.SetResult(true)` after `tsc.SetResult(false)` or vice versa.  Does `TaskCompletionSource` ensure that subsequent calls are ignored?

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: @mattheworiordan OK, understood. At that point in the code the exception would be in the PresenceMap, so I will probably want move the exception handling up into the Presence class in order to access the Channel to move it into a failed state.

mattheworiordan: Yup, makes sense, I'll trust you to work out how to propagate up! 👍

src/IO.Ably.Shared/Realtime/Presence.cs
withakay: Good question. The short answer is No, the safe way to handle this is to use TrySetResult(). I will update the code.

withakay: @mattheworiordan I have updated this and made some other changes. I have added a manual call to CheckAndSet to handle the case where the desired state might already have been reached. I have also changed the event handlers from anonymous methods to local methods and unsubscribed them from events before exiting the method.

jeddawson:
Hey guys! Sorry for my complete disappearing act, but work was kinda crazy for a few weeks there. Now that the holidays are coming I'm hopeful I'll be back to normal and have some downtime for helping out if there's anywhere I can contribute.

BTW, @withakay this all looks really fantastic!

mattheworiordan:
@withakay this looks good.  Quick question, in this PR you now have a few places where an exception is thrown, but it seems you only catch exceptions in one place and move to the failed state.  Is this because all exceptions bubble up to that point, or are there places where exceptions (such as presence being invalid) will not result in a channel failure?

withakay:
@mattheworiordan 

>it seems you only catch exceptions in one place and move to the failed state. Is this because all exceptions bubble up to that point

Yes.

Presence messages are handled via the OnPresence method - an exception caught there will fail the channel. All of the exception raising that I have added is in methods that are directly or indirectly called from within the OnPresence method only and only from there. 
I am confident I am not leaking any exceptions anywhere else.

mattheworiordan:
I am confident I am not leaking any exceptions anywhere else. 

👍

mattheworiordan:
Can we merge this then?

withakay:
@mattheworiordan yes, I was just waiting for you to approve it :)

PR #158, opened 2017-11-01
Test improvements

withakay:
This PR contains updates and fixes that were done whilst working on PR #157, but that were not directly related to that work, so rather than pollute that PR I have extracted that work here. I would suggest that PR #157 is reviewed before dealing with this one.

Essentially some helpers and improvements to make some test more reliable along with a bug fix.

The headline changes are 

1. I have changed some accessors from private to internal, this allows properties to be accessed from the tests without exposing them publicly.

2. I fixed a bug where the Logger would not be set by from the ClientOptions - This would have no affect on customers as we don't allow them to set the logger in ClientOptions at the moment, but the behaviour was incorrect, as revealed by a failing test.


mattheworiordan:
@withakay this PR is filled with commits from #157.  Could we not base this on 157 so that I can review this PR without those commits?

withakay:
@mattheworiordan It is based on branch 'issue-114-pr', which is the branch for #157. Not sure why it is doubling up the commits, the changed files are as I would expect though.

mattheworiordan:
> Not sure why it is doubling up the commits, the changed files are as I would expect though.

Because the commits are different.  Can you rebase this on 114 again and force push?  For example, first commit in 114 is e9bf989, yet the matching commit in this repo is ee082f7

withakay:
Moved to https://github.com/ably/ably-dotnet/pull/160

withakay:
@mattheworiordan I have created a new PR #160

PR #160, opened 2017-11-06
Test improvements

withakay:
This PR contains updates and fixes that were done whilst working on PR #157, but that were not directly related to that work, so rather than pollute that PR I have extracted that work here. I would suggest that PR #157 is reviewed before dealing with this one.

Essentially some helpers and improvements to make some test more reliable along with a bug fix.

The headline changes are 

1. I have changed some accessors from private to internal, this allows properties to be accessed from the tests without exposing them publicly.

2. I fixed a bug where the Logger would not be set by from the ClientOptions - This would have no affect on customers as we don't allow them to set the logger in ClientOptions at the moment, but the behaviour was incorrect, as revealed by a failing test.

### DIFF
--- a/src/IO.Ably.Tests/Realtime/ChannelSpecs.cs
+++ b/src/IO.Ably.Tests/Realtime/ChannelSpecs.cs
@@ -522,7 +522,7 @@ namespace IO.Ably.Tests.Realtime
                 Assert.True(called);
             }
 
-            [Fact]
+            [Retry] //replaces fact
             [Trait("spec", "RTL5e")]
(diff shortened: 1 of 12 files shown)


src/IO.Ably.Tests/Realtime/ChannelSpecs.cs
```diff
@@ -838,7 +838,7 @@ public void ShouldImplicitlyAttachAChannel()
                 channel.State.Should().Be(ChannelState.Attaching);
             }
 
-            [Fact]
+            [Retry(3)]
```
mattheworiordan: Nice :)

PR #173, opened 2018-01-23
Shared Tests

withakay:
** Note, This is essentially a house keeping PR, there are no new features implemented. **

I have moved the test code to a shared project which is imported into 2 test projects that target different version of the .NET runtime. 
This allows the tests to run against the traditional .NET Framework on Windows and also against .NET Core on Windows, macOS and Linux (I have tested on macOS and Windows).
This is a step towards having on device (or at least on emulator) testing.
I have also rationalised some of the naming. For example 'Net45' was used in project name that only targeted .NET 4.6.1 or higher, I have replaced this in favour of 'Framework' which is more consistent with the (now rather confusing) MS naming.


mattheworiordan:
Thanks.

1. Why is b924778 and b5c18c9 part of this PR?
1. I don't see any tests results in this PR.  Did Travis not run the tests now for some reason?

withakay:
@mattheworiordan 
1) I failed to push those changes that were committed to master to the remote after I published 0.8.7. I then forked from my local master to created the shared-tests branch. 
I have now pushed the changes to master but the PR has not updated update, so I will create a new PR.

2) The Travis tests always failed so they didn't seem at all useful, after discussing with Martin I disabled them.

withakay:
closed in favour of #174

PR #174, opened 2018-01-24
Use a shared project for tests to facilitate cross platform testing.

withakay:
I have moved the test code to a shared project which is imported into 2 test projects that target different version of the .NET runtime.
This allows the tests to run against the traditional .NET Framework on Windows and also against .NET Core on Windows, macOS and Linux (I have tested on macOS and Windows).
This is a step towards having on device (or at least on emulator) testing.
I have also rationalised some of the naming. For example 'Net45' was used in project name that only targeted .NET 4.6.1 or higher, I have replaced this in favour of 'Framework' which is more consistent with the (now rather confusing) MS naming.


withakay:
This PR replaced #173

funkyboy:
@withakay fine by me.

PR #176, opened 2018-01-26
Circleci

withakay:
Added script to retry failed tests on linux/unix systems. 
Minor update to some test to make them more robust

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,5 +10,6 @@ jobs:
     steps:
       - checkout
       - run: git submodule update --init --recursive --remote
-      - run: dotnet test src/IO.Ably.Tests.DotNetCore20/IO.Ably.Tests.DotNetCore20.csproj
+      - run: chmod +x .circleci/run_tests.sh
+      - run: cd .circleci && ./run_tests.sh
 
--- a/.circleci/run_tests.sh
+++ b/.circleci/run_tests.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+
+TEST_RESULT="passed"
+COUNTER=0
+
+#=== start functions ===
+
+# runs all tests and logs the run for later use
+function run_all_tests {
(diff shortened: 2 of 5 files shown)


withakay:
As you can see circle CI has picked up the new script and ran the tests. The was 1 failure which was re-run, at which point it passed. 
I know this a bit hacky, but I think at the moment this the most time/cost effective and pragmatic solution.
@funkyboy I know you suggested getting this working on Windows first and I will look at that again, but for various reasons this was easier to do (I am happy to explain why sometime).

funkyboy:
@withakay LGTM.
Should we worry about this bit? 
`/ably-dotnet/src/IO.Ably.Tests.DotNetCore20/IO.Ably.Tests.DotNetCore20.csproj : warning NU1701: Package 'XunitRetry 2.0.0' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.`

withakay:
> Should we worry about this bit?

@funkyboy No, it is not an issue here, it runs fine. Generally assemblies built to target .NETFramework v4.6.1 are compatible with .NET Standard 2 (in actual fact .NET Standard was derived from .NET Framework 4.6.1s APIs so it is vice versa really). There are some cases (COM interop PInvoke spring to mind) where there are incompatibilities and MS don't want to support 15 years of legacy code, so they throw up this warning. 
It would be fairly straight forward to update the XunitRetry project and make this warning go away, maybe one day when I have some free time I will do a PR for that :)

mattheworiordan:
Nice, this looks great.  @withakay out of interest, why did you need to move from Travis to Circle?  I don't mind, just interested.

mattheworiordan:
BTW. Nice to see some 🍏

withakay:
> why did you need to move from Travis to Circle?

Two reasons:
1) I have in mind that I might reinstate Travis running a build/test against mono again down the line.  
2) I have not used Travis for a while (and I also do not have access to the Ably Travis account), but have used a circle a lot, so lower friction for me to setup.

Issue #171, opened 2018-01-03
v.87 The current state [Suspended] does not allow messages to be sent.; Code: 500


caseyniemann:
App crashes coming out of suspend after .87 update.  IO.Ably.AblyExceptionReason: The current state [Suspended] does not allow messages to be sent.; Code: 500
Raw
  at IO.Ably.Transport.ConnectionManager.Send (IO.Ably.Types.ProtocolMessage message, System.Action`2[T1,T2] callback, IO.Ably.ChannelOptions channelOptions) [0x00182] in <762ba4d2da794608b090b081586466af>:0 
  at IO.Ably.Realtime.RealtimeChannel.SendMessage (IO.Ably.Types.ProtocolMessage protocolMessage, System.Action`2[T1,T2] callback) [0x00044] in <762ba4d2da794608b090b081586466af>:0 
  at IO.Ably.Realtime.RealtimeChannel.HandleStateChange (IO.Ably.Realtime.ChannelState state, IO.Ably.ErrorInfo error, IO.Ably.Types.ProtocolMessage protocolMessage) [0x000a8] in <762ba4d2da794608b090b081586466af>:0 
  at IO.Ably.Realtime.RealtimeChannel.SetChannelState (IO.Ably.Realtime.ChannelState state, IO.Ably.ErrorInfo error, IO.Ably.Types.ProtocolMessage protocolMessage) [0x00093] in <762ba4d2da794608b090b081586466af>:0 
  at IO.Ably.Realtime.RealtimeChannel.Attach (System.Action`2[T1,T2] callback) [0x00081] in <762ba4d2da794608b090b081586466af>:0 
  at IO.Ably.Realtime.RealtimeChannel.Subscribe (System.Action`1[T] handler) [0x0001d] in <762ba4d2da794608b090b081586466af>:0 
  at AgriSync.Services.AblySignaling+<Subscribe>d__27.MoveNext () [0x0005a] in <33d2a30e32344c1ebb9b0210955b33cf>:0

mattheworiordan:
Hi @caseyniemann, what do you mean the app crashes?  Are you receiving an error stating you cannot send messages in the suspended state, or is the app completely exiting as a result?

caseyniemann:
@mattheworiordan after the app is idle and you come out of suspend, we get a crash....likely due to disconnect/reconnect which ably handles internally...the app does not have to completely exit....just need you to put in a brief delay upon resume.

caseyniemann:
[...]

PR #177, opened 2018-02-02
Null check WebSocket Client

withakay:
Fix for the null reference errors mentioned in #171

### DIFF
--- a/src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
+++ b/src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
@@ -80,7 +80,7 @@ namespace IO.Ably.Transport
                     Logger.Debug("Closing websocket. Close status: " +
                                  Enum.GetName(typeof(WebSocketCloseStatus), ClientWebSocket.CloseStatus) + ", Description: " + ClientWebSocket.CloseStatusDescription);
 
-                if(ClientWebSocket.State != WebSocketState.Closed)
+                if(ClientWebSocket?.State != WebSocketState.Closed)
                 {
                     await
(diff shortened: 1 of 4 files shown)


src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
```diff
@@ -113,20 +113,20 @@ public void SendData(byte[] data)
...
-            if (ClientWebSocket.State == WebSocketState.Open)
+            if (ClientWebSocket?.State == WebSocketState.Open)
             {
                 var sendTask = ClientWebSocket.SendAsync(data, type, true, token);
                 sendTask.ConfigureAwait(false);
                 return sendTask;
             }
 
-            Logger.Warning($"Trying to send message of type {type} when the socket is already closed. Ack for this message will fail shortly.");
+            Logger.Warning($"Trying to send message of type {type} when the socket is already closed or null. Ack for this message will fail shortly.");
```
mattheworiordan: Not sure about that error message.  What does it meant to a user if a socket is `null`?

withakay: Having given this some more thought, I can't think of a legitimate reason for the ClientWebSocket to be null here, so I think the thing to do is set the Connection State to Failed and raise an exception. WDYT?

withakay:
@mattheworiordan I have updated the code to set an Error state that propagates up to the ConnectionManager, causing a Disconnected state to be set. If auto connect is set then the connection reconnects.
I have made some small adjustments to allow for testing and added a test to cover this. Let me know what you think?

mattheworiordan:
Thanks, LGTM, 👍

PR #178, opened 2018-02-08
v1.0 Features: RSC7b, RSA4a, RSA4b, RSA4c1, RSA4c3

withakay:
EDIT: **This PR became unwieldy, so work has proceeded on other branches (forked from from the same branch as this PR). As such this PR should be left until other v1.0 related PRs are merged**

v1.0 Spec items  

* RSC7b
* RSA4a
* RSA4b
* RSA4c1
* RSA4c3


src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -163,48 +163,65 @@ private void SetAuthMethod()
...
-                if (callbackResult is TokenRequest || callbackResult is string)
+                    if (callbackResult is TokenRequest || callbackResult is string)
+                    {
+                        postData = GetTokenRequest(callbackResult);
+                        request.Url = $"/keys/{postData.KeyName}/requestToken";
+                    }
+                    else
+                    {
+                        shouldCatch = false;
+                        throw new AblyException($"AuthCallback returned an unsupported type ({callbackResult.GetType()}. Expected either TokenDetails or TokenRequest", 80019);
```
funkyboy: This is a good and very explanatory message 👍

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -163,48 +163,65 @@ private void SetAuthMethod()
...
+                        return new TokenDetails(response.TextResponse, Now);
 
-                var signedData = response.TextResponse;
-                var jData = JObject.Parse(signedData);
+                    var signedData = response.TextResponse;
+                    var jData = JObject.Parse(signedData);
 
-                if (TokenDetails.IsToken(jData))
-                    return JsonHelper.DeserializeObject<TokenDetails>(jData);
+                    if (TokenDetails.IsToken(jData))
```
funkyboy: I don't know if this is common practice in C#, but I shiver a bit when I don't see `{}`, even in simple statements like this. Remember Heartbleed? :)

src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
```diff
@@ -38,30 +38,32 @@ public void WithSecureFalse_CreatesNonSecureRestUrlsWithDefaultRestHost()
...
-            var fileVersion = typeof(Defaults).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>().Version;
+            var values = handler.LastRequest.Headers.GetValues("X-Ably-Version").ToArray();
             values.Should().NotBeEmpty();
-            values.First().Should().Be("dotnet-" + fileVersion);
+            values.First().Should().Be(Defaults.ProtocolVersion);
         }
 
         [Fact]
-        public async Task WhenCallingUrl_AddsDefaultAblyLibraryVersionHeader()
+        [Trait("spec", "RSC7b")]
```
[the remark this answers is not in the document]
withakay: No, it should not!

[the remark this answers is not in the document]
funkyboy: I usually see

```
[Fact]
[Trait ... ]
- public ...
```

so this feels off :)

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -163,48 +163,65 @@ private void SetAuthMethod()
...
 
-                postData = JsonHelper.Deserialize<TokenRequest>(signedData);
+                    postData = JsonHelper.Deserialize<TokenRequest>(signedData);
 
-                request.Url = $"/keys/{postData.KeyName}/requestToken";
+                    request.Url = $"/keys/{postData.KeyName}/requestToken";
+                }
+                catch (AblyException ex)
+                {
+                    throw new AblyException(new ErrorInfo("Error calling Auth URL, token request failed. See inner exception for details.", 80019, ex.ErrorInfo.StatusCode), ex);
```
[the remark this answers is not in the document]
withakay: I think so, the System.Exception class has an InnerException property and every exception derives from that class. I could be more explicit: "See the InnerException property for details of the underlying exception."  Something like that?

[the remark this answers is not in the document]
funkyboy: "See the InnerException property for details" is then crystal clear to me :)

src/IO.Ably.Shared/AblyAuth.cs
withakay: I *think* it is fairly standard practice, certainly it is widely used in this library by the other developers. I think because the defacto standard in c# is to put `{` & `}` on new lines, this gets used a lot to make code more terse. It is pretty hard to do by accident with Visual Studio (or the JetBrains c# IDE) as they recognise the construct and tab out when you hit return, but I do see your point  ¯\_(ツ)_/¯

src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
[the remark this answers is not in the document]
withakay: Sorry, my coffee had not kicked in and I miss read your comment, I meant to agree with you!

Technically the [Trait...] can be there, it is just acting as a tag, so it doesn't break anything, but it was not my intention to do it that way. I will fix this.

//facepalm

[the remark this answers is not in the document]
withakay: @funkyboy can we start this one again? :)

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -163,48 +163,65 @@ private void SetAuthMethod()
...
+                    var response = await CallAuthUrl(mergedOptions, @params);
 
-                if (response.Type == ResponseType.Text) //Return token string
-                    return new TokenDetails(response.TextResponse, Now);
+                    if (response.Type == ResponseType.Text) //Return token string
+                        return new TokenDetails(response.TextResponse, Now);
 
-                var signedData = response.TextResponse;
-                var jData = JObject.Parse(signedData);
+                    var signedData = response.TextResponse;
```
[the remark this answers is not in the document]
withakay: I am not sure what a linter would done for the above code, it has been indented because it is in a try/catch block, so it has been correctly indented. 
That said, there is no well supported command line linter/code formatting tool for c# that I know of yet, although I am aware there are things in development for dotnet core. There is FxCop, which does static analysis and can be used to enforce style rules, but it does not actually fix issues automatically. There was previously some talk of integrating FxCop, but the problem is that because the fixes are not automated, there would, I expect, be a lot of manual editing required (or, I suppose, less strict rules could be enforced) to get the FxCop rules to pass. Given that the code is generally pretty clean and the low number of people that work on it, it has not felt like a priority. FWIW I use Resharper which has a well adopted default style and highlights inconsistencies, so I try to stick to its rules.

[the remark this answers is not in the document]
funkyboy: @withakay Whatever is simpler. For example in Ruby projects we use [Rubocop](https://github.com/bbatsov/rubocop) and in JS we will [add a linter soon](https://github.com/ably/ably-js/issues/441). 
It could even be a script that you run before committing. 
The important part is to avoid unneeded/irrelevant red or green diffs when possible.

withakay:
@paddybyers I seem to have borked a rebase from master, sorry about that. I will get it cleaned up and resubmit the PR

src/IO.Ably.Android/IO.Ably.Android.csproj
```diff
@@ -82,6 +82,7 @@
     <Compile Include="..\CommonAssemblyInfo.cs">
       <Link>Properties\CommonAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="Platform.cs" />
```
paddybyers: Why is this change (and the other project/build-related) changes part of this PR?

src/IO.Ably.Android/Platform.cs
```diff
@@ -0,0 +1,10 @@
+﻿using IO.Ably.Transport;
```
paddybyers: Again, this doesn't seem to belong in this PR.

src/IO.Ably.Shared/AblyAuth.cs
paddybyers: I used to be OK with this but now I avoid it. I think a linter is a good idea because it settles questions like this. Although you don't have to stick to the linter's defaults, you have to have a good reason to change them.

src/IO.Ably.Shared/AblyException.cs
```diff
@@ -46,6 +46,16 @@ public AblyException(Exception ex)
 
             }
 
+        /// <summary>
+        /// Creates an AblyException and populates ErrorInfo with the supplied parameters.
+        /// </summary>
+        /// <param name="reason"></param>
+        /// <param name="code"></param>
+        public AblyException(string reason, int code) : this(new ErrorInfo(reason, code))
```
paddybyers: Why is a `statusCode` not part of the signature here?

src/IO.Ably.Shared/Http/AblyHttpClient.cs
```diff
@@ -206,7 +208,7 @@ internal bool IsRetryableResponse(HttpResponseMessage response)
         {
             return ErrorInfo.IsRetryableStatusCode(response.StatusCode);
         }
-
+
```
paddybyers: Trailing whitespace. Does your IDE not eliminate these things?

src/IO.Ably.Shared/Http/AblyHttpClient.cs
```diff
@@ -302,8 +304,5 @@ private string GetQuery(AblyRequest request)
                 return "?" + query;
             return string.Empty;
         }
-
     }
-
-
 }
```
paddybyers: Here and elsewhere there are files with no trailing linefeed. Can you fix those please?

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -26,6 +27,180 @@ static TokenParams CreateTokenParams(Capability capability, TimeSpan? ttl = null
...
+            catch (AblyException e)
+            {
+                e.ErrorInfo.Code.Should().Be(80019);
+            }
+            // check the retry code was run
+            testLogger.MessageSeen.Should().BeTrue();
+            // only once
+            testLogger.SeenCount.Should().Be(1);
+        }
+
```
paddybyers: Is 2 blank lines intended here? It's not consistent.

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
```diff
@@ -186,8 +249,8 @@ public TestAblyAuth(ClientOptions options, AblyRest rest) : base(options, rest)
             }
         }
 
-        public AuthoriseTests(ITestOutputHelper output) : base(output)
-        {
-        }
+
```
paddybyers: 2 blank lines

src/IO.Ably.Tests.Shared/Infrastructure/AblySandboxFixture.cs
```diff
@@ -8,7 +8,7 @@ namespace IO.Ably.Tests
 {
     public class AblySandboxFixture : IDisposable
     {
-        public readonly static DateTimeOffset StartInterval = DateHelper.CreateDate(DateTimeOffset.UtcNow.Year - 1, 2, 3, 15, 5);
+        public static readonly DateTimeOffset StartInterval = DateHelper.CreateDate(DateTimeOffset.UtcNow.Year - 1, 2, 3, 15, 5);
```
paddybyers: Just curious, but was this a change you made or something the IDE did on your behalf?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs
```diff
@@ -123,8 +123,8 @@ public void ShouldSetTransportLibVersionParamater()
         {
             var client = GetClientWithFakeTransport();
             LastCreatedTransport.Parameters.GetParams().Should().ContainKey("lib");
-            var v = LastCreatedTransport.Parameters.GetParams()["lib"];
-            Regex.Match(v, @"^dotnet-0.8.(\d)$").Success.ShouldBeEquivalentTo(true);
+            var p = LastCreatedTransport.Parameters.GetParams();
```
paddybyers: Can we not have a more meaningful identifier?

src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
```diff
@@ -12,6 +12,8 @@ public class RestSandBoxSpecses : SandboxSpecs
     {
         public RestSandBoxSpecses(AblySandboxFixture fixture, ITestOutputHelper output) : base(fixture, output) { }
 
+
```
paddybyers: 2 lines

src/IO.Ably.Shared/AblyAuth.cs
withakay: FWIW I have added [an issue to look at adding a linter/code formatting](https://github.com/ably/ably-dotnet/issues/179) as either a recommit or build step, at the moment Visual Studio does that job and I am using the MS dotnet teams settings, which would let the above pass. I am happy to change the above, but more generally there are lot of places this form is used, should there be a new issue created to change those?

src/IO.Ably.Android/Platform.cs
withakay: This is for RSC7b, also as per this comment in the spec spreadsheet "It would be nice to know the platform so perhaps set the variant automatically such as xamario-ios, xamaran-android"

`Platform.cs` is providing the optional variant in `X-Ably-Lib: [lib][.optional variant]?-[version]`

src/IO.Ably.Shared/AblyException.cs
withakay: The constructor is overloaded, there is a signature with the http status code here: https://github.com/ably/ably-dotnet/blob/3a0215ac4a78ac9827821b7dac22af0007cf73e3/src/IO.Ably.Shared/AblyException.cs#L62

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
```diff
@@ -1,85 +1,123 @@
...
+            var client = GetRestClient(null, opts => { opts.TokenDetails = dummyTokenDetails; });
+            // check AuthOptions.Force is false (the default)
+            client.Options.Force.Should().BeFalse();
+            // get the current token
+            var newTokenDetails = client.AblyAuth.CurrentToken;
+            // new token should match the dummy token
+            newTokenDetails.Should().BeSameAs(dummyTokenDetails);
+            // authorise again
+            var sameTokenDetails = await client.Auth.AuthorizeAsync();
+            // the same token should be returned
```
mattheworiordan: Am I being stupid here?  Shouldn't calling authorize again explicitly create a new token?  The spec above states "Instructs the library to create a token immediately and ensures Token Auth"

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
```diff
@@ -1,85 +1,123 @@
...
+        [Trait("spec", "RSA10a")]
+        public async Task Authorize_WithNotExpiredCurrentTokenAndForceTrue_ReturnsNewToken()
+        {
+            // create a fake token that has not expired
+            var dummyTokenDetails = new TokenDetails() { Expires = TestHelpers.Now().AddHours(1) };
+            // create new reset client using the dummyTokenDetails
+            var client = GetRestClient(null, opts =>
+            {
+                opts.TokenDetails = dummyTokenDetails;
+                opts.Force = true;
```
mattheworiordan: It seems odd we're implementing v1.0 auth features that assume the force attribute is now gone, yet here it exists.  Would it not be better to couple the changes of removing the `force` attribute completely into this?

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
```diff
@@ -153,20 +191,45 @@ public void TokenShouldNotBeSetBeforeAuthoriseIsCalled()
...
+            await testAblyAuth.AuthorizeAsync(customTokenParams, customAuthOptions);
 
             testAblyAuth.RequestTokenCalled.Should().BeTrue("Token creation was not delegated to RequestToken");
             testAblyAuth.LastRequestTokenParams.Should().BeSameAs(customTokenParams);
             testAblyAuth.LastRequestAuthOptions.Should().BeSameAs(customAuthOptions);
         }
 
+        [Fact]
+        [Trait("spec", "RSA10l")]
+        public void Authorize_RestClientAuthorizeMethodsShouldBeMarkedObsolete()
```
mattheworiordan: Will an obsolete method call result in something being logged?  How is the user shown the "deprecation warning"?

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
withakay: > Would it not be better to couple the changes of removing the force attribute completely into this?

Probably, I still don't know the spec like you do! Can you point me to relevant parts of the spec?

mattheworiordan: > Can you point me to relevant...

See https://github.com/ably/docs/blob/master/content/client-lib-development-guide/versions/features-0-8__1-0.diff, specifically search for `force` and notice where it's been removed.  It makes sense to remove the `force` attribute as part of this PR.

src/IO.Ably.Tests.Shared/AuthTests/AuthorizeTests.cs
mattheworiordan: Yes, I would do so as this would be more consistent with our libs.

withakay:
@mattheworiordan I have removed the Force option and made changes necessary to support that. Please let me know what you think?

@paddy I think I have addressed the issues you raised here in other branches (Style/Formatting issues and also https://github.com/ably/ably-dotnet/pull/186)

paddy:
it'd be cool if you would mention @paddybyers and not me =)

src/IO.Ably.Tests.Shared/Realtime/RealtimeSandboxSpecs.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                throw e;
+            }
+            caught.Should().BeTrue();
+        }
+
+
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSA4c")]
```
mattheworiordan: I don't think this test is really testing what we want here.  You are explicitly calling `RequestToken` in this test, and thus the behaviour of a failed auth attempt is affecting how that method behaves.  Instead, what we are looking to test is the implicit authentication the library performs when a token is needed by the client lib, and the auth means fails.  In the case of REST libs, that often results in an immediate exception, whereas in the case of Realtime, it mostly results in a state change and no error being emitted from the call that implicit triggers an auth.

In addition to this, there are many requirements within RSA4c1,2,3 that are not tested here, and the five or so permutations listed in `RSA4c` of different failure types too are not tested.  Whilst this test may not be intended to be exhaustive, I would like to at least then see the other cases described and marked as pending so that it's possible to look at a commit and know what functionality it will cover now and in the future when the necessary work is done on the client implementation side.

src/IO.Ably.Shared/AblyAuth.cs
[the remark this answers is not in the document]
mattheworiordan: @funkyboy unfortunately there's not a lot we can do once code necessarily needs to be indented to accommodate some new logic surrounding it.  One solution is to try and keep code logic in encapsulated functions as much as possible, so any additional logic introduced has less impact on the code because on the calling code has new indentation.  You can of course just append [`?w=1`](https://github.com/ably/ably-dotnet/pull/178/commits/816724e21ddc40898eb303ac87232515602df564?w=1) to all commits in Github which will hide whitespace, but annoyingly it then limits what you can do with it i.e. you cannot add comments directly in most cases.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -180,31 +180,38 @@ private void SetAuthMethod()
...
+                {
+                    throw new AblyException(new ErrorInfo("Error calling Auth URL, token request failed. See inner exception for details.", 80019, ex.ErrorInfo.StatusCode), ex);
+                }
             }
             else
             {
                 if (keyId.IsEmpty() || keyValue.IsEmpty())
                 {
-                    throw new AblyException("TokenAuth is on but there is no way to generate one");
+                    throw new AblyException("TokenAuth is on but there is no way to generate one", 80019 );
```
mattheworiordan: Why did you Resharper linter not pick this extra space up?  Anyway, I see you've created a new issue for a linter which we can review in time.

src/IO.Ably.Shared/AblyAuth.cs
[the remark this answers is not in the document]
mattheworiordan: Given we have overloaded constructors for `AblyException` to pass in just reason and code, or reason, code and status code, and seemingly ErrorInfo and Exception, why use `ErrorInfo` at all and not simply allow reason, code, status code, and exception to be passed in.  My thinking is this would make the constructors simpler in some ways as you're never really needing to create an ErrorInfo object specifically for the error which is a little confusing to me.  I do understand that sometimes when you do already have an `ErrorInfo`, simply passing it on is definitely useful.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -26,6 +27,112 @@ static TokenParams CreateTokenParams(Capability capability, TimeSpan? ttl = null
...
+        {
+            var client = await GetRealtimeClient(protocol);
+            client.Connect();
+            // wait until connected
+            await client.WaitForState();
+            // set a bogus AuthUrl
+            client.Options.AuthUrl = new Uri("http://localhost:8910");
+            try
+            {
+                client.Auth.RequestToken();
```
mattheworiordan: Much like my other feedback, this explicit request using a realtime client library is not what `RSA4*` is about.  It's about the Realtime lib doing asynchronous authentication and how it handles that depending on the failure and the current state of the library.  Your approach of explicitly requesting a token is wrong unfortunately, because requesting a token has nothing to do with the client library authenticating itself i.e. a realtime client could be connected, and a user can later call `RequestToken` to get a new token but that does not change the existing connection.  `Authorize` on the other hand is at least more similar to what you are attempting to do here.  So if this test is passing it's quite plausible you have a bug in the .NET lib where `RequestToken` is coupled to the current connection's authentication.  Also, I would prefer to see the library automatically trigger the need to authenticate because a token is expiring (which is what this test is mostly about), as opposed to being explicitly called. Because if we only test with explicit calls, in the case where people rely on it to automatically renew tokens when needed, we'll not have suitable test coverage.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -60,9 +102,12 @@ static TokenParams CreateTokenParams(Capability capability, TimeSpan? ttl = null
             }
             catch (AblyException e)
             {
-                e.ErrorInfo.StatusCode.Should().Be(System.Net.HttpStatusCode.Unauthorized);
-                e.ErrorInfo.Code.Should().BeInRange(40140, 40150);
+                e.ErrorInfo.Code.Should().Be(80019);
             }
+            // check the retry code was run
+            testLogger.MessageSeen.Should().BeTrue();
```
mattheworiordan: It would be a lot nicer to intercept auth requests as opposed to the Logger. Whilst the logger is useful, it's no guarantee that the request was ever made, just that we logged we made the request.

withakay:
@mattheworiordan 
Sorry this PR has become so unwieldy :( I will take on board you comments regarding commits.

Regarding RSC7b, there is a test for that [here](https://github.com/ably/ably-dotnet/blob/a59a2edeab8771a9f0505b8c892f610b6464d66c/src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs#L52).

I will respond to the other comments in line.

src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
```diff
@@ -42,6 +44,44 @@ public class WithTokenAuthAndInvalidToken : RestSandBoxSpecses
...
+                await Task.Delay(TimeSpan.FromSeconds(2));
+
+                //Add this to fool the client it is a valid token
+                almostExpiredToken.Expires = DateTimeOffset.UtcNow.AddHours(1);
+
+                //Trying again with the new token
+                var client = await GetRestClient(protocol, options =>
+                {
+                    options.TokenDetails = almostExpiredToken;
+                    options.ClientId = "123";
```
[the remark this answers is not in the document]
withakay: For `ClientId` I think you are right, it is superfluous (per RSA7b2 I think), I will remove it.
For the `Key` property, my understanding is that needs reseting because the spec states '...no means to renew the token is provided (either an API key, authCallback or authUrl)...' and the `GetRestClient()` helper method sets up a `Key` for the sandbox, thus providing a means to renew. Does that seem correct to you?

src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
```diff
@@ -42,6 +44,44 @@ public class WithTokenAuthAndInvalidToken : RestSandBoxSpecses
...
+                    options.Key = "";
+                });
+
+                await client.StatsAsync();
```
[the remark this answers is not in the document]
withakay: This was removed in a later commit.

src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
```diff
@@ -42,6 +44,44 @@ public class WithTokenAuthAndInvalidToken : RestSandBoxSpecses
...
+                client.AblyAuth.CurrentToken.IsValidToken().Should().BeTrue();
+
+                try
+                {
+                    client.Channels.Get("random").Publish("event", "data");
+                }
+                catch (AblyException e)
+                {
+                    e.ErrorInfo.StatusCode.Should().Be(System.Net.HttpStatusCode.Unauthorized);
```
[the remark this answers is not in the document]
withakay: I have noted the errors in the test logic and will take another look. 

> Out of interest, how are you keeping track of the fact that a test like this only partially completes the requirement as it covers both Realtime & REST yet this test is REST only. 

I added note to the Spec spreadsheet and in a later commit I added a comment to the code `/* only tests rest so does not cover 'in the case of the realtime library, transition the connection to the FAILED state' */

> do you think we should add a pending test so that we don't lose track of this?

Yes, that is probably a better solution

src/IO.Ably.Shared/AblyAuth.cs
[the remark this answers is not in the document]
withakay: > simply allow reason, code, status code, and exception to be passed in

This feels like a somewhat separate problem, should I make a new issue for it?

withakay:
@mattheworiordan I have forked from this branch to work on to work on RSA4a+b in the hope this would be more manageable, see PR #201, I could close this PR until I have picked off the individual items if that is less confusing?

PR #183, opened 2018-02-20
RSA10l

withakay:
Implementation of RSA10l

`"Has an alias method RestClient#authorise and RealtimeClient#authorise that will log a deprecation warning stating that this alias method will be removed..."`

### DIFF
--- a/src/IO.Ably.Tests.Shared/AuthTests/AuthoriseTests.cs
+++ b/src/IO.Ably.Tests.Shared/AuthTests/AuthoriseTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reflection;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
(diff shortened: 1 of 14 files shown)


src/IO.Ably.Tests.Shared/AuthTests/AuthoriseTests.cs
```diff
@@ -15,15 +16,55 @@ public void TokenShouldNotBeSetBeforeAuthoriseIsCalled()
...
         }
 
+        /*
+         * (RSA10a) Instructs the library to create a token immediately and ensures Token Auth is used for all future requests.
+         * See RTC8 for re-authentication behaviour when called for a realtime client
+         */
+
         [Fact]
         [Trait("spec", "RSA10a")]
         public async Task Authorise_WithNotExpiredCurrentTokenAndForceFalse_ReturnsCurrentToken()
```
funkyboy: Would probably be a good idea to rename also test names. For consistency and to not mislead people that read tests as a companion to documentation.

withakay: @funkyboy Sorry, I don't follow. What do you suggest the test name should be?

funkyboy: Authorize_...

withakay: doh!

withakay:
@funkyboy I have updated the test names to be consistent.

funkyboy:
@withakay LGTM 👍

PR #187, opened 2018-02-23
RTE3

withakay:
Tests for RTE3

`(RTE3) EventEmitter#on registers the provided listener for either all events when no event argument is provided, or for only a single named event when an event argument is provided. If on is called more than once with the same listener and event, the listener is added multiple times to its listener registry. Therefore, as an example, assuming the same listener is registered twice using on, and an event is emitted once, the listener would be invoked twice.`


src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -104,8 +108,67 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
...
+                t.SetCompleted();
+            });
+            em.DoDummyEmit(1, "on");
+            var success = await t.Task;
+            success.Should().BeTrue();
+            m.Should().Be("on");
+            // currently one listener
+            counter.Should().Be(1);
+            t = new TaskCompletionAwaiter(100);
+            var tt = new TaskCompletionAwaiter(100);
```
mattheworiordan: I realise this is a .NET thing, but why do we need these waits?

withakay:
@mattheworiordan I have updated the tests based on your feedback.

PR #188, opened 2018-02-23
RTE4

withakay:
added test for RTE4

`(RTE4) EventEmitter#once registers the provided listener for either the first event that is emitted when no event argument is provided, or for only the first occurrence of a single named event when an event argument is provided. If once is called more than once with the same listener, the listener is added multiple times to its listener registry. Therefore, as an example, assuming the same listener is registered twice using once, and an event is emitted once, the listener would be invoked twice. However, all subsequent events emitted would not invoke the listener as once ensures that each registration is only invoked once`

### DIFF
--- a/src/IO.Ably.Shared/EventEmitter.cs
+++ b/src/IO.Ably.Shared/EventEmitter.cs
@@ -119,18 +119,16 @@ namespace IO.Ably
             _lock.EnterUpgradeableReadLock();
             try
             {
-                if (_emitters.Any(x => x.Action == listener) == false)
+                _lock.EnterWriteLock();
+                try
                 {
-                    _lock.EnterWriteLock();
-                    try
-                    {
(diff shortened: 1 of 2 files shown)


withakay:
> I would like to see a once listener registered twice for a single emit personally as that is what the spec describes. 

Sorry for missing that 😔! 
Adding in the test revealed that the Once implimenetation did NOT allow duplicate handlers to be registered, so I removed that check.

>  I would also like to see a once listener not fired when an event is emitted with a non-matching name

I *think* that is already covered (or I am misunderstanding you). The section I am thinking of looks like this:

```
// only catch 1 events
em.Once(1, (Action<TestEventEmitterArgs>)Handler2);
em.DoDummyEmit(2, "on");

// no events should be handled, t & tt should be false here
t.Should().BeFalse();
tt.Should().BeFalse();
```

PR #189, opened 2018-02-23
RTE5

withakay:
test for RTE5

`(RTE5) EventEmitter#off deregisters a listener. If called with a specific event and a listener, it removes all registrations that match both the given listener and the given event; if called only with a listener, it removes all registrations matching the given listener, regardless of whether they are associated with an event or not; if called with no arguments, it removes all registrations, for all events and listeners`

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
@@ -289,12 +289,26 @@ namespace IO.Ably.Tests.Realtime
             em.DoDummyEmit(1, "off");
             message.Should().Be("off");
             counter.Should().Be(1);
-
             Reset();
+
             em.Off();
             em.DoDummyEmit(1, "off");
             em.DoDummyEmit(2, "off");
             em.DoDummyEmit(3, "off");
+            counter.Should().Be(0);
+            Reset();
+
+            em.On(1, (Action<TestEventEmitterArgs>)Listener1);
+            em.On(2, (Action<TestEventEmitterArgs>)Listener2);
+            em.On(1, (Action<TestEventEmitterArgs>)Listener2);
+            em.DoDummyEmit(1, "off");
+            message.Should().Be("off");
+            counter.Should().Be(2);
+            Reset();
+
+            em.Off();
+            em.DoDummyEmit(1, "off");
+            em.DoDummyEmit(2, "off");
             counter.Should().Be(0);
             Reset();
 
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
@@ -265,9 +265,16 @@ namespace IO.Ably.Tests.Realtime
             var t = new TaskCompletionAwaiter(100);
             string m = string.Empty;
             int counter = 0;
-
             bool handled = false;
-            void Handler(TestEventEmitterArgs args)
+
+            void Handler1(TestEventEmitterArgs args)
+            {
+                counter++;
+                m = args.Message;
+                handled = true;
+            }
+
+            void Handler2(TestEventEmitterArgs args)
             {
                 counter++;
                 m = args.Message;
@@ -275,9 +282,9 @@ namespace IO.Ably.Tests.Realtime
             }
 
             // if called with no arguments, it removes all registrations, for all events and listeners
-            em.On(1, (Action<TestEventEmitterArgs>)Handler);
-            em.On(2, (Action<TestEventEmitterArgs>)Handler);
-            em.On(3, (Action<TestEventEmitterArgs>)Handler);
+            em.On(1, (Action<TestEventEmitterArgs>)Handler1);
+            em.On(2, (Action<TestEventEmitterArgs>)Handler1);
+            em.On(3, (Action<TestEventEmitterArgs>)Handler1);
             em.DoDummyEmit(1, "off");
             handled.Should().BeTrue();
             m.Should().Be("off");
@@ -292,36 +299,42 @@ namespace IO.Ably.Tests.Realtime
 
             // if called only with a listener, it removes all registrations matching the given listener,
             // regardless of whether they are associated with an event or not;
-            em.On(1, (Action<TestEventEmitterArgs>)Handler);
-            handled = false;
-            em.DoDummyEmit(1, "off");
-            handled.Should().BeTrue();
-            counter.Should().Be(2);
-            handled = false;
-            em.Off((Action<TestEventEmitterArgs>)Handler);
-            em.DoDummyEmit(1, "off");
-            handled.Should().BeFalse();
-            counter.Should().Be(2);
-
-            // If called with a specific event and a listener,
-            // it removes all registrations that match both the given listener and the given event
-            em.On(1, (Action<TestEventEmitterArgs>)Handler);
+            em.On(1, (Action<TestEventEmitterArgs>)Handler1);
+            em.On(1, (Action<TestEventEmitterArgs>)Handler2);
             handled = false;
             em.DoDummyEmit(1, "off");
             handled.Should().BeTrue();
             counter.Should().Be(3);
-
-            // no handler for this event, so this should have no effect
             handled = false;
-            em.Off(2, (Action<TestEventEmitterArgs>)Handler);
+            em.Off((Action<TestEventEmitterArgs>)Handler2);
             em.DoDummyEmit(1, "off");
             handled.Should().BeTrue();
             counter.Should().Be(4);
-            em.Off(1, (Action<TestEventEmitterArgs>)Handler);
             handled = false;
+            em.Off((Action<TestEventEmitterArgs>)Handler1);
             em.DoDummyEmit(1, "off");
             handled.Should().BeFalse();
             counter.Should().Be(4);
+
+            // If called with a specific event and a listener,
+            // it removes all registrations that match both the given listener and the given event
+            em.On(1, (Action<TestEventEmitterArgs>)Handler1);
+            handled = false;
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeTrue();
+            counter.Should().Be(5);
+
+            // no handler for this event, so this should have no effect
+            handled = false;
+            em.Off(2, (Action<TestEventEmitterArgs>)Handler1);
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeTrue();
+            counter.Should().Be(6);
+            em.Off(1, (Action<TestEventEmitterArgs>)Handler1);
+            handled = false;
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeFalse();
+            counter.Should().Be(6);
         }
 
         public EventEmitterSpecs(ITestOutputHelper output) : base(output)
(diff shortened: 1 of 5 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -209,6 +209,78 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
...
+            }
+
+            // if called with no arguments, it removes all registrations, for all events and listeners
+            em.On(1, (Action<TestEventEmitterArgs>)Handler);
+            em.On(2, (Action<TestEventEmitterArgs>)Handler);
+            em.On(3, (Action<TestEventEmitterArgs>)Handler);
+            em.DoDummyEmit(1, "off");
+            var success = await t.Task;
+            success.Should().BeTrue();
+            m.Should().Be("off");
```
mattheworiordan: Should you not check the counter is equal to 1 here i.e. given you check it's again below thus provign the off has worked? I realise you could say it's worked because it's 1 and not 3, but this would be clearer IMHO

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -209,6 +209,78 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
...
+                m = args.Message;
+                t.SetCompleted();
+            }
+
+            // if called with no arguments, it removes all registrations, for all events and listeners
+            em.On(1, (Action<TestEventEmitterArgs>)Handler);
+            em.On(2, (Action<TestEventEmitterArgs>)Handler);
+            em.On(3, (Action<TestEventEmitterArgs>)Handler);
+            em.DoDummyEmit(1, "off");
+            var success = await t.Task;
```
mattheworiordan: Why is this task needed?  Isn't everything emitted synchronously?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -209,6 +209,78 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
...
+            // if called only with a listener, it removes all registrations matching the given listener,
+            // regardless of whether they are associated with an event or not;
+            em.On(1, (Action<TestEventEmitterArgs>)Handler);
+            t = new TaskCompletionAwaiter(100);
+            em.DoDummyEmit(1, "off");
+            success = await t.Task;
+            success.Should().BeTrue();
+            counter.Should().Be(2);
+            t = new TaskCompletionAwaiter(100);
+            em.Off((Action<TestEventEmitterArgs>)Handler);
```
mattheworiordan: I don't think this is a good test because you're only test one of the two conditions i.e. calling off with a single event handler for an event handler that is not registered should have no effect, and calling with a single event handler when it matches only one of two should confirm that one still fires i.e.  this test would still pass if you swapped `em.Off((Action<TestEventEmitterArgs>)Handler);` with just `em.Off();` and thus they could share the same implementation.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -209,6 +209,78 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
...
+            counter.Should().Be(3);
+            t = new TaskCompletionAwaiter(100);
+            // no handler for this event, so this should have no effect
+            em.Off(2, (Action<TestEventEmitterArgs>)Handler);
+            em.DoDummyEmit(1, "off");
+            success = await t.Task;
+            success.Should().BeTrue();
+            counter.Should().Be(4);
+            t = new TaskCompletionAwaiter(100);
+            em.Off(1, (Action<TestEventEmitterArgs>)Handler);
```
mattheworiordan: As above, we're not testing the other conditions where the off does not match

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: You're right, it isn't needed, I could just use a bool. Back to the drawing board!

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: I think I actually removed that because I thought it was superfluous, I will add it back.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: Updated to the following

```em.On(1, (Action<TestEventEmitterArgs>)Handler1);
            em.On(1, (Action<TestEventEmitterArgs>)Handler2);
            handled = false;
            em.DoDummyEmit(1, "off");
            handled.Should().BeTrue();
            counter.Should().Be(3);
            handled = false;
            em.Off((Action<TestEventEmitterArgs>)Handler2);
            em.DoDummyEmit(1, "off");
            handled.Should().BeTrue();
            counter.Should().Be(4);
            handled = false;
            em.Off((Action<TestEventEmitterArgs>)Handler1);
            em.DoDummyEmit(1, "off");
            handled.Should().BeFalse();
            counter.Should().Be(4);```

Comment (mattheworiordan) [reply src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs]: 👍

File: src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs (hunk 4 above)
Comment (withakay) [reply src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs]: I have updated this test to cover the missing scenarios

```
// If called with a specific event and a listener,
// it removes all registrations that match both the given listener and the given event
em.On(1, (Action<TestEventEmitterArgs>)Listener1);
em.On(1, (Action<TestEventEmitterArgs>)Listener2);
em.On(2, (Action<TestEventEmitterArgs>)Listener1);
handled = false;
em.DoDummyEmit(1, "off");
handled.Should().BeTrue();
counter.Should().Be(6);

handled = false;
em.DoDummyEmit(2, "off");
handled.Should().BeTrue();
counter.Should().Be(7);

// no handler for this event, so this should have no effect
em.Off(3, (Action<TestEventEmitterArgs>)Listener1);
handled = false;
em.DoDummyEmit(1, "off");
handled.Should().BeTrue();
counter.Should().Be(9);

// no handler for this listener, so this should have no effect
em.Off(2, (Action<TestEventEmitterArgs>)Listener2);
handled = false;
em.DoDummyEmit(1, "off");
handled.Should().BeTrue();
counter.Should().Be(11);

// remove handler 1 of 2 remaining
em.Off(1, (Action<TestEventEmitterArgs>)Listener1);
handled = false;
em.DoDummyEmit(1, "off");
handled.Should().BeTrue();
counter.Should().Be(12);

// remove the final handler
em.Off(1, (Action<TestEventEmitterArgs>)Listener2);
handled = false;
em.DoDummyEmit(1, "off");
handled.Should().BeFalse();
counter.Should().Be(12);
```

**Comment by withakay** [comment]
@mattheworiordan @funkyboy Can you take another look, thanks.

File: src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs [hunk 5]
```diff
@@ -257,6 +257,108 @@ void Handler2(TestEventEmitterArgs args)
...
+            em.DoDummyEmit(2, "off");
+            em.DoDummyEmit(3, "off");
+            handled.Should().BeFalse();
+            counter.Should().Be(1);
+
+            // if called only with a listener, it removes all registrations matching the given listener,
+            // regardless of whether they are associated with an event or not;
+            em.On(1, (Action<TestEventEmitterArgs>)Listener1);
+            em.On(1, (Action<TestEventEmitterArgs>)Listener2);
+            handled = false;
```
funkyboy: Maybe this should be moved 2 lines up? Just so it's clear it's a precondition for the test.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -257,6 +257,108 @@ void Handler2(TestEventEmitterArgs args)
...
+            counter.Should().Be(12);
+
+            // remove the final handler
+            em.Off(1, (Action<TestEventEmitterArgs>)Listener2);
+            handled = false;
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeFalse();
+            counter.Should().Be(12);
+        }
+
```
funkyboy: What's the rationale behind the `counter` variable?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -257,6 +257,108 @@ void Handler2(TestEventEmitterArgs args)
...
+            counter.Should().Be(3);
+            em.Off((Action<TestEventEmitterArgs>)Listener2);
+            handled = false;
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeTrue();
+            counter.Should().Be(4);
+            em.Off((Action<TestEventEmitterArgs>)Listener1);
+            handled = false;
+            em.DoDummyEmit(1, "off");
+            handled.Should().BeFalse();
```
funkyboy: Is this statement needed?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: No problem. I have consistently placed `handled = false` after `em.On/Off` calls, would you prefer them all to be before?

funkyboy: I do. It's a personal preference, but I think the code flows better. Not a strong requirement tho.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: Strictly speaking, I suppose not, as the counter not ticking up proves the same point, but I do think it adds semantic meaning.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: There are 2 listener methods, the counter is used to demonstrate that they have been called the correct number of times, there are times when the same listener should be called twice because it has been added twice for example, a counter seemed like a simple way of tracking such things.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: @funkyboy I have pushed a change that places the `handled = false` statements before the `On/Off` calls

withakay:
EBI comments left by @paddybyers addressed

PR #193, opened 2018-03-13
RTE6

withakay:
A test for for RTE6

`(RTE6) EventEmitter#emit emits an event, calling registered listeners with the given event name and any other given arguments. If an exception is raised in any of the listeners, the exception is caught by the EventEmitter and the exception is logged to the Ably logger. Tests must exist to ensure exceptions raised in client code do not propagate and inhibit other event processing within the client library.
`

Additionally I have extracted the TestLogger class into its own file as it is not SandboxSpec specific.


src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -65,7 +65,7 @@ public void WhenClosingAConnection_ItShouldRaiseClosingAndClosedEvents()
         {
             var client = GetClientWithFakeTransport();
 
-            //Start collecting events after the connection is open
+            // Start collecting events after the connection is open
```
funkyboy: I am guessing this branch was not yet "linterized"?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -359,7 +359,50 @@ void Listener2(TestEventEmitterArgs args)
             counter.Should().Be(12);
         }
 
-        public EventEmitterSpecs(ITestOutputHelper output) : base(output)
+        [Fact]
+        [Trait("spec", "RTE6")]
+        public void WithEventEmitter_WhenExceptionRaised_ExceptionIsCaughtAndLogged()
+        {
+            string exceptionMessage = "Listener1 exception";
```
funkyboy: I don't see any reason to keep this var around, unless you are going to use it in assertions afterwards. But I don't see it reused so far.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -359,7 +359,50 @@ void Listener2(TestEventEmitterArgs args)
...
+
+            em.Once(1, (Action<TestEventEmitterArgs>)Listener1);
+            em.Once(1, (Action<TestEventEmitterArgs>)Listener2);
+            em.DoDummyEmit(1, string.Empty);
+            handled1.Should().BeTrue();
+            handled2.Should().BeTrue();
+            logger.MessageSeen.Should().BeTrue();
+        }
+
+        public EventEmitterSpecs(ITestOutputHelper output)
```
funkyboy: Maybe this belongs to another commit?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: It is passed to the TestLogger constructor 

`var logger = new TestLogger(messageToTest: exceptionMessage);` 

and then it is passed to the Exception that is raised 

`throw new Exception(exceptionMessage);`

So it avoids duplication there, but my hope was that it would also make it obvious that the TestLogger uses that string to do its work.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: Correct

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
withakay: I don't understand why. Do you mean it belongs in different PR perhaps? I could revert this file if you think that is better, it won't make a difference to the test.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
funkyboy: Sorry, I missed that it was reused in the `throw` statement.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
funkyboy: Well, if you remove this and it doesn't make any difference, that means it definitely belongs to another PR :)

withakay:
@funkyboy I have reverted the "fix some style violations" commit.

funkyboy:
@withakay that also reverts the change on line 68, which we want.

withakay:
@funkyboy 

> that also reverts the change on line 68, which we want.

We discussed this and agreed that I should 'revert the revert', which I did with `get reset` and `git push -f`, so could you give this PR your seal of approval?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
funkyboy: I still think this doesn't belong to this PR.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/EventEmitterSpecs.cs
```diff
@@ -359,7 +359,50 @@ void Listener2(TestEventEmitterArgs args)
...
+            void Listener2(TestEventEmitterArgs args)
+            {
+                handled2 = true;
+            }
+
+            em.On(1, (Action<TestEventEmitterArgs>)Listener1);
+            em.On(1, (Action<TestEventEmitterArgs>)Listener2);
+            em.DoDummyEmit(1, string.Empty);
+            handled1.Should().BeTrue();
+            handled2.Should().BeTrue();
```
paddybyers: An issue with this test is that it would pass if the listeners were called in the wrong order (ie the one throwing the exception happened after `handled1` and `handled2` are both updated. Can you either explicitly check the order they are called, or put a third listener after the one that throws?

withakay:
@paddybyers 
> it should test conditions such as what happens when emitting an event if listeners are added or removed within the listener callback itself. (that includes listeners removing themselves, removing other listeners, removing all listeners, adding a new listener, etc).

AFAICT the spec doesn't say what should happen in the above cases.
Intuitively I feel that if a listener adds another listener that second listener should be called with the current Emit cycle and similarly if a listener removes another listener that was due to be called, it should not. [Currently that is *not* what happens](https://github.com/ably/ably-dotnet/blob/d0131f05adde846503f9666198ef3116127df77d/src/IO.Ably.Shared/EventEmitter.cs#L275). As far as I can tell [the Java implementation] does work how I expect (https://github.com/ably/ably-java/blob/2af72fb88e0d5e0b47820c1c1cab6a2235f5df73/lib/src/main/java/io/ably/lib/util/EventEmitter.java#L84). Could you please clarify the expected behaviour for me, or point me at the right docs?

paddybyers:
> Could you please clarify the expected behaviour for me, or point me at the right place in the docs?

See https://github.com/ably/docs/blob/master-1-1/content/client-lib-development-guide/features.textile, RTE6a. This change obviously wasn't backported to 1.0.

withakay:
@paddybyers I have added a new test to cover RTE6a and the comments you left previously.

PR #194, opened 2018-03-14
RSA8e test improvement 

withakay:
A small addition to the existing tests for RSA8e to validate the property names and access level of the AuthOptions

### DIFF
--- a/src/IO.Ably.Shared/AuthOptions.cs
+++ b/src/IO.Ably.Shared/AuthOptions.cs
@@ -14,7 +14,7 @@ namespace IO.Ably
         /// The callback used to get a new <see cref="IO.Ably.TokenDetails"/> or <see cref="IO.Ably.TokenRequest"/>.
         /// AuthCallback is used by internally by <see cref="IO.Ably.AblyAuth"/>.RequestTokenAsync.
         /// </summary>
-        public Func<TokenParams, Task<object>> AuthCallback;
+        public Func<TokenParams, Task<object>> AuthCallback { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/IO.Ably.Shared/AuthOptions.cs
```diff
@@ -14,7 +14,7 @@ public class AuthOptions
         /// The callback used to get a new <see cref="IO.Ably.TokenDetails"/> or <see cref="IO.Ably.TokenRequest"/>.
         /// AuthCallback is used by internally by <see cref="IO.Ably.AblyAuth"/>.RequestTokenAsync.
         /// </summary>
-        public Func<TokenParams, Task<object>> AuthCallback;
+        public Func<TokenParams, Task<object>> AuthCallback { get; set; }
```
withakay: this change was made to prevent a violation of SA1401 - A field within a C# class has an access modifier other than private.

src/IO.Ably.Tests.Shared/AuthTests/AuthOptionsTests.cs
```diff
@@ -1,32 +1,70 @@
...
+            "UseTokenAuth"
+        };
+
+        private AuthOptions GetCompleteOptions()
         {
             return new AuthOptions()
             {
+                AuthCallback = param => null,
                 AuthHeaders = new Dictionary<string, string> { { "Test", "Test" } },
+                AuthMethod = HttpMethod.Get,
```
funkyboy: Might wanna test what happens when you omit this line, as `GET` is the default method.

PR #195, opened 2018-03-15
RSC16

withakay:
A very minor change, I cleaned up the test and added an assertion to show there is no time zone offset

### DIFF
--- a/src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
@@ -31,10 +31,14 @@ namespace IO.Ably.Tests
         public async Task Time_ShouldReturnAValidDateTimeOffset(Protocol protocol)
         {
             var client = await GetRestClient(protocol);
+            var serverTime = await client.TimeAsync();
 
-            var now = await client.TimeAsync();
+            // server time should be similar to the system time
+            // here we allow the system clock to be 15 minutes fast or slow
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Tests.Shared/Rest/RestSandBoxSpecs.cs
```diff
@@ -31,10 +31,14 @@ public RestSandBoxSpecses(AblySandboxFixture fixture, ITestOutputHelper output)
         public async Task Time_ShouldReturnAValidDateTimeOffset(Protocol protocol)
         {
             var client = await GetRestClient(protocol);
+            var serverTime = await client.TimeAsync();
 
-            var now = await client.TimeAsync();
+            // server time should be similar to the system time
+            // here we allow the system clock to be 15 minutes fast or slow
+            serverTime.Should().BeCloseTo(DateTimeOffset.UtcNow, (int)TimeSpan.FromMinutes(30).TotalMilliseconds);
```
funkyboy: Is it 15 or 30?

withakay: I thought this was correct: 15 mins before + 15 mins after = 30 min.
But I went back and read [the source code](https://github.com/fluentassertions/fluentassertions/blob/c26c9641e0c33e51c790a91a3fe37e15a5898182/Src/FluentAssertions/Primitives/DateTimeOffsetAssertions.cs#L123) and it seems my understanding was wrong, I will update this.

PR #196, opened 2018-03-15
changes and tests to support RSA10k

withakay:
`(RSA10k) If the AuthOption argument’s queryTime attribute is true, it will obtain the server time once and persist the offset from the local clock. All future token requests generated directly or indirectly via a call to authorize will not obtain the server time, but instead use the local clock offset to calculate the server time.`

Internally the library uses `DateTimeOffset.UtcNow` I have not done any work related to the following part of the RSA10k spec as it didn't seem relevant.

`The client library itself MAY internally discard the cached local clock offset in situations in which it may have been invalidated, such as if there is a local change to the date, time, or timezone, of the client device. For clarity however, there is no requirement for this cache invalidation to be available to consumers of the client library API.`

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -19,8 +19,14 @@ namespace IO.Ably
             Options = options;
             _rest = rest;
             Logger = options.Logger;
+            ServerTime = () => _rest.TimeAsync();
+            ServerTimeOffset = () => null;
             Initialise();
         }
+
+        protected Func<Task<DateTimeOffset>> ServerTime { get; set; }
+
+        protected Func<DateTimeOffset?> ServerTimeOffset { get; private set; }
 
         internal Func<DateTimeOffset> Now { get; set; }
 
@@ -86,6 +92,12 @@ namespace IO.Ably
             }
 
             LogCurrentAuthenticationMethod();
+        }
+
+        private async void SetServerTimeOffset()
+        {
+            TimeSpan diff = Now() - await ServerTime();
+            ServerTimeOffset = () => Now() - diff;
         }
 
         private void SetAuthMethod()
@@ -173,9 +185,11 @@ namespace IO.Ably
 
             var @params = MergeTokenParamsWithDefaults(tokenParams);
 
-            if (mergedOptions.QueryTime.GetValueOrDefault(false))
+            if (mergedOptions.QueryTime.GetValueOrDefault(false)
+                && !ServerTimeOffset().HasValue)
             {
-                @params.Timestamp = await _rest.TimeAsync();
+                SetServerTimeOffset();
+                @params.Timestamp = ServerTimeOffset();
             }
 
             EnsureSecureConnection();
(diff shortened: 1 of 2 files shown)


src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -430,9 +444,11 @@ private void SetCurrentAuthOptions(AuthOptions options)
 
             var @params = MergeTokenParamsWithDefaults(tokenParams);
 
-            if (mergedOptions.QueryTime.GetValueOrDefault(false))
+            if (mergedOptions.QueryTime.GetValueOrDefault(false)
```
paddybyers: I don't understand this logic. Surely we need to do something like:

```
            if (mergedOptions.QueryTime.GetValueOrDefault(false)) {
                if(!ServerTimeOffset().HasValue) {
                    SetServerTimeOffset();
                }
                @params.Timestamp = ServerTimeOffset();
            }
```

As things stand you're not setting the `Timestamp` at all after the first time the offset is initialised afaict.

The logic is duplicated below (dry that?)

paddybyers: I'm talking about the case that queryTime is true. That requires you to set a server-provided time, or a time determined using a known offset to server time. So if `queryTime` is true, you must assign `Timestamp` with something, and can't leave it at the default value (which is locally-determined).

Your current logic, afaict, doesn't do that. You have:

```
if (mergedOptions.QueryTime.GetValueOrDefault(false) && !ServerTimeOffset().HasValue) {
    SetServerTimeOffset();
    @params.Timestamp = ServerTimeOffset();
}
```

That is, if `ServerTimeOffset().HasValue` then the if condition isn't taken, and `Timestamp` is not set.

withakay: Ah yes, I see what you mean. I think a fix would look like this (assuming that if a Timestamp exists already, we don't want to overwrite it?)

```
if (mergedOptions.QueryTime.GetValueOrDefault(false)
    && !ServerTimeOffset().HasValue)
{
    SetServerTimeOffset();
}

// @params.Timestamp is a nullable type (DateTimeOffset?)
if (!@params.Timestamp.HasValue)
{
    // ServerTimeOffset() will return a null if SetServerTimeOffset() has never been called
    @params = ServerTimeOffset(); 
}
```

paddybyers: The spec (RSA9d) says:

> Generates a timestamp from current time if not provided

So I think the logic needs to be:

```
if(!params.Timestamp) {
  if(options.QueryTime) {
    if(!ServerTimeOffset().HasValue) {
      SetServerTimeOffset();
    }
    params.Timestamp = ServerTimeOffset();
  } else {
    params.timestamp = (local)
  }
}
```

withakay: I don't think we need to set the `(local)` here as the RSA9d requirements are met later on when the `TokenRequest` object is populated [(code here)](https://github.com/ably/ably-dotnet/blob/2c14b2d8756b54ecd178238b654c9abbb73545c5/src/IO.Ably.Shared/TokenRequestPostData.cs#L80) and there are [several pre-existing tests for RSA9d](
https://github.com/ably/ably-dotnet/blob/29c014a7d89e4627f1c288ec1f11b0001cbbeb7b/src/IO.Ably.Tests.Shared/AuthTests/AuthorisationTests.cs#L187) that pass (although I have not reviewed them).

I have just pushed some changes that I think cover this and have extended the test to show that the `tokenRequest.Timestamp` is set correctly when `QueryTime` is or isn't set.

PR #198, opened 2018-03-15
TR3

withakay:
(TR3) ProtocolMessage Flag enum has the following values, where a flag with value n is defined to be set if the bitwise AND of the flags field with 2ⁿ is nonzero
- (TR3a) 0: HAS_PRESENCE
- (TR3b) 1: HAS_BACKLOG
- (TR3c) 2: RESUMED
- (TR3d) 3: HAS_LOCAL_PRESENCE
- (TR3e) 4: TRANSIENT
- (TR3q) 16: PRESENCE
- (TR3r) 17: PUBLISH
- (TR3s) 18: SUBSCRIBE
- (TR3t) 19: PRESENCE_SUBSCRIBE

I have added the above (replacing the `const` with `enums`), renamed `MessageFlags` to `Flags` to better align with the spec.

I have removed the HasPresenceFlag property in favour of a `HasFlag(Flag flag)` method because the current naming was going to become unclear, i.e. does `HasPresence` mean 'has a PRESENCE flag' or 'has a HAS_PRESENCE flag'

I had added a simple test that shows all the emums are added and demonstrates HasFlag.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ProtocolMessageSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ProtocolMessageSpecs.cs
@@ -9,19 +9,32 @@ namespace IO.Ably.Tests.Shared.Realtime
 {
     public class ProtocolMessageTests
     {
+
+
         [Fact]
         [Trait("spec", "TR3")]
+        public void ProtocolMessageFlagHaveCorrectValues()
+        {
+            // TR3a to TR3e
+            ((int)ProtocolMessage.Flag.HasPresence).Should().Be(1 << 0);
+            ((int)ProtocolMessage.Flag.HasBacklog).Should().Be(1 << 1);
+            ((int)ProtocolMessage.Flag.Resumed).Should().Be(1 << 2);
+            ((int)ProtocolMessage.Flag.HasLocalPresence).Should().Be(1 << 3);
+            ((int)ProtocolMessage.Flag.Transient).Should().Be(1 << 4);
+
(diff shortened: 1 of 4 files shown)


src/IO.Ably.Tests.Shared/Realtime/ProtocolMessageSpecs.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace IO.Ably.Tests.Shared.Realtime
+{
+    public class ProtocolMessageTests
+    {
+        [Fact]
+        [Trait("spec", "TR3")]
+        [Trait("spec", "TR4i")]
+        public void FlagsContainsBitFlags()
+        {
+            // Arrange
```
paddybyers: Can we not pick these values up from the library instead of duplicating the declarations here?

withakay: I could get the values from the `Flag` enum (e.g. `(int)ProtocolMessage.Flag.HasPresence`, but I can't see the benefit as then I would need validate the values are correct. Actually, there should be something here to validate all the values so as to fully cover TR3a-r? I will add that

paddybyers:
> https://github.com/ably/ably-dotnet/pull/198/commits/d01e7a47cc90bb1d170237b2a8c7bdfe2929a153

LGTM

withakay:
@paddybyers I have improved the test, adding a test to check all the the specific values for the `Flags` and then pulling those values from the lib to prevent duplication.

PR #201, opened 2018-03-21
RSA4a+b

withakay:
I have reworked tests for RSA4a and RSA4b and replaced the inadequate tests for RSA4c with stubs for the time being. I have put these on a new branch to (hopefully) make it easier to review (let me know if you think there is a better way).

I would love to get some feed back on this, make sure I am aligning with the spec


src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -33,36 +36,114 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+
+            bool connected = false;
+            realtimeClient.Connection.Once(ConnectionState.Connected, (_) => { connected = true; });
+
+            // assert that there is no pre-existing error
+            realtimeClient.Connection.ErrorReason.Should().BeNull();
+
+            await realtimeClient.WaitForState(ConnectionState.Failed);
+            realtimeClient.Connection.State.Should().Be(ConnectionState.Failed);
+            connected.Should().BeFalse();
```
funkyboy: Likely there's no need to test this, but doesn't hurt :)

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -75,145 +156,172 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+                Ttl = TimeSpan.FromMilliseconds(1000)
+            });
 
-            // wait until connected
-            await client.WaitForState();
+            // this realtime client will have a key for the sandbox, thus a means to renew
+            var realtimeClient = await GetRealtimeClient(protocol, (options, _) =>
+            {
+                options.TokenDetails = token;
+                options.LogLevel = LogLevel.Debug;
```
funkyboy: Is this needed?

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -75,145 +156,172 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
 
-            // RSA4c3 should still be connected
-            client.Connection.State.Should().Be(ConnectionState.Connected);
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSA4c")]
+        [Trait("spec", "RSA4c1")]
+        [Trait("spec", "RSA4c3")]
+        public async Task AuthToken_WithConnectedRealtimeClient_WhenAuthCallbackFails_ShouldRaiseError(Protocol protocol)
+        {
```
funkyboy: `throw new Exception("WIP test stub");` ?

withakay: In another PR @mattheworiordan suggested it would be better to have stubs to demonstrate unfinished items, throwing an exception like this makes the tests fail so I am not likely to lose track of them.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
withakay: Nope, well spotted :)

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
funkyboy: Right, I'd expect a `throw new Exception("WIP test stub");` between `{ }`. No?

withakay: fixed.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -33,36 +36,114 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+            // set the Key to an empty string to override the sandbox settings
+            var restClient = await helper.GetRestClientWithRequests(protocol, almostExpiredToken, invalidateKey: true);
 
+            // check the client thinks the token is valid
+            restClient.AblyAuth.CurrentToken.IsValidToken().Should().BeTrue();
             try
             {
-                client.Channels.Get("random").Publish("event", "data");
-                throw new Exception("Unexpected success, the proceeding code should have raised and AblyException");
+                await restClient.Channels.Get("random").PublishAsync("event", "data");
```
paddybyers: I'd prefer that we use channel names everywhere that are guaranteed to be unique, to avoid any possibility of tests interfering. Either generate random names, or use names that are derived in some way from the test name (qualified for protocol, etc, if the test is run multiple times).

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -33,36 +36,114 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+                throw new Exception("Unexpected success, the proceeding code should have raised an AblyException");
```
paddybyers: typo "preceding"

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -33,36 +36,114 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
             }
             catch (AblyException e)
             {
+                // the server responds with a token error
+                // (401 HTTP status code and an Ably error value 40140 <= code < 40150)
```
paddybyers: You should be able to determine the specific expected code for the case you're setting up.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -33,36 +36,114 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSA4a")]
+        public async Task RealtimeClient_ConnectedWithExpiringToken_WhenTokenExpired_ShouldNotRetryAndHaveError(Protocol protocol)
+        {
+            var helper = new RSA4Helper(this);
+
+            // Create a token that is valid long enough for a successful connection to occur
+            var authClient = await GetRestClient(protocol);
+            var almostExpiredToken = await authClient.AblyAuth.RequestTokenAsync(new TokenParams { ClientId = "123", Ttl = TimeSpan.FromMilliseconds(2000) });
```
paddybyers: I would make this longer. Running on a congested CI system can add several seconds delay to things. maybe 8s ?

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -74,8 +74,8 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
             restClient.AblyAuth.CurrentToken.IsValidToken().Should().BeTrue();
             try
             {
-                await restClient.Channels.Get("random").PublishAsync("event", "data");
-                throw new Exception("Unexpected success, the proceeding code should have raised an AblyException");
+                await restClient.Channels.Get("RSA4a".AddRandomSuffix()).PublishAsync("event", "data");
+                throw new Exception("Unexpected success, the preceeding code should have raised an AblyException");
```
paddybyers: http://www.dictionary.com/browse/preceding

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
withakay: OK, in general is testing for a specific code better when the spec just mentions a range?

paddybyers: > the spec just mentions a range

The spec mentions a range in a slightly different context; it says that any code in this range means that an error is treated as non-fatal. However, in your test you're creating a specific set of conditions, and you're checking for a specific outcome. In that case, I think you should check that the returned code is what you expect; that's part of confirming that the test did as you expect.

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
withakay: 🤦‍♂️ fixed

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
withakay: Makes sense. Changes committed.

withakay:
@paddybyers I think I have addressed all the comments you made.

PR #203, opened 2018-03-23
AuthCallback return an AuthCallbackResult

withakay:
As discussed out of band, I have updated `AuthCallback` to return a `Task<AuthCallbackResult>` rather than `Task<object>`. 

I have added inline XmlDocs that should make using AuthCallback easier and more intuitive.

I have updated AblyAuth to support this new type and I have updated all the tests that involve AuthCallbacks.

Additionally I have cleaned up the AblyAuth and AblyException classes


paddybyers:
Is this now a breaking change?

paddybyers:
> Yes, as part of a v1.0 release

OK, yes, I forgot this was 1.0

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -192,30 +192,59 @@ private void SetAuthMethod()
...
+                        return details;
                     }
 
-                    if (callbackResult is TokenRequest || callbackResult is string)
+                    if (callbackResult.TokenCompatibleObject is TokenRequest tokenRequest)
                     {
-                        postData = GetTokenRequest(callbackResult);
-                        request.Url = $"/keys/{postData.KeyName}/requestToken";
+                        postData = tokenRequest;
+                        request.Url = RequestTokenUrl(postData.KeyName);
```
funkyboy: Can be
`request.Url = RequestTokenUrl(tokenRequest.KeyName);`

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -192,30 +192,59 @@ private void SetAuthMethod()
...
-                        throw new AblyException($"AuthCallback returned an unsupported type ({callbackResult.GetType()}. Expected either TokenDetails or TokenRequest", 80019);
+                        throw new AblyException(
+                            $"AuthCallbackResult contains an unsupported type ({callbackResult.TokenCompatibleObject.GetType()}. Expected either TokenDetails or TokenRequest",
+                            80019);
                     }
                 }
+                catch (AblyException aex) when (shouldCatch && aex.ErrorInfo?.StatusCode == HttpStatusCode.Forbidden)
+                {
+                    throw new AblyException(
+                            "Error calling AuthCallback, token request failed. See inner exception for details.",
```
funkyboy: Can we extract some more relevant info from `aex` and display it here?

src/IO.Ably.Shared/AblyException.cs
```diff
@@ -30,61 +28,71 @@ public AblyException()
...
+        /// Initializes a new instance of the <see cref="AblyException"/> class.
         /// Creates AblyException. ErrorInfo is automatically generated based on the inner exception message. StatusCode is set to 50000
         /// </summary>
         /// <param name="ex">Original exception to be wrapped.</param>
         public AblyException(Exception ex)
             : this(new ErrorInfo("Unexpected error :" + ex.Message, 50000), ex)
-            {
-            }
+        {
+        }
```
funkyboy: Hasn't the linter been merged into master? If yes, rebase this on master, so we won't see these anymore.

src/IO.Ably.Shared/AuthCallbackResult.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public AuthCallbackResult()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthCallbackResult"/> class
+        /// </summary>
+        /// <param name="tokenCompatibleObject">a signed <see cref="TokenRequest"/>, <see cref="TokenDetails"/> or a token string.</param>
+        public AuthCallbackResult(object tokenCompatibleObject)
```
funkyboy: Why is this of type generic object?

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -717,15 +721,22 @@ void DefaultOptionsAction(ClientOptions options, TestEnvironmentSettings setting
             public Task<AblyResponse> AblyResponseWith401Status(AblyRequest request)
             {
                 Requests.Add(request);
-                var r = new AblyResponse(string.Empty, "application/json", string.Empty.GetBytes()) { StatusCode = HttpStatusCode.Unauthorized };
+                var r = new AblyResponse(string.Empty, "text/plain", string.Empty.GetBytes()) { StatusCode = HttpStatusCode.Unauthorized };
                 return Task.FromResult(r);
```
funkyboy: Why not `throw AblyException.FromResponse(r);` ?

src/IO.Ably.Shared/AblyAuth.cs
withakay: It can be, but postData need to be set (it is access later on) so I don't see an advantage either way. I will change it if you want?

src/IO.Ably.Shared/AblyAuth.cs
withakay: I don't think so, we have no idea what the `aex` will be and based on what Simon said it is somewhat irrelevant. Additionally `aex` is used to set the InnerException. This `catch` block is a fail safe and also a special case: If the `AuthCallback` throws an exception that should really have been handled in the `AuthCallback` and used to set an ErrorInfo and that Exception is an `AblyException` and that also has a StatusCode of 403 (Forbidden) then it is caught here. This will most likely be changed again when I can do more work on RSA4d

src/IO.Ably.Shared/AblyException.cs
withakay: Not master, but it has been merged to a branch I am working off of. This branch is forked from the branch that but the reformatted code was merged into, for some reason this one was not picked up by StyleCop. I will at some point do another PR to integrate the linter into the CI process

funkyboy: I think the StyleCop should kick in before committing.

withakay: I have a local pre-commit hook that runs the build fails on certain warnings, I am still tweaking which somewhat and some should definitely just be warnings IMO. Also, I don't know an nice way to distribute git hooks with a repo, all the options I come up with seem pretty lacklustre, do you have any ideas about that?

src/IO.Ably.Shared/AuthCallbackResult.cs
withakay: tokenCompatibleObject can be a `string`, `TokenDetails` or `TokenRequest` there is no way to represent three distinct types in a strict manner. Ideally c# would have union types, but that is a few releases off, so it has to be an `object`

src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
withakay: Your probably right, I added that helper in anticipation that I would need it, but it is not actually called. I will remove it

src/IO.Ably.Shared/AblyException.cs
funkyboy: Not specifically with .NET, but in the past I successfully used something like https://coderwall.com/p/vt0lpg/pre-commit-hooks

src/IO.Ably.Shared/AblyAuth.cs
funkyboy: Yep, seems unnecessary.

src/IO.Ably.Shared/AblyException.cs
ashikns: Distributing hooks as part of a git repo - always something I wanted but never found an answer. As for .Net, I've tried
 - Distributing StyleCop cli with the repo and including that as a post build process in the .csproj.
 - Using Roslyn analyzers.

PS: I don't know how I ended up here in this PR or comment.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -192,30 +192,59 @@ private void SetAuthMethod()
...
+                    if (callbackResult.TokenCompatibleObject is TokenRequest tokenRequest)
                     {
-                        postData = GetTokenRequest(callbackResult);
-                        request.Url = $"/keys/{postData.KeyName}/requestToken";
+                        postData = tokenRequest;
+                        request.Url = RequestTokenUrl(postData.KeyName);
+                    }
+                    else if (callbackResult.TokenCompatibleObject is string tokenString)
+                    {
+                        postData = TokenRequestFromString(tokenString);
```
paddybyers: I think you're misunderstanding the `tokenString` case. It looks like everywhere you're treating it as a serialised `TokenRequest` but it's not; it is a token string. So it's for the case that the `authCallback`/`authURL` obtained a `TokenDetails` and then returned **only** the token string from within that (ie https://docs.ably.io/client-lib-development-guide/features/#TD2).

Here it means that you will probably use the token string to construct a `TokenDetails` (but where the TTL etc are unknown) and proceed without calling `requestToken()`.

Below there are other situations where I think you're handled it as a serialised `TokenRequest`.

withakay: Ah OK, the pre-existing code did the same and I assumed it was correct. More fool me! I shall take another pass.

withakay:
@paddybyers I have updated the code based on your comments RE the token string. 
When I create the `TokenDetails` and setting the expiry to 1 day based on the assumption the real expiry will be less. Let me know what you think?

withakay:
@funkyboy can you please re-review when you get a chance?

Issue #204, opened 2018-04-05
Presence sync doesn't complete on the first chance


ashikns:
Presence WaitForSyncAsync has a bug where the internal PresenceMap completes initial sync, but this is not picked up by WaitForSyncAsync because the relevant event was fired InitialSyncComplete has been set to true. Currently sync will only be marked as complete after a subsequent Map.StartSync() and Map.EndSync() has been called. This causes a delay of few seconds in getting channel presence.

Exact details and a fix can be found at: https://github.com/ashikns/ably-dotnet/commit/d764550a3c01b10af53ef19830a081cac946f6f5

paddybyers:
Hi, thanks for looking into this and for the contribution. The maintainer of this library is away until next week but we'll take a look as soon as we can.

withakay:
@ashikns thanks for your work, would you like to make a pull request so we can look at getting this integrated?

PR #206, opened 2018-04-12
Fix delay in initial presence sync

ashikns:
This fixes the delay caused due to WaitForSyncAsync not picking up sync complete the first time, described in issue https://github.com/ably/ably-dotnet/issues/204

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -128,6 +128,7 @@ namespace IO.Ably.Realtime
             _channel.StateChanged += OnChannelOnStateChanged;
             InitialSyncCompleted += OnSyncEvent;
             Map.SyncNoLongerInProgress += OnSyncEvent;
+            Map.InitialSyncHasCompleted += OnSyncEvent;
             // Do a manual check in case we are already in the desired state
             CheckAndSet();
             await tsc.Task;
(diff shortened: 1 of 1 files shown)


withakay:
Looks good to me 👍 

A note to other the reviewers: CI checks don't seem to be running for PRs from other repositories (which I will look into when I can). In the mean time I have cloned @ashikns repo and ran the tests locally, without issue.

paddybyers:
This looks OK to me, but in all the other PresenceMap implementations there's just a single member that indicates that a sync is in progress (eg https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/realtime/Presence.java#L1013). Why in this case do we need two members (`_isSyncInProgress`, `_initialSyncCompleted`) ? I don't see why we can't replicate the functionality without having both.

ashikns:
You could technically fix this by setting _InitialSyncCompleted = true;_ earlier, for example inside the try block. It would be lesser code, but it looked kinda odd to me and in the end I just decided to mirror code in the parent class.

withakay:
@paddybyers 

> I don't see why we can't replicate the functionality without having both.

You are probably right, the `InitialSyncCompleted` parts were added by myself when I added WaitForSyncAsync(), refactoring those parts is probably beyond the scope of this PR. I could add a new issue to look at cleaning up the PresenceMap implementation at late date, WDYT?

paddybyers:
> I could add a new issue to look at cleaning up the PresenceMap implementation at late date, WDYT?

I'm ok with that; @withakay ?

Thanks

withakay:
@paddybyers I made a new issue to look at the PresenceMap implementation, are you now okay to approve this PR?

paddybyers:
> are you now okay to approve this PR?

yes, thanks

PR #212, opened 2018-04-24
RSA4c+d

withakay:
Changes and tests to cover RSA4c (1,2,3) and RSA4d

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -225,22 +225,23 @@ namespace IO.Ably
                     else
                     {
                         shouldCatch = false;
-                        throw new AblyException($"AuthCallback returned an unsupported type ({callbackResult.GetType()}. Expected either TokenDetails or TokenRequest", 80019);
+                        throw new AblyException($"AuthCallback returned an unsupported type ({callbackResult.GetType()}. Expected either TokenDetails or TokenRequest", 80019, HttpStatusCode.BadRequest);
                     }
                 }
                 catch (Exception ex) when (shouldCatch)
                 {
-                    HttpStatusCode? statusCode = null;
-                    int code = 80019;
-                    if (ex is AblyException ablyException)
+                    var statusCode = HttpStatusCode.Unauthorized;
+                    if (ex is AblyException aex)
                     {
-                        statusCode = ablyException.ErrorInfo.StatusCode;
-                        code = ablyException.ErrorInfo.Code == 40300 ? ablyException.ErrorInfo.Code : 80019;
+                        statusCode = aex.ErrorInfo.StatusCode == HttpStatusCode.Forbidden
+                            ? HttpStatusCode.Forbidden
+                            : HttpStatusCode.Unauthorized;
                     }
 
                     throw new AblyException(
                         new ErrorInfo(
-                        "Error calling AuthCallback, token request failed. See inner exception for details.", code, statusCode), ex);
+                        "Error calling AuthCallback, token request failed. See inner exception for details.", 80019,
+                        statusCode), ex);
                 }
             }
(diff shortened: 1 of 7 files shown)


src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -261,66 +264,143 @@ private static TokenParams CreateTokenParams(Capability capability, TimeSpan? tt
...
+                }
+
+                TaskCompletionAwaiter tca = new TaskCompletionAwaiter(1000);
+                var realtimeClient = await GetRealtimeClient(protocol, Options);
+
+                realtimeClient.Connect();
+                await realtimeClient.WaitForState(ConnectionState.Connected);
+                realtimeClient.Connection.On(change =>
+                {
+                    // this callback should not be called
```
funkyboy: If I am reading this correctly, there should be a `fail` statement here. No?

withakay: @cesare the failure is on line 353 `(await tca.Task).Should().BeFalse(context);`

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -213,9 +213,17 @@ private void SetAuthMethod()
                 }
                 catch (Exception ex) when (shouldCatch)
                 {
+                    HttpStatusCode? statusCode = null;
+                    int code = 80019;
+                    if (ex is AblyException ablyException)
```
paddybyers: I don't think this implements the requirement. The outer error has code 80019 and it wraps the original error as `cause`. I don't know why this references 40300.

paddybyers: Where are you getting that text from?

The requirement I'm looking at is in https://github.com/ably/docs/blob/integration-spec-1-1/content/client-lib-development-guide/features.textile

The required check for RSA4d is `statusCode == 403`, not `code == xx`. And either way once you have the inner error you generate the ErrorInfo to return with `code = 80019, statusCode = <inner err>.statusCode, cause = <inner err>`. The difference when RSA4d applies is in the Realtime case you transition to `failed`.

withakay: That text is from the from the Google Sheet titled "Ably .NET Realtime Client Library Spec - v1.1"

withakay: @paddybyers fixed by https://github.com/ably/ably-dotnet/pull/212/commits/1d1d88a602c12d960807272fb6547ba5e4378ceb & https://github.com/ably/ably-dotnet/pull/212/commits/7572963e1334edaeec73c858730791962cc9cee9

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -1001,55 +997,10 @@ public async Task WhenFakeAuthMessageReceived_ShouldAttemptTokenRenewal(Protocol
...
-            var initialClientId = client.ClientId;
-
-            client.Connection.Once(ConnectionEvent.Disconnected, state2 =>
-            {
-                client.Connection.Once(ConnectionEvent.Connected, state3 =>
-                {
-                    reconnectAwaiter.SetCompleted();
-                });
-            });
+            await Task.Delay(1000);
```
paddybyers: Is this a "magic" delay?

withakay: There are points where some internal events are raised and if I recall it was possible to hit the next line before they had been handled, so this delay allow for that. So yes, somewhat magical.

Issue #205, opened 2018-04-11
Channel apparently successfully attaches, but OnAttachTimeout fires anyway


SimonWoolf:
Log from a customer (extracted from https://support.ably.io/helpdesk/tickets/2939 ):

```
IRealTimeClient.Channels.Get:

11:37:43.958 Setting up timer '#testchnl:Attached awaiter timer' to run action after 10 seconds.
11:37:43.959 #testchnl: Changing state to: 'Attaching'. 
11:37:43.960 RealtimeChannel.SendMessage:Attach
11:37:43.960 Current state: Connected. Sending message: { action=Attach }
11:37:43.962 Sending message (Attach) to transport

[Callback]Attaching: 

11:37:45.518 Proccessing queued message: { action=Attached }
11:37:45.518 #testchnl: Changing state to: 'Attached'. 
11:37:45.520 StartSync | Channel: testchnl, SyncInProgress: False

[Callback]Attached: 

11:37:50.005 Proccessing queued message: { action=Heartbeat }
11:37:53.960 Timer '#testchnl:Attached awaiter timer' interval 10 seconds elapsed and calling action.
11:37:53.963 #testchnl didn't Attach within 00:00:10. Setting state back to Attaching
11:37:53.964 #testchnl: Changing state to: 'Attaching'. Error: Reason: Channel didn't attach within the default timeout; Code: 50000
11:37:53.964 RealtimeChannel.SendMessage:Attach
11:37:53.964 Current state: Connected. Sending message: { action=Attach }
11:37:53.964 Sending message (Attach) to transport
11:37:53.972 Proccessing queued message: { action=Attached }
11:37:53.972 #testchnl: Changing state to: 'Attached'. 
11:37:53.972 StartSync | Channel: testchnl, SyncInProgress: True

[Callback]Attaching: Reason: Channel didn't attach within the default timeout; Code: 50000

[Callback]Attached:
```

SimonWoolf:
More information from the customer:
```
I am using v0.8.9 for the client lib in Unity.

I am also using an alternative websocket transport as shown in this repo.
https://github.com/ably-forks/ably-dotnet-alternative-transports

This is the websocket transport I am using.
https://github.com/sta/websocket-sharp
```

withakay:
Unity is not a supported platform (yet) but if we can can get some code to reproduce this we can take a look. Ideally a sample Unity project that demonstrates the issue, if possible?

yunxuchen:
[...]

PR #213, opened 2018-04-24
Issue 205

withakay:
A fix for issue #205.

A brief outline of the problem this fixes:

When `channel.Attach()` is called the channel moves into the `Attaching` state and and starts up a `ChannelAwaiter` that is waiting for the `Attach` state to be attained. Previously when that happened the callback fired but the timer was not cancelled. After the waiter times out (10 seconds as per DF1b), if the channel is attached this moves it to `Attaching` and the cycle starts again. Calling `Detach()` would clear the `AttachedAwaiter` but would start a `DetachedAwaiter` and a similar cycle would begin for Detached/Detaching.

The fix is quite simple as the `ChannelAwaiter` already listens to the Channels state changes, when the desired state is reached the timer is now aborted. I have also refactored this code a little to DRY it up and added a test.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelSandboxSpecs.cs
@@ -122,6 +122,24 @@ namespace IO.Ably.Tests.Realtime
             result.Error.StatusCode.Should().Be(HttpStatusCode.Unauthorized);
         }
 
+        [Theory]
+        [ProtocolData]
+        [Trait("bug", "Issue #205")]
+        public async Task ShouldReturnToPreviousStateIfAttachMessageNotReceivedWithinDefaultTimeout2(Protocol protocol)
+        {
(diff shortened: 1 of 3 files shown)


paddybyers:
I must be dim but I don't understand how the test works

src/IO.Ably.Tests.Shared/Realtime/ChannelSandboxSpecs.cs
```diff
@@ -732,6 +731,31 @@ await client.Channels.Get("test")
...
+            {
+                didAttach = b;
+                if (info != null)
+                {
+                    throw new Exception($"Attach returned an error: {info.Message}");
+                }
+            });
+            await Task.Delay(1000);
+            didAttach.Should().BeTrue();
+            channel.InternalStateChanged += (sender, change) => throw new AblyException(change.Error);
```
withakay: @paddybyers before the fix `channel.InternalStateChanged` would have fired after `RealtimeRequestTimeout` had elapsed and the channel state was changed to `Attaching`

PR #217, opened 2018-05-01
Hotfix for issue 216

withakay:
Basic support for `(TO3k6) fallbackHosts`

### DIFF
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -11,6 +11,7 @@ namespace IO.Ably
         private string _clientId;
         private Func<DateTimeOffset> _nowFunc;
         private bool _useBinaryProtocol = false;
+        private string[] _fallbackHosts;
 
         /// <summary>
         ///
@@ -81,6 +82,29 @@ namespace IO.Ably
         /// For development environments only. Allows a non default host for the rest service
         /// </summary>
         public string RestHost { get; set; }
+
+        /// <summary>
+        /// Gets or sets an array of custom Fallback hosts to be (optionally) used in place of the defaults.
+        /// If an empty array is specified, then fallback host functionality is disabled.
+        /// </summary>
+        public string[] FallbackHosts
+        {
+            get
+            {
+                if (_fallbackHosts is null)
+                {
+                    return Defaults.FallbackHosts;
+                }
+                return _fallbackHosts;
+            }
+            set => _fallbackHosts = value;
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to use default FallbackHosts even when overriding
+        /// environment or restHost/realtimeHost
+        /// </summary>
+        public bool FallbackHostsUseDefault { get; set; }
 
         internal bool IsLiveEnvironment => Environment.IsEmpty() || Environment == "live";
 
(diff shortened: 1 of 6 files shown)


src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -82,6 +83,29 @@ internal string GetClientId()
...
 
+        /// <summary>
+        /// Gets or sets an array of custom Fallback hosts to be (optionally) used in place of the defaults.
+        /// If an empty array is specified, then fallback host functionality is disabled.
+        /// </summary>
+        public string[] FallbackHosts
+        {
+            get
+            {
+                if (_fallbackHosts is null)
```
funkyboy: What if I pass an empty array here? Will this `if` still be true?

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -82,6 +83,29 @@ internal string GetClientId()
...
+                {
+                    return Defaults.FallbackHosts;
+                }
+                return _fallbackHosts;
```
funkyboy: Maybe I am reading this wrong but it seems you are returning the value before setting it. I'd expect the opposite, like set before you return.

src/IO.Ably.Tests.Shared/Rest/RestSpecs.cs
```diff
@@ -491,6 +491,61 @@ private AblyRest CreateClient(Action<ClientOptions> optionsClient)
...
+            {
+                _response.StatusCode = HttpStatusCode.BadGateway;
+                List<string> attemptedList = new List<string>();
+
+                var client = CreateClient(options => options.FallbackHosts = new[] { "www.example.com" });
+                await Assert.ThrowsAsync<AblyException>(() => MakeAnyRequest(client));
+                attemptedList.AddRange(_handler.Requests.Select(x => x.RequestUri.Host).ToList());
+
+                attemptedList.Count.Should().Be(2);
+                attemptedList[0].Should().Be("rest.ably.io");
```
funkyboy: Heads up: this will be "sandbox-rest.ably.io" if/when you'll test against sandbox.

src/IO.Ably.Shared/ClientOptions.cs
funkyboy: Guess this answers my question :) https://github.com/ably/ably-dotnet/pull/217/files#diff-fa73db3159959a90678f25e33b334dc4R519

src/IO.Ably.Shared/Http/AblyHttpClient.cs
```diff
@@ -108,10 +108,8 @@ internal void CreateInternalHttpClient(TimeSpan timeout, HttpMessageHandler mess
                         continue;
                     }
                     throw;
-                    //_host = hosts[currentTry - 1];
-                    
                 }
-                catch (TaskCanceledException ex) when (IsRetryableError(ex) && Options.IsDefaultHost)
+                catch (TaskCanceledException ex) when (IsRetryableError(ex) && (Options.IsDefaultHost || Options.FallbackHostsUseDefault))
                 {
                     Logger.Warning("Error making a connection to Ably servers. Retrying", ex);
```
funkyboy: It maybe more informative to print which Ably server was problematic to connect to

withakay: Good idea. 
I have changed this file substantially as part of work towards v1.0 so I would rather minimise changes here to avoid merge hell.
I have created an issue (#218) so I can address this later (along with another issue I noticed)

src/IO.Ably.Tests.Shared/Rest/RestSpecs.cs
withakay: This test is mocked, so not an issue here, but I will try to remember that!

mattheworiordan:
@funkyboy are you able to contact Neville with instructions on how to use this now from his servers?

funkyboy:
@mattheworiordan you mean him? https://github.com/ndastur

paddybyers:
@funkyboy yes

PR #221, opened 2018-05-04
RTN14a, RTN14b, RTN15h

withakay:
Updates to support the changes in the v1.0 spec to RTN14a, RTN14b and the addition of RTN15h.

### DIFF
--- a/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
+++ b/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
@@ -63,7 +63,8 @@ namespace IO.Ably.Transport.States.Connection
                 case ProtocolMessage.MessageAction.Error:
                     {
                         // If the error is a token error do some magic
-                        if (Context.ShouldWeRenewToken(message.Error))
+                        bool shouldRenew = Context.ShouldWeRenewToken(message.Error);
+                        if (shouldRenew)
                         {
                             try
                             {
@@ -92,7 +93,7 @@ namespace IO.Ably.Transport.States.Connection
                             return true;
                         }
 
-                        if (!Context.ShouldWeRenewToken(message.Error))
+                        if (message.Error?.IsTokenError == true && !shouldRenew )
                         {
                             TransitionState(new ConnectionDisconnectedState(Context, message.Error, Logger));
                             return true;
--- a/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
+++ b/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
@@ -62,6 +62,12 @@ namespace IO.Ably.Transport.States.Connection
 
                 case ProtocolMessage.MessageAction.Error:
                     {
+                        if (message.Error?.Code == 40400)
+                        {
+                            TransitionState(new ConnectionFailedState(Context, message.Error, Logger));
+                            return true;
+                        }
+
                         // If the error is a token error do some magic
                         if (Context.ShouldWeRenewToken(message.Error))
                         {
@@ -74,7 +80,7 @@ namespace IO.Ably.Transport.States.Connection
                             catch (AblyException ex)
                             {
                                 Logger.Error("Error trying to renew token.", ex);
-                                TransitionState(new ConnectionFailedState(Context, ex.ErrorInfo, Logger));
+                                TransitionState(new ConnectionDisconnectedState(Context, ex.ErrorInfo, Logger));
                                 return true;
                             }
                         }
@@ -83,6 +89,18 @@ namespace IO.Ably.Transport.States.Connection
                         {
                             Context.Connection.Key = null;
                             Context.HandleConnectingFailure(message.Error, null);
+                            return true;
+                        }
+
+                        if (message.Error?.IsTokenError == true && !Context.Connection.RestClient.AblyAuth.TokenRenewable)
+                        {
+                            TransitionState(new ConnectionFailedState(Context, message.Error, Logger));
+                            return true;
+                        }
+
+                        if (!Context.ShouldWeRenewToken(message.Error))
+                        {
+                            TransitionState(new ConnectionDisconnectedState(Context, message.Error, Logger));
                             return true;
                         }
 
--- a/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
+++ b/src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
@@ -62,12 +62,6 @@ namespace IO.Ably.Transport.States.Connection
 
                 case ProtocolMessage.MessageAction.Error:
                     {
-                        if (message.Error?.Code == 40400)
-                        {
-                            TransitionState(new ConnectionFailedState(Context, message.Error, Logger));
-                            return true;
-                        }
-
                         // If the error is a token error do some magic
                         if (Context.ShouldWeRenewToken(message.Error))
                         {
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -169,8 +169,7 @@ namespace IO.Ably.Tests.Realtime
         [Theory]
         [ProtocolData]
         [Trait("spec", "RTN14b")]
-        public async Task
-            WithExpiredRenewableToken_ShouldAutomaticallyRenewTokenAndNoErrorShouldBeEmitted(Protocol protocol)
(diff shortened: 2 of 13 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -259,29 +258,95 @@ public async Task
...
+                awaiter.Tick();
+            });
+
+            await awaiter.Task;
+
+            stateChanges.Should().HaveCount(2);
+            stateChanges[0].Current.Should().Be(ConnectionState.Disconnected);
+            stateChanges[0].Reason.Should().BeNull();
+            stateChanges[1].Current.Should().Be(ConnectionState.Connected);
+            stateChanges[0].Reason.Should().BeNull();
```
funkyboy: Should this be `stateChanges[1]` ?

withakay: Yes, will fix

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -259,29 +258,95 @@ public async Task
...
+        }
+
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RTN15h")]
```
mattheworiordan: @paddybyers if I recall https://docs.ably.io/client-lib-development-guide/features/#RTN15h is wrong and you were going to address that i.e. why do we move to the failed state as opposed to disconnect and retry?  Can you confirm?

src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
```diff
@@ -62,6 +62,12 @@ public override void Close()
 
                 case ProtocolMessage.MessageAction.Error:
                     {
+                        if (message.Error?.Code == 40400)
```
paddybyers: I don't understand where this condition comes from. There's nothing in the spec that singles out 40400 for special treatment.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
paddybyers: Correct = the behaviour for RTN15h has changed (since this was really a spec error in the 1.0 version). We should implement the 1.1 behaviour, which is to transition to `DISCONNECTED`; see https://github.com/ably/docs/pull/439

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailureSpecs.cs
```diff
@@ -71,6 +72,7 @@ public class ConnectingFailureSpecs : ConnectionSpecsBase
                 opts.Key = string.Empty;
                 opts.TokenDetails = tokenDetails;
                 opts.UseBinaryProtocol = false;
+                opts.LogLevel = LogLevel.Debug;
```
paddybyers: Was this intended to be temporary?

withakay: yes, fixed via https://github.com/ably/ably-dotnet/pull/221/commits/682509043d73af76a5a76d94539ebdb5bc169ca7

src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectingState.cs
withakay: That must have crept in when testing, it is benign but it is not required, removed in https://github.com/ably/ably-dotnet/pull/221/commits/cf56fe52653e89b26e0dbf3d5bcb41e7f3386ba1

withakay: @paddybyers I take it back, this code was added to support RTN14a, when I set an invalid key (the key is set to `baba.bobo:bosh`) I get a response from Ably with the error `{Reason: No application found with id baba; Code: 40400; HttpStatusCode: 404 (NotFound)}`. 
Is there a better way to deal with this?

withakay:
@paddybyers @funkyboy I have made changes to support RTN15h and also to support RTN14a via client side validation (previously the invalid key was being sent to the server and the error response handled). Can you please re-review?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -296,84 +296,124 @@ await GetRealtimeClient(protocol)
...
             });
 
             await awaiter.Task;
-
-            stateChanges.Should().HaveCount(2);
-            stateChanges[0].Current.Should().Be(ConnectionState.Disconnected);
-            stateChanges[0].Reason.Should().BeNull();
-            stateChanges[1].Current.Should().Be(ConnectionState.Connected);
-            stateChanges[1].Reason.Should().BeNull();
+            stateChanges.Should().HaveCount(3);
```
funkyboy: Here I'd test the full sequence, much like https://github.com/ably/ably-java/blob/develop/lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java#L659

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -296,84 +296,124 @@ await GetRealtimeClient(protocol)
...
         [ProtocolData]
         [Trait("spec", "RTN15h2")]
-        public async Task WhenDisconnectedMessageContainsTokenError_IfTokenRenewFails_ShouldBecomeDisconnectedEmitError(Protocol protocol)
+        public async Task WhenDisconnectedMessageContainsTokenError_IfTokenRenewFails_ShouldBecomeDisconnectedAndEmitError(Protocol protocol)
         {
-            var awaiter = new TaskCompletionAwaiter(10000, 2);
+            var awaiter = new TaskCompletionAwaiter(10000);
             var authClient = await GetRestClient(protocol);
+
             var tokenDetails = await authClient.AblyAuth.RequestTokenAsync(new TokenParams { ClientId = "123", Ttl = TimeSpan.FromSeconds(2) });
```
funkyboy: Heads up: this ttl value might be too low. In other test we use 5.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -258,7 +258,42 @@ await GetRealtimeClient(protocol)
...
+            {
+                options.TokenDetails = tokenDetails;
+                options.AutoConnect = false;
+            });
+
+            client.Connect();
+            await client.WaitForState(ConnectionState.Connected);
+
+            // null the key so the token is not renewable
+            client.Options.Key = null;
```
funkyboy: Sorry if I asked this already: is this method public only for the test context? Or also for the final user?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -258,7 +258,42 @@ await GetRealtimeClient(protocol)
...
+                awaiter.Tick();
+            });
+
+            client.Connection.Once(ConnectionEvent.Disconnected, state => throw new Exception("should not become DISCONNECTED"));
+            client.Connection.Once(ConnectionEvent.Connected, state => throw new Exception("should not become CONNECTED"));
+
+            await awaiter.Task;
+
+            client.Connection.State.Should().Be(ConnectionState.Failed);
+            client.Connection.ErrorReason.Should().NotBeNull();
```
funkyboy: Might want to test also for `code` and a string matching the `reason` error here.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: The access modifier is ‘internal’, it is not public and is not available to the end user.

paddybyers:
I'm confused now. The test `WhenDisconnectedMessageContainsTokenError_IfTokenIsRenewable_ShouldNotEmitError` doesn't look to me like the token is renewable. How does this succeed in reconnecting?

The test `WhenDisconnectedMessageContainsTokenError_IfTokenRenewFails_ShouldBecomeDisconnectedEmitError` verifies that it makes a single renewal attempt, and enters `failed` if that does not succeed. However, the spec (1.1) requires that that's not what happens: it will retry for as long as it has a means to attempt token renewal (provided that the error it gets back each time is a token error).

PR #223, opened 2018-05-22
RTN15i

withakay:
Changes to support RTN15i

`(RTN15i) If an ERROR ProtocolMessage is received, this indicates a fatal error in the connection. The server will close the transport immediately after. The client should transition to the FAILED state triggering all attached channels to transition to the FAILED state as well. Additionally the Connection#errorReason should be set with the error received from Ably`


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -341,6 +342,59 @@ await GetRealtimeClient(protocol)
...
+             Additionally the Connection#errorReason should be set with the error received from Ably
+             */
+
+            var client = await GetRealtimeClient(protocol, (options, settings) => { });
+            var channel = client.Channels.Get("RTN15i".AddRandomSuffix());
+            channel.Attach();
+            await channel.WaitForState(ChannelState.Attached);
+
+            List<ConnectionState> states = new List<ConnectionState>();
+            var errors = new List<ErrorInfo>();
```
funkyboy: Nitpicking. Can you make these consistent? Like both:
`var xyz = new ...` or `List<Type> xyz = new`

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFallbackSpecs.cs
```diff
@@ -154,6 +111,20 @@ await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.Mes
...
+        public async Task WithDefaultHostAndRecoverableError_ConnectionGoesToDisconnectedInsteadOfFailedAndRetryInstantly()
+        {
+            var client = GetConnectedClient();
+
+            await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Disconnected)
+            {
+                Error = new ErrorInfo() { StatusCode = HttpStatusCode.GatewayTimeout }
+            });
+
+            client.Connection.State.Should().Be(ConnectionState.Disconnected);
```
funkyboy: I wonder how this works. When you enter a disconnected state (assuming it's the first time you get disconnected), the client should stays there for a quite short time and then move to connecting (because of the retry immediately clause).
How does this test catch the exact moment in which the connection state is disconnected?

I mention this because we recently had a similar problem with ably-java, which we solved like this: https://github.com/ably/ably-java/commit/3c305dfee7a735580e8811c073b2c5f74f690e5f#diff-060685a947a7032665de7ac9d33bb6ffR1215

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: That is why we do reviews :)

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFallbackSpecs.cs
withakay: The state changes get fired off as tasks using the thread pool, so possibly I am getting lucky in what could be a potential race condition. But so far I have yet to see it fail. Additionally, if I modify the test so that it waits for CONNECTING after DISCONNECTED that always works as expected too.

withakay: > are realtime clients closing their connections when a test is over?

Explicitly, not usually (a quick search of the tests shows 8 occurrences of `Close()` being called on an `AblyRealtime` Client). 

For a couple of reasons I don't think this is nearly as bad as it sounds

1) A lot of the tests are entirely synthetic/mocked and don't use the sandbox, so don't actually hold a connection at all (as is the case with the test referenced here)

2) As the AblyRealtime client falls out of scope it become eligible for Garbage Collection, so in practice they are not retained for very long. 

I do think it would be *better* to explicitly call `Close()`

The idiomatic way would be to have `AblyRealtime`  use the [Disposable pattern](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern).

This would change the way Ably *should* be used in .Net, but it would not be a breaking change. It would look like this.

```
using (var client = new AblyRealtime())
{
    // do stuff with client

} // client.Dispose() called automatically when falls out of scope
```

The current way of instantiating the client (i.e. not wrapping in a `using` statement) would still work, but the IDE would likely highlight the line with a warning.

funkyboy: I think just calling `Close(`) is fine. 
That's meant to favor being a "good citizen" in "sandbox town". 
Sandbox will detect when a connection goes stale and will close it, but calling `Close()` from the client is more _correct_.

src/IO.Ably.Shared/Transport/States/Connection/ConnectionConnectedState.cs
```diff
@@ -36,27 +36,20 @@ public override void Close()
                     await Context.SetState(new ConnectionClosedState(Context, message.Error, Logger));
                     return true;
                 case ProtocolMessage.MessageAction.Disconnected:
-                    var error = message.Error;
-                    var result = await Context.RetryBecauseOfTokenError(error);
-                    if (result == false)
+                    if (await Context.CanUseFallBackUrl(message.Error))
                     {
-                        await Context.SetState(new ConnectionDisconnectedState(Context, message.Error, Logger));
+                        Context.Connection.Key = null;
```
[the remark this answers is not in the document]
withakay: That was preexisting logic I reused (see L55 in old file), as I recall removing that line caused problems, so I left it.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: fixed by https://github.com/ably/ably-dotnet/pull/223/commits/d8b5b520acea16752bf0440ce60ba9e3a2ebe743

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -341,6 +342,59 @@ await GetRealtimeClient(protocol)
...
+                {
+                    errors.Add(args.Reason);
+                }
+
+                states.Add(args.Current);
+            };
+
+            var dummyError = new ErrorInfo
+            {
+                Code = 40140,
```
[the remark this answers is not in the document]
withakay: fixed by https://github.com/ably/ably-dotnet/pull/223/commits/e002fb6a06ac76df2827f787b10932d30385a88a

withakay:
@paddybyers I have responded to the change requests, can you please re-review?

PR #226, opened 2018-06-06
RTN22 and RTN22a

withakay:
This PR covers RTN22 and RTN22a

`(RTN22) Ably can request that a connected client re-authenticates by sending the client an AUTH ProtocolMessage. The client must then immediately start a new authentication process as described in RTC8`

`(RTN22a) Ably reserves the right to forcibly disconnect a client that does not re-authenticate within an acceptable period of time, or at any time the token is deemed no longer valid. A client is forcibly disconnected following a DISCONNECTED message containing an error code in the range 40140 <= code < 40150. This will in effect force the client to re-authenticate and resume the connection immediately, see RTN15h`

As part of this work I have extracted `ConnectionInfo` and `IConnectionContext` from the `IConnectionManager.cs` file into discrete files. This is as per the style guide.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -985,12 +985,8 @@ namespace IO.Ably.Tests.Realtime
         [Theory]
         [ProtocolData]
         [Trait("spec", "RTN22")]
-        public async Task WhenFakeAuthMessageReceived_ShouldAttemptTokenRenewal(Protocol protocol)
+        public async Task WhenAuthMessageReceived_ShouldAttemptTokenRenewal(Protocol protocol)
         {
-            var authClient = await GetRestClient(protocol);
-            var almostExpiredToken = await authClient.AblyAuth.RequestTokenAsync(new TokenParams { ClientId = "123", Ttl = TimeSpan.FromMilliseconds(35) });
-
-            var reconnectAwaiter = new TaskCompletionAwaiter();
             var client = await GetRealtimeClient(protocol, (options, settings) =>
             {
                 options.UseTokenAuth = true;
@@ -1001,55 +997,10 @@ namespace IO.Ably.Tests.Realtime
             var initialToken = client.RestClient.AblyAuth.CurrentToken;
             var initialClientId = client.ClientId;
 
-            client.Connection.Once(ConnectionEvent.Disconnected, state2 =>
-            {
-                client.Connection.Once(ConnectionEvent.Connected, state3 =>
-                {
-                    reconnectAwaiter.SetCompleted();
-                });
-            });
-
             await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Auth));
-            var didReconect = await reconnectAwaiter.Task;
-            didReconect.Should().BeTrue();
-            client.RestClient.AblyAuth.CurrentToken.Should().NotBe(initialToken);
-            client.ClientId.Should().Be(initialClientId);
-            client.Close();
-        }
 
-        [Theory]
-        [ProtocolData]
-        [Trait("spec", "RTN22")]
-        public async Task WhenAuthMessageReceived_ShouldAttemptTokenRenewal(Protocol protocol)
-        {
-            var authClient = await GetRestClient(protocol);
+            await Task.Delay(1000);
 
-            var reconnectAwaiter = new TaskCompletionAwaiter(60000);
-            var client = await GetRealtimeClient(protocol, (options, settings) =>
-            {
-                options.AuthCallback = tokenParams =>
-                {
-                    var results = authClient.AblyAuth.RequestToken(new TokenParams { ClientId = "RTN22", Ttl = TimeSpan.FromSeconds(35) });
-                    return Task.FromResult<object>(results);
-                };
-                options.ClientId = "RTN22";
-            });
-
-            await client.WaitForState(ConnectionState.Connected);
-
-            var initialToken = client.RestClient.AblyAuth.CurrentToken;
-            var initialClientId = client.ClientId;
-
-            client.Connection.Once(ConnectionEvent.Disconnected, state2 =>
-            {
-                client.Connection.Once(ConnectionEvent.Connected, state3 =>
-                {
-                    reconnectAwaiter.SetCompleted();
-                });
-            });
-
-            var didReconect = await reconnectAwaiter.Task;
-            didReconect.Should().BeTrue();
             client.RestClient.AblyAuth.CurrentToken.Should().NotBe(initialToken);
             client.ClientId.Should().Be(initialClientId);
             client.Close();
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -282,7 +282,7 @@ namespace IO.Ably.Transport
             {
                 if (ShouldWeRenewToken(error))
                 {
-                    await RetryAuthentication();
+                    await RetryAuthentication(updateState: true);
                 }
                 else
                 {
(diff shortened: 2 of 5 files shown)


src/IO.Ably.Shared/IO.Ably.Shared.projitems
```diff
@@ -14,6 +14,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)AblyRealtime.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)AblyRest.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)ApiKey.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Transport\ConnectionInfo.cs" />
```
funkyboy: Other SDKs call this class `ConnectionDetails`. Would it be a breaking change to rename it?

src/IO.Ably.Shared/Transport/ConnectionInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+        public ConnectionInfo(ProtocolMessage message)
+        {
+            if (message == null)
+            {
+                throw new ArgumentNullException(nameof(message), "Null message");
+            }
+
+            if (message.Action != ProtocolMessage.MessageAction.Connected)
+            {
+                throw new InvalidOperationException("Can only create Connection info from Connected message. Current passed: " + message.Action);
```
funkyboy: Maybe something more specific like "You can create a ConnectionInfo only from a Connected action _protocol_ message. You passed a value with action ... "?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -756,6 +756,64 @@ public async Task
...
+
+            await client.WaitForState(ConnectionState.Connected);
+
+            var initialToken = client.RestClient.AblyAuth.CurrentToken;
+            var clientId = client.ClientId;
+            var t = client.Connection.ConnectionManager.Transport;
+
+            client.Connection.Once(ConnectionEvent.Connected, state =>
+            {
+                awaiter.SetCompleted();
```
funkyboy: I don't understand how this is triggered, given that you wait for `Connected` above.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -756,6 +756,64 @@ public async Task
...
+
+            await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Auth));
+            await awaiter.Task;
+
+            client.RestClient.AblyAuth.CurrentToken.Should().NotBe(initialToken);
+            client.ClientId.Should().Be(clientId);
+
+            client.Close();
+        }
+
```
funkyboy: I'd add also a test when a real `AUTH` message is sent by the server, e.g. https://github.com/ably/ably-js/blob/add-JWT-tests/spec/realtime/auth.test.js#L978

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -756,6 +756,64 @@ public async Task
...
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RTN22a")]
+        public async Task WhenDisconnectedMessageContainsTokenError_ForcesClientToReauthenticate(Protocol protocol)
+        {
+            var awaiter = new TaskCompletionAwaiter(10000, 2);
+            var client = await GetRealtimeClient(protocol, (options, settings) =>
+            {
+                options.UseTokenAuth = true;
+                options.AutoConnect = true;
```
funkyboy: This should be true by default, isn't it?

src/IO.Ably.Shared/Transport/ConnectionInfo.cs
withakay: The error message is pretty bad, FWIW I didn't write it, I just extracted this class from another file that. I will update it

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: The `Auth` ProtocolMessage (L781, below) triggers a re-authentication, this causes the connection to disconnect and connect (assuming the auth succeeds). Once the connection is reestablished we check the token has changed, proving the reauth occurred.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: That is a redundant option, I will remove it.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
funkyboy: I see. Then I think it reads better if you wait for connected, then disconnected, and then connected again, and then perform the check within the second reconnection callback. 
Similar to: https://github.com/ably/ably-ruby/blob/add-jwt-test/spec/acceptance/realtime/auth_spec.rb#L1171

withakay:
@funkyboy The second test has an error, it should be sending `ProtocolMessage.MessageAction.Disconnected` (not `Auth`). I will fix that.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: Something like this?

```
var reconnectedAwaiter = new TaskCompletionAwaiter();
var client = await GetRealtimeClient(protocol, (options, settings) =>
{
    options.UseTokenAuth = true;
});

var initialToken = client.RestClient.AblyAuth.CurrentToken;
var initialClientId = client.ClientId;

client.Connection.Once(ConnectionEvent.Disconnected, state2 =>
{
    client.Connection.Once(ConnectionEvent.Connected, state3 =>
    {
        awaiter.SetCompleted();
        client.RestClient.AblyAuth.CurrentToken.Should().NotBe(initialToken);
        client.ClientId.Should().Be(initialClientId);
    });
});

await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Auth));
var didReconnect = await reconnectedAwaiter.Task;
didReconnect.Should().BeTrue();
client.Close();
```
Personally that doesn't read any easier to me, but I *may* be biased and I am happy to do it as you like :)

funkyboy: Whether you keep it flat or nested is your call. I think the important part is testing the sequence: connected, disconnected, connected.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: I have added a test that causes a real `AUTH` message to be sent. Should I delete the tests with the Faked messages, it is mostly duplicated code?

withakay:
@funkyboy I think I have addressed all you comments, please can you re-review when you get a chance? Thanks.

funkyboy:
@withakay maybe this got lost? https://github.com/ably/ably-dotnet/pull/226#discussion_r194673849
Just curious. And certainly something we can address in a separate PR.

src/IO.Ably.Shared/IO.Ably.Shared.projitems
withakay: Technically that would be a breaking change, although I can't think why anyone would be using that 'in the wild'. 
There is actually already a class called `ConnectionDetails` in a in a file called [ConnectionDetailsMessage.cs](https://github.com/ably/ably-dotnet/blob/e76f6c94df952171a38bc84a2f5cfd1ac1a42e6f/src/IO.Ably.Shared/Types/ConnectionDetailsMessage.cs), this is essentially used a DTO. I expect there reason it was done this way is because JSON/MsgPack serialisation, but that is just a guess. I have created issue #231 to keep track of this

withakay:
@funkyboy oops, I have commented in line and created an issue to track: #231

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -756,6 +756,147 @@ public async Task
             await WaitForState(client, ConnectionState.Connecting);
         }
 
+        [Theory]
```
paddybyers: Perhaps I'm misreading this, but it looks like all these tests are expecting a server-sent `AUTH` to result in disconnection and reconnection. Is that correct?

withakay:
Note: There is a likely hood that there are changes in https://github.com/ably/ably-dotnet/pull/234 will impact on aspects of this PR that need attention

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
paddybyers: @withakay can you comment on this? It looks to me that this is still the case

withakay: You are reading the test correctly, a disconnect is expect by these tests. In the ConnectionManager, when a reauth is requested, it explicitly sets the state to `disconnected` and then `connecting`. Changing this impacts some other tests so I need to wrk on fixing those.

PR #228, opened 2018-06-08
RTN24, RTN21 and partial RTN4h

withakay:
`(RTN24) A connected client may receive a CONNECTED ProtocolMessage from Ably at any point (though is typically triggered by a reauth, see RTC8a). The connectionDetails in the ProtocolMessage must override any stored details, see RTN21. The Connection should emit an UPDATE event with a ChannelStateChange object, which should have both previous and current attributes set to CONNECTED, and the reason attribute set to to the error member of the CONNECTED ProtocolMessage (if any). (Note that UPDATE should be the only event emitted: in particular, the library must not emit an CONNECTED event if the client was already connected, see RTN4h).	`		

`(RTN21) If the CONNECTED ProtocolMessage contains a connectionDetails property, the attributes within ConnectionDetails will be used as the defaults for this client library, overriding any configured options at the time the CONNECTED ProtocolMessage is received`

`(RTN4h) It can emit an UPDATE event (this is the only ConnectionEvent which is not a ConnectionState). This is used for changes to connection conditions for which the ConnectionState (e.g. CONNECTED) does not change. (The library must never emit a ConnectionState event for a state equal to the previous state).`

The final part of RTN4h `The library must never emit a ConnectionState event for a state equal to the previous state` is not covered by these tests. 

There is a [pre-existing test for RTN8b](https://github.com/ably/ably-dotnet/blob/e76f6c94df952171a38bc84a2f5cfd1ac1a42e6f/src/IO.Ably.Tests/Realtime/ConnectionStateTests/ConnectedStateSpecs.cs#L119) that additionally covers part of RTN21 at a lower level.

### DIFF
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -116,30 +116,35 @@ namespace IO.Ably.Transport
                 {
                     lock (_stateSyncLock)
                     {
-                        if (State.State == newState.State)
+                        if (!newState.IsUpdate)
                         {
-                            if (Logger.IsDebug)
+                            if (State.State == newState.State)
                             {
-                                Logger.Debug($"xx State is already {State.State}. Skipping SetState action.");
+                                if (Logger.IsDebug)
+                                {
+                                    Logger.Debug($"xx State is already {State.State}. Skipping SetState action.");
+                                }
+
+                                return;
                             }
 
-                            return;
+                            AttemptsInfo.UpdateAttemptState(newState);
+                            // Abort any timers on the old state
+                            State.AbortTimer();
                         }
 
-                        // Abort any timers on the old state
-                        State.AbortTimer();
                         if (Logger.IsDebug)
                         {
                             Logger.Debug($"xx {newState.State}: BeforeTransition");
                         }
 
                         newState.BeforeTransition();
+
                         if (Logger.IsDebug)
                         {
                             Logger.Debug($"xx {newState.State}: BeforeTransition end");
                         }
 
-                        AttemptsInfo.UpdateAttemptState(newState);
                         Connection.UpdateState(newState);
                     }
 
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -814,6 +814,58 @@ namespace IO.Ably.Tests.Realtime
             client.RestClient.AblyAuth.CurrentToken.Should().NotBe(initialToken);
         }
 
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RTN24")]
+        [Trait("spec", "RTN21")]
+        [Trait("spec", "RTN4h")]
+        public async Task WhenConnectedMessageReceived_ShouldEmitUpdate(Protocol protocol)
+        {
+            var updateAwaiter = new TaskCompletionAwaiter(5000);
+            var client = await GetRealtimeClient(protocol, (options, settings) =>
+            {
+                options.UseTokenAuth = true;
+                options.AutoConnect = true;
+            });
+
+            await client.WaitForState(ConnectionState.Connected);
+
+            client.Connection.ConnectionStateTtl.Should().NotBe(TimeSpan.MaxValue);
+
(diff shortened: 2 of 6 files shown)


src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -116,30 +116,35 @@ public Task SetState(ConnectionStateBase newState, bool skipAttach = false)
...
                     lock (_stateSyncLock)
                     {
-                        if (State.State == newState.State)
+                        if (!newState.IsUpdate)
                         {
-                            if (Logger.IsDebug)
+                            if (State.State == newState.State)
                             {
-                                Logger.Debug($"xx State is already {State.State}. Skipping SetState action.");
+                                if (Logger.IsDebug)
```
funkyboy: This might become quite tedious to write. Have you considered just writing `Logger.Debug( ... )` and move the `if` check within `Logger.Debug( ... )`

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -814,6 +814,61 @@ public async Task
...
+        [Trait("spec", "RTN24")]
+        [Trait("spec", "RTN21")]
+        [Trait("spec", "RTN4h")]
+        public async Task WhenConnectedMessageReceived_ShouldEmitUpdate(Protocol protocol)
+        {
+            var updateAwaiter = new TaskCompletionAwaiter(5000);
+            var client = await GetRealtimeClient(protocol, (options, settings) =>
+            {
+                options.UseTokenAuth = true;
+                options.AutoConnect = true;
```
funkyboy: This shouldn't be needed

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -814,6 +814,61 @@ public async Task
...
+            });
+
+            await client.WaitForState(ConnectionState.Connected);
+
+            client.Connection.ConnectionStateTtl.Should().NotBe(TimeSpan.MaxValue);
```
funkyboy: Why is this relevant to test?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -814,6 +814,61 @@ public async Task
...
+            var key = client.Connection.Key;
+
+            client.Connection.Once(state =>
+            {
+                // RTN4h - can emit UPDATE event
+                if (state.Event == ConnectionEvent.Update)
+                {
+                    // should have both previous and current attributes set to CONNECTED
+                    state.Current.Should().Be(ConnectionState.Connected);
+                    state.Previous.Should().Be(ConnectionState.Connected);
```
funkyboy: I'd add also a check for the error field

src/IO.Ably.Shared/Transport/ConnectionManager.cs
withakay: I have, and this is a bit of an oddity, technically it is not needed but the code base already uses this method a lot so I have just been trying to be consistent for now. IT could be something that we clean up on the project level when higher priority things are out the way...?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: As per RTN21 `ConnectionStateTtl` should be set to `TimeSpan.MaxValue` via the `ConnectionDetails` later (line 852), this shows that it didn't start out at that value.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: Good spot, I will add a check for that.

src/IO.Ably.Shared/Transport/ConnectionManager.cs
funkyboy: Sure. Can you open an issue about it so we don’t forget?

withakay: Added issue #229 to cover this.

PR #230, opened 2018-06-12
RTN11b, RTN11c, RTN11d

withakay:
Changes and tests to cover RTNb, c & d

`(RTN11) Connection#connect function:`

`(RTN11b) If the state is CLOSING, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the CLOSED ProtocolMessage arrives for the old connection, it doesn’t affect the new one.`

`(RTN11c) If the state is DISCONNECTED or SUSPENDED, aborts the retry process described in RTN14d and RTN14e and immediately tries to reconnect.`

`(RTN11d) If the state is FAILED, transitions all the channels to INITIALIZED, sets their errorReason to null, and sets the connection’s errorReason to null.`

(RTN11a is omitted as it has a [pre-existing test](https://github.com/ably/ably-dotnet/blob/5816d286d203eb6a248153b42b0e6c15668586e6/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs#L37))

### DIFF
--- a/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
+++ b/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
@@ -121,6 +121,10 @@ namespace IO.Ably.Realtime
             switch (connectionStateChange.Current)
             {
                 case ConnectionState.Connecting:
+                    if (connectionStateChange.Previous == ConnectionState.Failed)
+                    {
+                        SetChannelState(ChannelState.Initialized);
+                    }
                     break;
                 case ConnectionState.Disconnected:
                     if (State == ChannelState.Attaching)
--- a/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
+++ b/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
@@ -93,7 +93,27 @@ namespace IO.Ably.Realtime
 
         private void SubscribeToConnectionEvents()
         {
+            ConnectionManager.Connection.BeginConnect += ConnectionOnBeginConnect;
             ConnectionManager.Connection.InternalStateChanged += InternalOnInternalStateChanged;
+        }
+
+        private void UnSubscribeFromConnectionEvents()
+        {
+            ConnectionManager.Connection.BeginConnect -= ConnectionOnBeginConnect;
+            ConnectionManager.Connection.InternalStateChanged -= InternalOnInternalStateChanged;
+        }
+
+        private void ConnectionOnBeginConnect(object sender, EventArgs e)
+        {
+            switch (ConnectionState)
+            {
+                case ConnectionState.Failed:
+                    /* (RTN11d)
+                     * If the [Connection] state is FAILED,
+                     * transitions all the channels to INITIALIZED */
+                    SetChannelState(ChannelState.Initialized);
+                    break;
+            }
         }
 
         internal void InternalOnInternalStateChanged(object sender, ConnectionStateChange connectionStateChange)
@@ -101,11 +121,6 @@ namespace IO.Ably.Realtime
             switch (connectionStateChange.Current)
             {
                 case ConnectionState.Connecting:
-                    if (connectionStateChange.Previous == ConnectionState.Failed)
-                    {
-                        SetChannelState(ChannelState.Initialized);
-                    }
-
                     break;
                 case ConnectionState.Disconnected:
                     if (State == ChannelState.Attaching)
@@ -375,6 +390,7 @@ namespace IO.Ably.Realtime
             DetachedAwaiter?.Dispose();
             _handlers.RemoveAll();
             Presence?.Dispose();
+            UnSubscribeFromConnectionEvents();
         }
 
         internal void AddUntilAttachParameter(HistoryRequestParams query)
@@ -502,7 +518,7 @@ namespace IO.Ably.Realtime
                             /* RTP1 If [HAS_PRESENCE] flag is 0 or there is no flags field,
                              * the presence map should be considered in sync immediately
                              * with no members present on the channel */
-                            Presence.SkipSync();
+                    Presence.SkipSync();
                         }
 
                         AttachedSerial = protocolMessage.ChannelSerial;
--- a/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
+++ b/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
@@ -100,7 +100,13 @@ namespace IO.Ably.Realtime
         {
             switch (connectionStateChange.Current)
             {
-                // case ConnectionState.Connected:
+                case ConnectionState.Connecting:
+                    if (connectionStateChange.Previous == ConnectionState.Failed)
+                    {
+                        SetChannelState(ChannelState.Initialized);
+                    }
+
+                    break;
                 case ConnectionState.Disconnected:
                     if (State == ChannelState.Attaching)
                     {
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -129,7 +129,6 @@ namespace IO.Ably.Transport
                             }
 
                             AttemptsInfo.UpdateAttemptState(newState);
-                            // Abort any timers on the old state
                             State.AbortTimer();
                         }
 
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -98,6 +98,7 @@ namespace IO.Ably.Transport
 
         public void Connect()
         {
+            Connection.OnBeginConnect();
             State.Connect();
         }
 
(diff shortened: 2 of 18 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -127,6 +128,119 @@ await GetRealtimeClient(protocol)
...
+            var awaiter = new TaskCompletionAwaiter(3000);
+            client.Connection.On(ConnectionEvent.Closed, (state) =>
+            {
+                awaiter.SetCompleted();
+            });
+
+            client.Close();
+            await client.WaitForState(ConnectionState.Closing);
+
+            client.Connect();
```
funkyboy: How do you know that when this gets executed the state is still `Closing`? It could be `Closed`.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -127,6 +128,119 @@ await GetRealtimeClient(protocol)
...
+        }
+
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RTN11c")]
+        public async Task WithSuspendedConnection_WhenConnectCalled_ImmediatelyReconnect(Protocol protocol)
+        {
+            var client = await GetRealtimeClient(protocol);
+            await client.WaitForState();
+            await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Disconnected));
```
funkyboy: Do we need this? Can't we initialize it in a suspended state right away?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -127,6 +128,119 @@ await GetRealtimeClient(protocol)
...
+            });
+
+            client.Close();
+            await client.WaitForState(ConnectionState.Closing);
+
+            client.Connect();
+            await client.WaitForState();
+
+            client.ConnectionManager.Transport.Should().NotBe(initialTransport);
+
```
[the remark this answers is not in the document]
withakay: What do you mean by 'the connection is not the initialConnection' here. The `Connection` class does not get reinitialised (i.e. `client.Connection`, but the underlying web socket transport and connection do. I thought this seemed correct, because if  the`Connection` class get reinitialised, that would wipe out any `Connection.On()` subscriptions.  But maybe that was an incorrect assumption on my part?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: L150 (above) listens for a `Closed` event and if handled will 'complete' the awaiter task.
Then, on L164, there is an assertion that the awaiter was not completed.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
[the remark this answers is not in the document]
funkyboy: You are right. The instance of `client.Connection` has to be the same, but the `connectionId` should be different and I'd add a test for that.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
withakay: No, not needed, I have simplified this now.

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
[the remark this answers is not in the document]
withakay: Good call, testing this has revealed a bug. Working to fix now.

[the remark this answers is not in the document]
withakay: I have updated the test and fixed the issue (I was not clearing the connection `key` so the connection was resumed

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -100,7 +100,13 @@ internal void InternalOnInternalStateChanged(object sender, ConnectionStateChang
         {
             switch (connectionStateChange.Current)
             {
-                // case ConnectionState.Connected:
+                case ConnectionState.Connecting:
+                    if (connectionStateChange.Previous == ConnectionState.Failed)
```
paddybyers: Shouldn't this be a check against the current state, not the `connectionStateChange.Previous` ? There might have been other state changes since the event was generated.

src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -129,7 +129,6 @@ public Task SetState(ConnectionStateBase newState, bool skipAttach = false)
                             }
 
                             AttemptsInfo.UpdateAttemptState(newState);
-                            // Abort any timers on the old state
```
paddybyers: I don't understand why this was removed.

src/IO.Ably.Shared/Transport/States/Connection/ConnectionClosingState.cs
```diff
@@ -10,6 +10,10 @@ internal class ConnectionClosingState : ConnectionStateBase
         private const int CloseTimeout = 1000;
         private readonly ICountdownTimer _timer;
 
+        // used to mitigate a potential race condition where by OnAttachToContext()
+        // can be called after Connect() is called but before the new state is attached
+        private bool _inTransition = false;
```
paddybyers: I think this needs clarifying. If `OnAttachToContext()` is called, but any action is suppressed because `_inTransition == true` then does that action occur at a later time?

If this variable is used only for the connect transition, and not for any other transition (eg closing -> closed) then it should be named accordingly (ie `inConnectTransition` or something).

src/IO.Ably.Shared/Transport/ConnectionManager.cs
withakay:
```
// Abort any timers on the old state
State.AbortTimer();
```

Although the comment is not incorrect, I don't think it adds anything to help our understanding of the code, the instance and method names are quite clear. If a comment isn't adding anything then it is just noise in the source (IMO).

paddybyers: ok

src/IO.Ably.Shared/Transport/States/Connection/ConnectionClosingState.cs
withakay: I have updated the variable name as suggested (the variable is only used for the connect transition) and changed the comments to use the xmldoc format, in most .Net IDEs that will give a code hint to the developer. See commit 0533cc0cd0c65043caf2d09382401774bd9b390b

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
withakay: >Shouldn't this be a check against the current state, not the `connectionStateChange.Previous` ?

I could add another condition to ensure we are working with the correct current state.

```
if (Connection.State == connectionStateChange.Current
    && connectionStateChange.Previous == ConnectionState.Failed)
 {
     SetChannelState(ChannelState.Initialized);
 }
```

But, re-reading the spec the RTN11d spec item `[when Connection#connect is called] If the state is FAILED, transitions all the channels to INITIALIZED...` I think strictly speaking what I have done is incorrect.

I have hooked into the existing state change mechanisms, the assumption being that when the state moves to `Connecting` that `Connection#connect` has been called and then retrospectively setting the channels.

Probably what I should do is add a new event that gets fired when `Connection#connect` is called, before any state transitions and use that to check the current state.

withakay: @paddybyers, updated to use a new `BeginConnect` event, see 08106a9f31dad8a1d9d0a85bfedbf9e4ccb778d4 and a60309cf70f79ccdac67ae00f8bcc358e12bbf92

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -502,7 +518,7 @@ private void HandleStateChange(ChannelState state, ErrorInfo error, ProtocolMess
                             /* RTP1 If [HAS_PRESENCE] flag is 0 or there is no flags field,
                              * the presence map should be considered in sync immediately
                              * with no members present on the channel */
-                            Presence.SkipSync();
+                    Presence.SkipSync();
```
paddybyers: Is there an indentation problem with the comment above as well?

withakay:
@paddybyers 

> In particular, it should make sure that, when the CLOSED ProtocolMessage arrives for the old connection, it doesn’t affect the new one.

There is a test for the behavior (the new connection does not get closed).
The way this has been implemented is that a new websocket transport is created and the old one is `Closed()` and `Disposed()` so I don't think it is possible to show that `CLOSED` message arrives.

https://github.com/ably/ably-dotnet/blob/c78fa5f88bff1050fe1f7b72e22fea36ef96628a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs#L168

PR #233, opened 2018-06-15
RSA10a

withakay:
`(RSA10a) Instructs the library to create a token immediately and ensures Token Auth is used for all future requests. See RTC8 for re-authentication behaviour when called for a realtime client`

### DIFF
--- a/src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
@@ -348,7 +348,19 @@ namespace IO.Ably.Tests
                     change.Reason.Code.Should().Be(80019);
                     tca.SetCompleted();
                 });
-                await realtimeClient.Auth.AuthorizeAsync();
+
+                bool didThrowAblyException = false;
+                try
+                {
+                    await realtimeClient.Auth.AuthorizeAsync();
+                    Assert.True(false, "An exception should be raised before this line is reached.");
+                }
+                catch (AblyException e)
+                {
+                    didThrowAblyException = true;
+                }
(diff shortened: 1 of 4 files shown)


src/IO.Ably.Tests.Shared/AuthTests/AuthSandboxSpecs.cs
```diff
@@ -509,9 +521,16 @@ void AuthCallbackOptions(ClientOptions options, TestEnvironmentSettings settings
...
+            var tokenDetails1 = await ably.Auth.AuthorizeAsync(new TokenParams() { ClientId = "123" }, new AuthOptions());
             ably.AblyAuth.ClientId.Should().Be("123");
+
+            // uses Token Auth for all future requests (RSA10a)
             ably.AblyAuth.AuthMethod.Should().Be(AuthMethod.Token);
+
+            // create a token immediately (RSA10a)
+            // regardless of whether the existing token is valid or not
+            var tokenDetails2 = await ably.Auth.AuthorizeAsync(new TokenParams() { ClientId = "123" }, new AuthOptions());
+            tokenDetails1.Token.Should().NotBe(tokenDetails2.Token);
```
funkyboy: Why not testing also the clientId here?

withakay: I was concerned with adding cover for the ‘create a token immediately’ part. 
I can update this to use another clientId in the second call and check that is set?

funkyboy: There’s no need for that, but while you are at it I’d check that tokenDetails2 has the clientId you asked for.

PR #232, opened 2018-06-13
RTN12d

withakay:
A test to cover RTN12d

`(RTN12d) If the connection state is DISCONNECTED or SUSPENDED, aborts the retry process described in RTN14d and RTN14e and transitions the connection immediately to the CLOSED state		`

PR #234, opened 2018-06-18
RTC8 and RTC8a subsections

withakay:
~~NOTE: This PR was based on branch `RSA10a` but needed changes on branch `RSA12d`, so I created the `RTC8` branch and rebased that on `RSA12d`, as such this probably doesn't make sense to review until both `RSA10a` and `RSA12d` are merged.~~

PR for RSA10a here:  #233 
PR for RSA12d here: #232

**The above PRs are now merged, so PR is ready for review**

This PR covers the following spec items

`(RTC8) For a realtime client, Auth#authorize instructs the library to obtain a token using the provided tokenParams and authOptions and upgrade the current connection to use that token; or if not currently connected, to connect with the token. `
           
`(RTC8a) If the connection is in the CONNECTED state and auth#authorize is called or Ably requests a re-authentication (see RTN22), the client must obtain a new token, then send an AUTH ProtocolMessage to Ably with an auth attribute containing an AuthDetails object with the token string`

`(RTC8a1) If the authentication token change is successful, then Ably will send a new CONNECTED ProtocolMessage. The connectionDetails provided in the CONNECTED ProtocolMessage must override any existing defaults, see RTN21. The Connection should emit an UPDATE event per RTN24. A test should exist that performs an upgrade of capabilities without any loss of continuity or connectivity during the upgrade process. Another test should exist where the capabilities are downgraded resulting in Ably sending an ERROR ProtocolMessage with a channel property, causing the channel to enter the FAILED state. That test must assert that the channel becomes failed soon after the token update and the reason is included in the channel state change event `

`(RTC8a2) If the authentication token change fails, then Ably will send an ERROR ProtocolMessage triggering the connection to transition to the FAILED state. A test should exist for a token change that fails (such as sending a new token with an incompatible clientId) `

`(RTC8a3) The authorize call should be indicated as completed with the new token or error only once realtime has responded to the AUTH with either a CONNECTED or ERROR respectively.  
`

### DIFF
--- a/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
+++ b/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
@@ -93,7 +93,27 @@ namespace IO.Ably.Realtime
 
         private void SubscribeToConnectionEvents()
         {
+            ConnectionManager.Connection.BeginConnect += ConnectionOnBeginConnect;
             ConnectionManager.Connection.InternalStateChanged += InternalOnInternalStateChanged;
+        }
+
+        private void UnSubscribeFromConnectionEvents()
+        {
+            ConnectionManager.Connection.BeginConnect -= ConnectionOnBeginConnect;
+            ConnectionManager.Connection.InternalStateChanged -= InternalOnInternalStateChanged;
+        }
+
+        private void ConnectionOnBeginConnect(object sender, EventArgs e)
+        {
+            switch (ConnectionState)
+            {
+                case ConnectionState.Failed:
+                    /* (RTN11d)
+                     * If the [Connection] state is FAILED,
+                     * transitions all the channels to INITIALIZED */
+                    SetChannelState(ChannelState.Initialized);
+                    break;
+            }
         }
 
         internal void InternalOnInternalStateChanged(object sender, ConnectionStateChange connectionStateChange)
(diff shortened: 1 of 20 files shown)


src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -100,7 +100,13 @@ internal void InternalOnInternalStateChanged(object sender, ConnectionStateChang
         {
             switch (connectionStateChange.Current)
             {
-                // case ConnectionState.Connected:
+                case ConnectionState.Connecting:
+                    if (connectionStateChange.Previous == ConnectionState.Failed)
```
paddybyers: Is it guaranteed that the current state is the same as `connectionStateChange.Previous` ?

ashikns:
Hey guys sorry for hijacking the PR. We're working on something with a rather tight deadline (~10 days). We've been using Ably as our backend for a few months now and it has mostly worked without any issues, but I wanted to get your opinion on the stability of the dotnet library. If I wanted the most stable version of Ably dotnet library, what would be my best choice right now? Do I just go with 0.8.11 release? Does the PR's for 1.0 release bring any stability improvements?

withakay:
@ashikns you can always assume the latest stable release on NuGet is the version we recommend, which is currently 0.8.11.

withakay:
@paddybyers the Prs mentioned in the description have now been merged so this is ready for another look.

PR #238, opened 2018-07-25
RSC19

withakay:
A PR to cover RSC19 and all subsections as per the below spec:

(RSC19) RestClient#request function is provided as a convenience for customers who wish to use bleeding edge REST API functionality that is either not documented or is not included in the API for our client libraries. The REST client library provides a function to issue HTTP requests to the Ably endpoints with all the built in functionality of the library such as authentication, paging, fallback hosts, MsgPack and JSON support etc. The function:

- (RSC19a) Method signature is request(string method, string path, Dict<String, String> params?, JsonObject | JsonArray body?, Dict<String, String> headers?) -> HttpPaginatedResponse with arguments: method is a valid HTTP verb (must support "GET", "POST", and "PUT", should support "PATCH", may support others); path is the path component of the URL such as "/channels"; params and headers are optional arguments containing pairs of key value strings (multi-valued headers are not supported) that will result in query params and HTTP headers being added respectively in the request (the argument types can be idiomatic for the language such as Object in the case of Javascript); body is an optional JsonObject or JsonArray like object argument that can be easily serialized to MsgPack or JSON
- (RSC19b) All requests will unconditionally use the default authentication mechanism configured for the REST client i.e. basic or token authentication (see Auth)
- (RSC19c) The library will configure the Accept and Content-Type type headers to reflect whether the client is configured to use a binary or JSON based protocol (see RSC8). All requests are encoded and decoded into Json or MsgPack as appropriate automatically by the library. Binary body payloads are not supported at this time
- (RSC19d) request method returns an HttpPaginatedResponse object that inherits from the PaginatedResult object to provide details on the response plus paging support where applicable. See HP1 for more details
- (RSC19e) If the HTTP network request fails for reasons such as a timeout (after the underlying fallback host attempts have failed where applicable, see RSC15), then the request method should indicate an error in an idiomatic way for the platform_


src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -273,6 +273,12 @@ private Result DecodePayload(IMessage payload, ChannelOptions options)
             return result;
         }
 
+        public HttpPaginatedResponse ParseHttpPaginatedResponse(AblyRequest request, AblyResponse response, PaginatedRequestParams requestParams, Func<PaginatedRequestParams, Task<HttpPaginatedResponse>> executeDataQueryRequest)
+        {
+            LogResponse(response);
```
funkyboy: Should this be wrapped by a check on the log level?

src/IO.Ably.Shared/PaginatedResult.cs
```diff
@@ -5,32 +5,48 @@
...
+        protected Func<PaginatedRequestParams, Task<PaginatedResult<T>>> ExecuteDataQueryFunc { get; }
 
         public List<T> Items { get; set; } = new List<T>();
 
-        private PaginatedResult()
+        public PaginatedRequestParams NextDataQuery { get; protected set; }
+
+        public PaginatedRequestParams FirstDataQuery { get; protected set; }
+
+        public PaginatedRequestParams CurrentQuery { get; protected set; }
```
funkyboy: First 2 are `DataQuery`, last is just query. Might be nicer if the naming is more consistent.

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
```diff
@@ -0,0 +1,248 @@
...
+            var client = TrackLastRequest(await GetRestClient(protocol));
+            try
+            {
+                var paginatedResponse = await client.Request(HttpMethod.Post, "/does-not-exist");
+                paginatedResponse.ErrorMessage.Should().NotBeNullOrEmpty();
+            }
+            catch (AblyException e)
+            {
+                e.ErrorInfo.Code.Should().Be(40400);
+                e.ErrorInfo.Message.Should().NotBeNullOrEmpty();
```
funkyboy: Would be nice to check also for the `message` here.

funkyboy:
@withakay I forgot about this. It'd be nice to have a test that spans across pages.
Like [this one in Ruby](https://github.com/ably/ably-ruby/blob/a7e9fce8cf1bcbfac3c94420b821f0cabd7f94e6/spec/acceptance/rest/client_spec.rb#L1002)

src/IO.Ably.Shared/PaginatedResult.cs
withakay: Those names were pre-existing, this code just got moved about. But yes, naming could be better. In fact them having `Data` is a bit meaningless. `FirstQueryParams`, `CurrentQueryParams` etc is perhaps more indicative of usage?

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
withakay: `LogResponse` is private to MessageHandler, it does the checking there

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
withakay: fixed by https://github.com/ably/ably-dotnet/pull/238/commits/9e656b6ede48a0229ff7fae0f83b7feea18a8fad

src/IO.Ably.Shared/PaginatedResult.cs
funkyboy: Yes, I agree.

withakay:
@funkyboy 

>It'd be nice to have a test that spans across pages.

How does this look? https://github.com/ably/ably-dotnet/pull/238/commits/9e656b6ede48a0229ff7fae0f83b7feea18a8fad

src/IO.Ably.Shared/PaginatedResult.cs
withakay: Improved by https://github.com/ably/ably-dotnet/pull/238/commits/b73dc05ce59bbcd305e87bd1054139e0e0a2bd03

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
```diff
@@ -0,0 +1,66 @@
...
+            var result = await client.Request(HttpMethod.Get, "/channels/" + channelName, testParams, null, testHeaders);
+            result.Should().NotBeNull();
+            result.StatusCode.Should().Be(HttpStatusCode.OK); // 200
+            result.Success.Should().BeTrue();
+            result.ErrorCode.Should().Be(0);
+            result.ErrorMessage.Should().BeNull();
+            result.Items.Should().HaveCount(1);
+            result.Items.First().Should().BeOfType<JObject>();
+            var channelDetails = result.Items.First() as JObject; // cast from JToken
+            channelDetails["id"].ToString().Should().BeEquivalentTo(channelName);
```
paddybyers: Note that this API has changed - now each `ChannelDetails` object returned has a `channelId` instead of an `id`. See the in-progress documentation for this API: https://github.com/ably/docs/blob/829ca206b8acca4129449c41b4bc6ec9dbe8c6fb/content/general/channel-lifecycle-api.md

src/IO.Ably.Shared/Http/AblyHttpClient.cs
```diff
@@ -62,7 +62,7 @@ public async Task<AblyResponse> Execute(AblyRequest request)
                 {
                     Logger.Error("Cumulative retry timeout of {0}s was exceeded", Config.CommulativeFailedRequestTimeOutInSeconds);
                     throw new AblyException(
-                        new ErrorInfo($"Commulative retry timeout of {Config.CommulativeFailedRequestTimeOutInSeconds}s was exceeded.", 500, null));
+                        new ErrorInfo($"Commulative retry timeout of {Config.CommulativeFailedRequestTimeOutInSeconds}s was exceeded.", 50000, null));
```
paddybyers: Typo: cumulative

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
```diff
@@ -12,55 +12,237 @@
...
+        [ProtocolData]
+        public async Task RequestFails_NotFound(Protocol protocol)
         {
+            var client = TrackLastRequest(await GetRestClient(protocol));
+            try
+            {
+                var paginatedResponse = await client.Request(HttpMethod.Post, "/does-not-exist");
+                paginatedResponse.ErrorMessage.Should().NotBeNullOrEmpty();
+            }
+            catch (AblyException e)
```
paddybyers: This is wrong. Unlike the regular REST APIs (stats, history), where a "failure" statusCode results in an error, this `request()` API is a thinner wrapper and if it performs a request, and gets a response, then that's a result which is passed back normally to the caller. Only if it attempts a request and fails to get a response should it present that as an error back to the caller. RSC19e isn't clear enough about this I guess.

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
```diff
@@ -203,13 +203,14 @@ public async Task RequestFails_NotFound(Protocol protocol)
             var client = TrackLastRequest(await GetRestClient(protocol));
             try
             {
-                var paginatedResponse = await client.Request(HttpMethod.Post, "/does-not-exist");
-                paginatedResponse.ErrorMessage.Should().NotBeNullOrEmpty();
+                await client.Request(HttpMethod.Post, "/does-not-exist");
+                throw new Exception("This should not be reached, the preceeding call should throw");
```
paddybyers: typo: preceding

withakay:
@paddybyers 

> It's assuming error statusCodes (ie non 2xx) result in an exception. That's not how it's supposed to work. Hopefully this doesn't screw up the DRYing up you did here with the other APIs.

I have resolved this by adding flag to the Request object to prevent Exception being raised for non 2XX responses, it ended up being quite a simple fix, see https://github.com/ably/ably-dotnet/pull/238/commits/e2df0f33abfbd28de8b6da5da4a59821104e4386

> Is it easy to add tests that mock the http endpoint so that all of the methods can be tested?

There are packages available to mock http servers (either by injecting an mock http client that return response objects or by actually spinning up an small http server and telling it how to respond) but I have not used them before.
I can't see the advantage  of doing either really as `Request` takes an `HttpMethod` enum that is passed to the built in .Net HttpClient, so we would just be proving the MS HttpClient works as expected. Maybe I am missing something?

withakay:
@paddybyers 

Hangfire on this, I have realised that there is a problem with the current code, `PATCH` is not supported by the `HttpMethod` enum across all the versions we need to support (it was added to the most recent .Net Core release), so I need to do some refactoring.

withakay:
@paddybyers 
I have added support and a test for the various verbs that should be supported, by using a mock http response handler. It somewhat contrived, but better than nothing.

Please review when you are able.

src/IO.Ably.Tests.Shared/Rest/RequestSandBoxSpecs.cs
```diff
@@ -244,6 +244,25 @@ public async Task RequestFails_ErrorConnecting(Protocol protocol)
...
+            {
+                options.HttpMaxRetryCount = 1;
+                options.HttpOpenTimeout = TimeSpan.FromSeconds(1);
+                options.HttpRequestTimeout = TimeSpan.FromSeconds(1);
+            }));
+
+            client.HttpClient.CustomHost = "echo.ably.io/respondwith?status=400";
+            var response = await client.Request(HttpMethod.Post, "/");
+            response.StatusCode.Should().Be(HttpStatusCode.BadRequest);
+            response.ErrorCode.Should().Be(0);
```
paddybyers: `ErrorCode` and `ErrorMessage` should be populated if they are present in the response - this test should verify that that happens. Also this test should also check the expected value in `success`.

withakay: This test is actually to demonstrate that for other error codes the underlying HTTP Client does not throw an exception. I have added comments to that effect. 

But you were right that those asserts were missing, just from another test further up the code file, that I have now updated. See https://github.com/ably/ably-dotnet/pull/238/commits/40957ef5bcc4591db356d81fed421e2665f8b7e0

withakay: @paddybyers ^

PR #241, opened 2018-07-27
RTL3e

withakay:
Update for RTL3e

- (RTL3e) If the connection state enters the DISCONNECTED state, it will have no effect on the channel states.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
@@ -232,6 +232,27 @@ namespace IO.Ably.Tests.Realtime
                 _channel.State.Should().Be(ChannelState.Suspended);
             }
 
+            [Theory]
+            [InlineData(ChannelState.Attached)]
+            [InlineData(ChannelState.Attaching)]
+            [InlineData(ChannelState.Failed)]
+            [InlineData(ChannelState.Suspended)]
+            [InlineData(ChannelState.Detached)]
+            [InlineData(ChannelState.Detaching)]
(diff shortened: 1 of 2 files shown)


src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -232,6 +232,27 @@ public class ConnectionStateChangeEffectSpecs : ChannelSpecs
...
+            [InlineData(ChannelState.Suspended)]
+            [InlineData(ChannelState.Detached)]
+            [InlineData(ChannelState.Detaching)]
+            [InlineData(ChannelState.Initialized)]
+            [Trait("spec", "RTL3e")]
+            public async Task WhenConnectionIsDisconnected_ChannelStateShouldNotChange(ChannelState state)
+            {
+                (_channel as RealtimeChannel).SetChannelState(state);
+
+                _client.Close();
```
funkyboy: Is this recycling an instance of realtime client? I'd prefer each test to have its own instance, to avoid clashes when tests are run in a sequence.

funkyboy: I'd address this in a separate PR.

withakay: I could add a `Reset()` call to each test method that re-instance the client and channel, that would be simple to do. WDYT?

funkyboy: Feels a bit of a patch. I'd prefer each test to follow the pattern:
- instantiate options
- instantiate client(s)
- perform test

PR #242, opened 2018-07-27
RTL3d

withakay:
Change to support RTL3d

- (RTL3d) If the connection state enters the CONNECTED state, then a SUSPENDED channel will initiate an attach operation and transition to ATTACHING. If the attach operation times out, the channel should return to the SUSPENDED state

Additionally refactored the clumsy method name `InternalOnInternalStateChanged` to `OnConnectionInternalStateChanged`

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
@@ -6,6 +6,8 @@ using FluentAssertions;
 using FluentAssertions.Execution;
 using IO.Ably.MessageEncoders;
 using IO.Ably.Realtime;
+using IO.Ably.Tests.Infrastructure;
+using IO.Ably.Transport;
 using IO.Ably.Transport.States.Connection;
 using IO.Ably.Types;
 using Xunit;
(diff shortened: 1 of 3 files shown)


src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -216,6 +218,37 @@ public class ConnectionStateChangeEffectSpecs : ChannelSpecs
...
+            [Trait("spec", "RTL3d")]
+            public async Task WhenChannelIsSuspended_WhenConnectionBecomeConnectedAttemptAttach()
+            {
+                var client = GetConnectedClient();
+                var channel = client.Channels.Get("test".AddRandomSuffix());
+
+                await client.ConnectionManager.SetState(new ConnectionSuspendedState(client.ConnectionManager, Logger));
+                await client.WaitForState(ConnectionState.Suspended);
+
+                (channel as RealtimeChannel).SetChannelState(ChannelState.Suspended);
```
funkyboy: I don't understand if we are in a mocked context here. If not, I'd wait for the channel go get into this state, instead of setting it explicitly.

withakay: This is using a mocked connection.

PR #243, opened 2018-07-31
RTL2b, RTL2f, RTL2g, RTL12 & RTL3d

withakay:
This PR covers the following spec items (apologies, this is quite a large one, but it need to be done)

- (RTL2f) When a channel ATTACHED ProtocolMessage is received, the ProtocolMessage may contain a RESUMED bit flag indicating that the channel has been resumed. The corresponding ChannelStateChange (either ATTACHED per RTL2a, or UPDATE per RTL12) will contain a resumed boolean attribute with value true if the bit flag RESUMED was included. When resumed is true, this indicates that the channel attach resumed the channel state from an existing connection and there has been no loss of message continuity. In all other cases, resumed is false. A test should exist to ensure that resumed is always false when a channel first becomes ATTACHED, it is true when the channel is ATTACHED following a successful connection recovery, and is false when the channel is ATTACHED following a failed connection recovery		

- (RTL2g) It can emit an UPDATE event (this is the only ChannelEvent which is not a ChannelState). This is used for changes to channel conditions for which the ChannelState (e.g. ATTACHED) does not change. (The library must never emit a ChannelState for a state equal to the previous state).		

- (RTL12) An attached channel may receive an additional ATTACHED ProtocolMessage from Ably at any point. (This is typically triggered following a transport being upgraded or resumed to indicate a partial loss of message continuity on that channel, in which case the ProtocolMessage will have a resumed flag set to false). If and only if the resumed flag is false, this should result in the channel emitting an UPDATE event with a ChannelStateChange object. The ChannelStateChange object should have both previous and current attributes set to attached, the reason attribute set to to the error member of the ATTACHED ProtocolMessage (if any), and the resumed attribute set per the RESUMED bitflag of the ATTACHED ProtocolMessage. (Note that UPDATE should be the only event emitted: in particular, the library must not emit an ATTACHED event if the channel was already attached, see RTL2g).

In addition to the above I had to update the the existing RTN19b tests, they were incorrect for the current spec, but passing. After changes required for the above spec items the existing tests failed. Thanks to @funkyboy for giving me a nudge run the right direction with those.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
@@ -52,45 +52,6 @@ namespace IO.Ably.Tests.Realtime
                 channel.Presence.Should().BeOfType<Presence>();
             }
 
-            [Fact]
-            [Trait("spec", "RTL12")]
-            [Trait("intermittent", "true")]
-            public async Task OnceAttachedWhenConsequentAttachMessageArriveWithError_ShouldEmitErrorOnChannelButNoStateChange()
-            {
-                var client = GetConnectedClient();
-                var channel = client.Channels.Get("test");
-                SetState(channel, ChannelState.Attached);
-
-                ErrorInfo expectedError = new ErrorInfo();
-                ErrorInfo error = null;
-                channel.Error += (sender, args) =>
-                {
-                    error = args.Reason;
-                };
-                bool stateChanged = false;
-
-                await Task.Delay(10); // Allow the notification thread to fire
-
-                ChannelState newState = ChannelState.Initialized;
-                channel.On((args) =>
-                {
-                    stateChanged = true;
-                    newState = args.Current;
-                });
-
-                await
-                    client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Attached)
-                    {
-                        Error = expectedError,
-                        Channel = "test"
-                    });
-
-                await Task.Delay(10); // Allow the notification thread to fire
-
-                error.Should().BeSameAs(expectedError);
-                stateChanged.Should().BeFalse("State should not have changed but is now: " + newState);
-            }
-
             public GeneralSpecs(ITestOutputHelper output)
                 : base(output)
             {
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
@@ -144,10 +144,13 @@ namespace IO.Ably.Tests.Realtime
                 var channel = client.Channels.Get("test");
                 var tsc = new TaskCompletionAwaiter(2000);
 
-                channel.Once(stateChange =>
+                channel.Once(ChannelEvent.Attached, stateChange =>
                 {
+                    // RTL2f
                     stateChange.Current.Should().Be(ChannelState.Attached);
+                    stateChange.Previous.Should().Be(ChannelState.Initialized);
                     stateChange.Resumed.Should().BeFalse();
+                    stateChange.Error.Should().BeNull();
                     tsc.SetCompleted();
                 });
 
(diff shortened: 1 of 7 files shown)


src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -174,6 +134,120 @@ public void ShouldEmmitErrorWithTheErrorThatHasOccuredOnTheChannel()
...
             }
 
+            [Fact]
+            [Trait("spec", "RTL2f")]
+            [Trait("spec", "RTL2g")]
+            [Trait("spec", "RTL12")]
+            async Task WhenAttachedProtocolMessageWithResumedFlagReceived_EmittedChannelStateChangeShouldIndicateResumed()
+            {
+                var client = GetConnectedClient();
+                var channel = client.Channels.Get("test");
```
funkyboy: At some point in the future this will bite you :)
I didn't see an issue about this, but I propose to have a helper function that generates random names for test channels (maybe that accepts a prefix param, useful when testing permissions). All this to avoid clashes between tests using the same channel name.
Definitely to be tackled in a separate PR.

withakay: @funkyboy Yes, usually I use the "channelName".AddRandomSuffix() extension that a previous developer added to the test project, but this is a mocked connection so I didn't see the need and I guess I am bit lazy ;)

But now you have pointed it out I think that I really should get in the habit of using it everywhere by default. You have permission to hit me if I do this again.

src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -174,6 +174,97 @@ public void ShouldEmmitErrorWithTheErrorThatHasOccuredOnTheChannel()
...
+                channel.State.Should().Be(ChannelState.Attached);
+
+                channel.Once(stateChange => throw new Exception("This should not be reached because resumed flag was set"));
+
+                // send another Attached protocol message with the resumed flag set.
+                // This should not trigger any channel event (per RTL12).
+                await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.MessageAction.Attached)
+                {
+                    Channel = "test",
+                    Flags = 4 // resumed
```
paddybyers: You could define an enum for this

src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -174,6 +174,97 @@ public void ShouldEmmitErrorWithTheErrorThatHasOccuredOnTheChannel()
...
+                var channel = client.Channels.Get("test") as RealtimeChannel;
+                bool stateDidChange = false;
+
+                // set initial state
+                channel.SetChannelState((ChannelState)state);
+
+                channel.Once(stateChange =>
+                {
+                    stateDidChange = true;
+                    throw new Exception("state chage event should not be emitted");
```
paddybyers: typo: chage

src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
```diff
@@ -193,6 +195,27 @@ await client.FakeProtocolMessageReceived(new ProtocolMessage(ProtocolMessage.Mes
                 });
 
                 await Task.Delay(2000);
+                tsc = new TaskCompletionAwaiter(500);
+
+                // set detached so that the nexted Attached message with resume will pass
```
paddybyers: typo: nexted

src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
withakay: fixed by https://github.com/ably/ably-dotnet/pull/243/commits/d9338768c78842e0eab3aaace77a1ad9b825adcd

src/IO.Ably.Tests.Shared/Realtime/ChannelSpecs.cs
withakay: fixed by https://github.com/ably/ably-dotnet/pull/243/commits/33c3afc0c6b91ec20633f22c9bcbc2912169d51a

PR #244, opened 2018-08-01
RTL11 & RTL11a

withakay:
This PR covers RTL11 and RTL11a:

- (RTL11) If a channel enters the DETACHED, SUSPENDED or FAILED state, then all messages that are still queued for send on that channel should be deleted from the queue triggering a failure for the publish or presence methods invoked for those messages
- (RTL11a) For clarity, any messages awaiting an ACK or NACK are unaffected by channel state changes i.e. a channel that becomes detached following an explicit request to detach may still receive an ACK or NACK for messages published on that channel later


funkyboy:
@withakay Sorry, I missed that part. .NET tooling for tests is cool :)

PR #249, opened 2018-08-24
RSL1k  - Idempotent rest publishing

withakay:
(RSL1k) Idempotent publishing via REST is supported by populating the id attribute of Messageinstances passed to publish():	
	
- (RSL1k1) Idempotent publishing via library-generated Message @id@s is supported if idempotentRestPublishing (see TO3n) is enabled and one or more Message instances are passed to publish() and all @Message@s have an empty id attribute. The library generates a base id string by base64-encoding a sequence of at least 9 bytes obtained from a source of randomness. Each individual Message in the set of messages to be published is assigned a unique id of the form <base id>:<serial> (where serial is the zero-based index into the set).	

- (RSL1k2) Idempotent publishing via client-supplied Message @id@s is supported where a single Message is passed to publish() and it contains a non-empty id. The id is perserved on sending the message.	

- (RSL1k3) If more than one Message is passed to publish() and one or more of those messages contains a non-empty id attribute, then all message ids (present or absent) are preserved on sending the batch of messages.	

- (RSL1k4) An explicit test for idempotency of publishes with library-generated ids shall exist that simulates an error response to a successful publish of a batch of messages, expects an automatic retry by the library, and verifies that the net outcome is that the batch is published only once.	

- (RSL1k5) An explicit test for idempotency of publishes with client-supplied ids shall exist that involves multiple explicit publish requests for a given message and verifies that the net outcome is that the message is published only once.


withakay:
Please note the CI test failed because they run against sandbox, when run against dev all the idempotent tests pass.

src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
```diff
@@ -106,6 +107,144 @@ public async Task WithExplicitClientIdNotMatchingClientIdInOptions_ReturnsMessag
...
+            {
+                AssertMessage(m);
+            }
+        }
+
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSL1k2")]
+        [Trait("spec", "RSL1k3")]
+        public async Task IdempotentPublishing_ClientProvidedMessageIdsArePreserved(Protocol protocol)
```
paddybyers: I don't think this test is verifying that the ids are preserved and reach the Ably endpoint.

paddybyers: @withakay did you get a chance to look at that comment? ^

withakay: @paddybyers 
if I pull the channel history and check for the ids there does that cover it?

```
var history = await channel.HistoryAsync();
history.Items.Should().HaveCount(4);
history.Items[3].Id.Should().Be("RSL1k2");
history.Items[2].Id.Should().Be("RSL1k3");
```

paddybyers: > if I pull the channel history and check for the ids are present does that cover it?

yes

withakay: @paddybyers fixed by 
https://github.com/ably/ably-dotnet/pull/249/commits/8c8f83bdd7b5ee2af9bb33ca8534946611194a78

PR #255, opened 2018-10-04
RTP3

withakay:
- (RTP3) If a SYNC operation is underway but not yet complete, and the transport is disconnected unexpectedly, then if the connection is resumed successfully, it is the responsibility of the client library to complete the SYNC operation. The client library requests a SYNC resume by sending a SYNC ProtocolMessage with the channelSerial set to the same as the channelSerial of the most recently received SYNC.

I also did some slight refactoring for the Presence code, whilst implementing this feature, I appreciate this is not best practice and makes it harder to see what has changed but as Presence barely implemented hopefully reviewers can overlook it on this occasion!

**NOTE: CI is failing intermittently, I have submitted another PR #257 to help with that, please review that first and I can merge the changes down stream.**


paddybyers:
> CI is failing intermittently, I have submitted another PR #257 to help with that, please review that first and I can merge the changes down stream.

So do you want to merge/rebase first or should I review this PR now?

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -441,6 +441,75 @@ await client.Connection.ConnectionManager.OnTransportMessageReceived(new Protoco
...
+
+                await client.WaitForState(ConnectionState.Disconnected);
+                client.Connection.State.Should().Be(ConnectionState.Disconnected);
+
+                await client.WaitForState(ConnectionState.Connected);
+                client.Connection.State.Should().Be(ConnectionState.Connected);
+
+                await Task.Delay(500);
+
+                var messages = await channel.Presence.GetAsync();
```
paddybyers: Is there a test in here that the SYNC was resumed? It looks to me like this test would pass if the SYNC was restarted from the beginning.

withakay: I have updated to show that only 2 SYNC messages were sent (1 for each page) - If SYNC was restarted there would be 3 SYNC messages

PR #258, opened 2018-10-09
RTP11 and subsections

withakay:
Implements RTP11 with tests

(RTP11) Presence#get function:
- (RTP11a) Returns the list of current members on the channel in a callback. By default, will wait for the SYNC to be completed, see RTP11c1
- (RTP11b) Implicitly attaches the Channel if the channel is in the INITIALIZED state. However, if the channel is in or enters the DETACHED or FAILED state before the operation succeeds, it will result in an error
- (RTP11d) If the Channel is in the SUSPENDED state then the get function will by default, or if waitForSync is set to true, result in an error with code 91005 and a message stating that the presence state is out of sync due to the channel being in a SUSPENDED state. If however the get function is called with waitForSync set to false, then it immediately returns the members currently stored in the PresenceMap giving developers access to the members that were present at the time the channel became SUSPENDED
- (RTP11c) An optional set of params can be provided:
    - (RTP11c1) waitForSync (default true). When true, method will wait until SYNC is complete before returning a list of members. When false, known set of presence members is returned immediately, which may be incomplete if the SYNC is not finished
    - (RTP11c2) clientId filters members by the provided clientId
    - (RTP11c3) connectionId filters members by the provided connectionId

RTP11a "Returns the list of current members on the channel in a callback" - async/await is used so there is no callback as such. Async/await is more idiomatic and acheives the same result, but wrapper methods could be added if it is felt having a callback is important.

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -109,7 +109,7 @@ namespace IO.Ably.Realtime
             return await GetAsync(new GetParams() { WaitForSync = waitForSync });
         }
 
-        public async Task<IEnumerable<PresenceMessage>> GetAsync(string clientId, bool waitForSync = false)
+        public async Task<IEnumerable<PresenceMessage>> GetAsync(string clientId, bool waitForSync = true)
         {
             return await GetAsync(new GetParams() { ClientId = clientId, WaitForSync = waitForSync });
         }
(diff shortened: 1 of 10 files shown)


withakay:
@paddybyers I have added some comment labels are requested.

src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -429,11 +464,37 @@ internal void EndSync()
...
+                        try
+                        {
+                            /* Message is new to presence map, send it */
+                            var itemToSend = item.ShallowClone();
+                            itemToSend.Action = PresenceAction.Enter;
+                            UpdatePresence(itemToSend, (success, info) =>
+                            {
+                                if (!success)
+                                {
+                                    /*
```
paddybyers: indentation?

PR #261, opened 2018-10-15
RTP19, RTP19a and RTP6b

withakay:
Changes and tests for 

- (RTP19) If the PresenceMap has existing members when a SYNC is started, the client library must ensure that members no longer present on the channel are removed from the local PresenceMap once the sync is complete. In order to do this, the client library must keep track of any members that have not been added or updated in the PresenceMap during the sync process. Note that a member can be added or updated when received in a SYNC message or when received in a PRESENCE message during the sync process. Once the sync is complete, the members in the PresenceMap that have not been added or updated should be removed from the PresenceMap and a LEAVE event should be published for each. The PresenceMessage published should contain the original attributes of the presence member with the action set to LEAVE, PresenceMessage#id set to null, and the timestamp set to the current time. This behaviour should be tested as follows: ENTER presence on a channel, wait for SYNC to complete, inject a member directly into the local PresenceMap so that it only exists locally and not on the server, send a SYNC message with the channel attribute populated with the current channel which will trigger a server initiated SYNC. A LEAVE event should then be published for the injected member, and checking the PresenceMap should reveal that the member was removed and the valid member entered for this connection is still present
- - (RTP19a) If the PresenceMap has existing members when an ATTACHED message is received without a HAS_PRESENCE flag, the client library should emit a LEAVE event for each existing member, and the PresenceMessage published should contain the original attributes of the presence member with the action set to LEAVE, PresenceMessage#id set to null, and the timestamp set to the current time. Once complete, all members in the PresenceMap should be removed as there are no members present on the channel

Also

- (RTP6b) Subscribe with a single action argument – such as ENTER, LEAVE, UPDATE or PRESENT – subscribes a listener to receive only presence messages with that action

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -432,6 +432,7 @@ namespace IO.Ably.Realtime
 
         internal void EndSync()
         {
+            _currentSyncChannelSerial = null;
             var residualMembers = Map.EndSync();
 
             /*
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -544,17 +544,20 @@ namespace IO.Ably.Realtime
                 /* Start sync, if hasPresence is not set end sync immediately dropping all the current presence members */
                 StartSync();
                 _syncAsResultOfAttach = true;
-
-                // TODO: for v1.0 RTP19a (see Java version for example https://github.com/ably/ably-java/blob/159018c30b3ef813a9d3ca3c6bc82f51aacbbc68/lib/src/main/java/io/ably/lib/realtime/Presence.java)
-                // if (!hasPresence)
-                // {
-                // /*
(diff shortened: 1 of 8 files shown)


src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -483,24 +544,35 @@ private void OnChannelStateChanged(object sender, ChannelStateChange e)
...
-                // }
-                SendQueuedMessages();
+                var hasPresence = e.ProtocolMessage != null && e.ProtocolMessage.HasFlag(ProtocolMessage.Flag.HasPresence);
+                if (!hasPresence)
+                {
+                    /*
+                    * RTP19a  If the PresenceMap has existing members when an ATTACHED message is received without a
+                    * HAS_PRESENCE flag, the client library should emit a LEAVE event for each existing member ...
+                    */
+                        EndSync();
```
paddybyers: Indentation error?

withakay: yes, fixed by https://github.com/ably/ably-dotnet/pull/261/commits/d1213d73bde26e4dfd46eeab64e8d92f9d5c1c5b

PR #263, opened 2018-10-16
properties should be placed before the constructor

withakay:
Small style fix. Per the style guide, properties should be placed before the constructor.


withakay:
Self merged as there is no functional change to this code, just a formatting.

PR #265, opened 2018-10-26
RTP5a & RTP5b

withakay:
- (RTP5a) If the channel enters the DETACHED or FAILED state then all queued presence messages will fail immediately, and the PresenceMap and internal PresenceMap (see RTP17) is cleared. The latter ensures members are not automatically re-entered if the Channel later becomes attached. Since channels in the DETACHED and FAILED states will not receive any presence updates from Ably, presence events (specifically LEAVE) should not be emitted when the PresenceMap is cleared as each presence member's state is unknown		

- (RTP5b) If a channel enters the ATTACHED state then all queued presence messages will be sent immediately and a presence SYNC may be initiated by the Ably service or the Ably service may indicate there is no SYNC necessary

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
@@ -1178,6 +1178,52 @@ namespace IO.Ably.Tests.Realtime
 
                     client.Close();
                 }
+
+                [Theory]
+                [ProtocolData]
+                [Trait("spec", "RTP5b")]
+                public async Task WhenChannelBecomesAttached_ShouldSendQueuedMessagesAndInitiateSYNC(Protocol protocol)
+                {
+                    var client1 = await GetRealtimeClient(protocol);
+                    var client2 = await GetRealtimeClient(protocol);
+
+                    await client1.WaitForState();
+                    await client2.WaitForState();
+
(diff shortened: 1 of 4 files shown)


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -1097,6 +1097,134 @@ public void PresenceMessage_IsNewerThan_RaisesExceptionWithMalformedIdStrings()
...
+                    var channel = client.Channels.Get("RTP5a".AddRandomSuffix()) as RealtimeChannel;
+
+                    int initialCount = 0;
+                    bool? success = null;
+                    ErrorInfo errInfo = null;
+                    await WaitForMultiple(2, partialDone =>
+                     {
+                         // insert an error when attaching
+                         channel.Once(ChannelEvent.Attached, args =>
+                         {
```
paddybyers: Why `Attached` and not `Attaching` ?

withakay: I don't think I had a reason, it was just a typo in the comment, but I went back and tried with `Attaching` and the test failed, so good job flagging that. It was an order execution issue, I have made a couple of small changes that fix this.

withakay:
@paddybyers I have made some changes based on your comment and pushed them for your review. Apologies though, an update to Visual Studio appears to have nuked my extensions, unfortunately this means some white space formatting errors crept back. Sorry!

PR #279, opened 2019-01-16
tests for RTP5c2

withakay:
- (RTP5c2) If either a SYNC is initiated as part of the attach and the SYNC is complete, or if the resumed flag is false and a SYNC is not expected, all members not present in the PresenceMap but present in the internal PresenceMap RTP17 must be re-entered automatically by the client using the clientId and data attributes from each. The members re-entered automatically must be removed from the internal PresenceMap RTP17 ensuring that members present on the channel are constructed from presence events sent from Ably since the channel became ATTACHED

2 tests to cover the 2 initial clauses presented by the spec above:

1. If either a SYNC is initiated as part of the attach and the SYNC is complete
2. if the resumed flag is false and a SYNC is not expected


withakay:
This is failing CI but tests OK locally, needs investigation 🧐

PR #280, opened 2019-01-17
Test cover for RTP5c3

withakay:
- (RTP5c3) If any of the automatic ENTER presence messages published in RTP5c2 fail, then an UPDATE event should be emitted on the channel with resumed set to true and reason set to an ErrorInfo object with error code value 91004 and the error message string containing the message received from Ably (if applicable), the code received from Ably (if applicable) and the explicit or implicit client_id of the PresenceMessage

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
@@ -1223,6 +1223,82 @@ namespace IO.Ably.Tests.Realtime
 
                     client1.Close();
                 }
+
+                [Theory]
+                [ProtocolData]
+                [Trait("spec", "RTP5c3")]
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -1223,6 +1223,82 @@ public async Task WhenChannelBecomesAttached_ShouldSendQueuedMessagesAndInitiate
...
+                        channel.Once(ChannelEvent.Attaching, change =>
+                        {
+                            // insert local member to automatically try to enter
+                            channel.Presence.InternalMap.Put(localMember);
+                            partialDone();
+                        });
+
+                        channel.Presence.Subscribe(PresenceAction.Enter, message =>
+                        {
+                            enterMessage = message; // should not hit
```
paddybyers: How do we know this is not hit?

PR #282, opened 2019-01-19
RTP16b

withakay:
- (RTP16b) If the connection is INITIALIZED, CONNECTING or DISCONNECTED or the channel is ATTACHING or INITIALIZED, and ClientOptions#queueMessages has not been explicitly set to false, then all presence messages will be queued and delivered as soon as the connection state returns to CONNECTED and the channel is ATTACHED

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
@@ -1381,9 +1381,11 @@ namespace IO.Ably.Tests.Realtime
 
                     errCount.Should().Be(4);
 
-                    client.Close();                    
+                    client.Close();
                 }
-                
+
+                [Theory]
+                [ProtocolData]
                 [Trait("spec", "RTP16a")]
                 public async Task ConnectionStateCondition_WhenConnectionIsConnected_AllPresenceMessageArePublishedImmediately(Protocol protocol)
                 {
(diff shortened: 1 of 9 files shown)


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -1274,9 +1274,318 @@ public async Task WhenChannelBecomesAttached_ShouldSendQueuedMessagesAndInitiate
...
+                    var channel = client.Channels.Get("RTP16a".AddRandomSuffix()) as RealtimeChannel;
+
+                    await client.WaitForState(ConnectionState.Connected);
+                    await client.ConnectionManager.SetState(new ConnectionDisconnectedState(client.ConnectionManager, client.Logger));
+                    await client.WaitForState(ConnectionState.Disconnected);
+
+                    List<int> queueCounts = new List<int>();
+                    Presence.QueuedPresenceMessage[] presenceMessages = null;
+
+                    channel.Presence.Enter(client.Connection.State.ToString(), (b, info) =>{ });
```
paddybyers: If queueMessages is false, and the message is not queued, then it should be rejected at this stage, so there should be a check here for the relevant error

withakay: Issue added to catch this later https://github.com/ably/ably-dotnet/issues/298

PR #286, opened 2019-01-25
RTN15c2

withakay:
- (RTN15c2) CONNECTED ProtocolMessage with the same connectionId as the current client, and an error. In this case, the server is indicating that the resume succeeded but with a non-fatal error, all channels are still attached, and some backlog messages may be unavailable. The ErrorInfo received should be set as the reason in the CONNECTED event, and the Connection#errorReason should be set. The client should not change the state of attached channels, and immediately process any queued messages for that channel. Any channels that are not resumed in full may receive an ATTACHED ProtocolMessage with an error, see RTL12

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -410,6 +410,52 @@ namespace IO.Ably.Tests.Realtime
 
         [Theory]
         [ProtocolData]
+        [Trait("spec", "RTN15c1")]
+        public async Task ResumeRequest_ConnectedProtocolMessageWithSameConnectionId_WithNoError(Protocol protocol)
+        {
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -454,6 +454,88 @@ public async Task ResumeRequest_ConnectedProtocolMessageWithSameConnectionId_Wit
...
+            client.Connection.ErrorReason.Should().Be(stateChange.Reason);
+
+            // wait for the channel to attach
+            await attachedAwaiter.Task;
+
+            // it chanel state change event should have the injected error
+            channelStateChange.Error.Message.Should().Be("Faked channel error");
+
+            // queued messages should now have been sent
+            client.ConnectionManager.PendingMessages.Should().HaveCount(0);
```
paddybyers: As with RTN15c1, can we verify that the message does get published?

PR #289, opened 2019-01-25
RTN15c5

withakay:
- (RTN15c5) ERROR ProtocolMessage indicating a failure to authenticate as a result of a token error (see RTN15h). The transport will be closed by the server. The spec described in RTN15h must be followed for a connection being resumed with a token error


withakay:
There is a test for RTN15h1 here 
https://github.com/ably/ably-dotnet/pull/291

PR #306, opened 2019-03-11
JWT support

withakay:
Minor changes to support JWT and tests.


withakay:
I have added an issue to track the missing RTC8a4 tests here https://github.com/ably/ably-dotnet/issues/316

PR #309, opened 2019-03-16
TM3

withakay:
Message.FromEncoded and Message.FromEncodedArray

Adding this required quite a bit of refactoring as the message encoders were instanced with in the MessageHandler class and that was initialised with the AblyRealtime instance so I had to move something things around to make it work with the static methods. In addition to enabling the new methods to be added more easily I think the design is simpler now (and thus better).

The tests that I have added are based on those in the Ruby lib.

- (TM3) fromEncoded and fromEncodedArray are alternative constructors that take an (already deserialized) Message-like object (or array of such objects), and optionally a channelOptions, and return a Message (or array of such Messages) that's decoded and decrypted as specified in RSL6, using the cipher in the channelOptions if the message is encrypted, with any residual transforms (ones that the library cannot decode or decrypt) left in the encoding property per RSL6b. This is intended for users receiving messages other than from a REST or Realtime channel (for example, from a queue), to avoid them having to parse the encoding string themselves.

### DIFF
--- a/src/IO.Ably.Shared/MessageEncoders/Base64Encoder.cs
+++ b/src/IO.Ably.Shared/MessageEncoders/Base64Encoder.cs
@@ -5,8 +5,6 @@ namespace IO.Ably.MessageEncoders
 {
     internal class Base64Encoder : MessageEncoder
     {
-        private readonly Protocol _protocol;
-
         public override string EncodingName => "base64";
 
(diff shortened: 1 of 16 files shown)


withakay:
@paddybyers 

I don't think I have written test that use that explicitly, but I know there are existing tests (two for RSL6a1, for example) that do make use of it and other data from the ably-common repository.

PR #310, opened 2019-03-18
RTP17b test identification

withakay:
A quick PR to add a 'trait' to an existing test that covers RTP17b but was not labeled as such.

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -406,7 +406,7 @@ namespace IO.Ably.Realtime
                             break;
                         case PresenceAction.Leave:
                             broadcast &= Map.Remove(message);
-                            if (updateInternalPresence)
+                            if (updateInternalPresence && !message.IsSynthesized())
                             {
                                 InternalMap.Remove(message);
                             }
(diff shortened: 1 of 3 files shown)


SimonWoolf:
@withakay sorry, perhaps I'm missing something, but I'm not seeing either how that test covers RTP17b, or how the library implements it for LEAVEs? AFAICT from https://github.com/ably/ably-dotnet/blob/master/src/IO.Ably.Shared/Realtime/Presence.cs#L409-L412 the LEAVE event is currently being applied to the internal map conditioned only on the connectionId matching.

withakay:
@SimonWoolf 

Just so I am clear we are talking about the same thing, the spec states

- (RTP17b) The events that should be applied to the RTP17 presence map are: any ENTER, PRESENT or UPDATE event with a connectionId that matches the current client’s connectionId; any LEAVE event with a connectionId that matches the current client’s connectionId and is not a ‘synthesized leave’ (an event that has a connectionId which is not an initial substring of its id, per RTP2b1 )

The test demonstrates that the `InternalMap` (the object described in RTP17) only has entries for for the above events when the `connectionId` matches

```

**Comment by SimonWoolf** [comment]
> The test demonstrates that the InternalMap (the object described in RTP17) only has entries for for the above events when the connectionId matches

Yes, so it's only violating the `and is not a ‘synthesized leave’ (an event that has a connectionId which is not an initial substring of its id, per RTP2b1 )` part. (Which is the bit that's actually new to RTP17b, the rest is just restating RTP17)

**Comment by withakay** [comment]
@SimonWoolf 🤯 

doh, so I think this is a simple fix in Presence.cs

```
case PresenceAction.Leave:
    broadcast &= Map.Remove(message);
    if (updateInternalPresence && !message.IsSynthesized())
    {
        InternalMap.Remove(message);
    }
```

WDYT?

(And obviously I need to update the test to cover this change too...)

SimonWoolf:
:+1: lgtm

Issue #298, opened 2019-02-13
Presence (v1.1) if QueueMessages = false should raise an error


withakay:
see comment here

https://github.com/ably/ably-dotnet/pull/282/files/44a1d77cda9a15e4b4030250fe4eb32699c66a95#r255880655

PR #312, opened 2019-03-19
RTP16b fix

withakay:
Changes to address comments in issue #298 

> If queueMessages is false, and the message is not queued, then it should be rejected at this stage, so there should be a check here for the relevant error.

The message was not being rejected, so I have added that functionality and updated the test

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Presence.cs
+++ b/src/IO.Ably.Shared/Realtime/Presence.cs
@@ -349,7 +349,7 @@ namespace IO.Ably.Realtime
             {
                 msg.Callback?.Invoke(
                     false,
-                    new ErrorInfo("Unable enqueue message because Options.QueueMessages is set to False.", 50000, HttpStatusCode.InternalServerError));
+                    new ErrorInfo("Unable enqueue message because Options.QueueMessages is set to False.", _connection.Connection.ConnectionState.DefaultErrorInfo.Code, HttpStatusCode.ServiceUnavailable));
 
                 return false;
             }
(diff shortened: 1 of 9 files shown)


src/IO.Ably.Shared/Realtime/Presence.cs
```diff
@@ -347,6 +343,21 @@ internal void UpdatePresence(PresenceMessage msg, Action<bool, ErrorInfo> callba
             }
         }
 
+        private bool PendingPresenceEnqueue(QueuedPresenceMessage msg)
+        {
+            if (!_connection.Options.QueueMessages)
+            {
+                msg.Callback?.Invoke(
+                    false,
+                    new ErrorInfo("Unable enqueue message because Options.QueueMessages is set to False.", 50000, HttpStatusCode.InternalServerError));
```
paddybyers: This isn't a 500 error. It should be the default error for the current connection state.

src/IO.Ably.Shared/Transport/States/Connection/ConnectionClosedState.cs
```diff
@@ -6,6 +6,8 @@ namespace IO.Ably.Transport.States.Connection
 {
     internal class ConnectionClosedState : ConnectionStateBase
     {
+        public new ErrorInfo DefaultErrorInfo => ErrorInfo.ReasonClosed;
```
paddybyers: So where are these codes declared?

withakay: https://github.com/ably/ably-dotnet/blob/c4a35caaebcc61121b190e31cbac6f937b1f76eb/src/IO.Ably.Shared/Types/ErrorInfo.cs

PR #323, opened 2019-04-28
HistoryRequestParams fix

withakay:
When upgrading to 1.1 `HistoryRequestParams` was replaced with PaginatedRequestParams (as part of a process of reducing code duplication) which has created a breaking change for people moving from 0.8.

The fix presented here is to create new `HistoryRequestParams` class that inherits from `PaginatedRequestParams`.

### DIFF
--- a/src/IO.Ably.Shared/PaginatedRequestParams.cs
+++ b/src/IO.Ably.Shared/PaginatedRequestParams.cs
@@ -28,6 +28,7 @@ namespace IO.Ably
 
         public object Body { get; set; }
 
+
         public string Path { get; set; }
 
         /// <summary>
@@ -237,4 +238,13 @@ namespace IO.Ably
             }
         }
     }
+
+    /// <inheritdoc />
+    /// <summary>
+    ///     Data request query used for querying history.
+    ///     Functionally identical to <see cref="PaginatedRequestParams"/> and present for backwards compatibility with 0.8 release
+    /// </summary>
+    public class HistoryRequestParams
+        : PaginatedRequestParams
+    { }
 }
--- a/src/IO.Ably.Shared/PaginatedRequestParams.cs
+++ b/src/IO.Ably.Shared/PaginatedRequestParams.cs
@@ -28,7 +28,6 @@ namespace IO.Ably
 
         public object Body { get; set; }
 
-
         public string Path { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


mattheworiordan:
0.8 -> 1.1 is a breaking release though. Are we happy to continue to support this type in spite of it not being anywhere in the spec? I appreciate it’s nice of us to provide forwards support, but equally major releases do give us an opportunity to improve the API and clean things up. I wasn’t involved in the discussion to add this type, but are we sure this is what we want?

ndastur:
Just been pointed to this by Paddy. Agree clean up a good idea but 1.1 documents still refer to old class.

withakay:
@mattheworiordan 
> Are we happy to continue to support this type in spite of it not being anywhere in the spec?

My feeling is the default position should be 'no'. We should, I think, assess individual cases and decide if it is worth the effort (or is even feasible). In this particular case the fix was trivial and has no obvious side effects (other than having us all working on a Sunday!). When it is that easy to be backwards compatible I think it is worth doing.

mattheworiordan:
> When it is that easy to be backwards compatible I think it is worth doing.

Sure, that is perhaps worth it, but if we do that, shouldn't we:

* Show a deprecation notice
* Update the docs to use the new style

Either way, we need to think about whether it makes sense to support this if we didn't have existing users of an old API.  If the answer is no, then we should deprecate and perhaps support in 1.1, 1.2 etc and drop in 2.0.

withakay:
@mattheworiordan I have added an `[Obsolete]` attribute to the `HistoryRequestParams` class, this will give a compile time warning (Warning CS0618: 'HistoryRequestParams' is obsolete: 'HistoryRequestParams may be removed in future versions, please use PaginatedRequestParams instead.') and in Visual Studio there will be a squiggle and popup.

image

(Yup, no idea why VS needs to display the warning twice 🤦‍♂️)

mattheworiordan:
Looks good. 

Are there any other param types that need updating too, such as `StatsRequestParams`?

withakay:
`StatsRequestParams` still exists and inherits from `PaginatedReqestParams` (Previously it inherited from `HistoryRequestParams`). 

https://github.com/ably/ably-dotnet/blob/0941fb16bb5245e2914dbc3a09bff81ca174c9a9/src/IO.Ably.Shared/DataRequestQuery.cs#L16

mattheworiordan:
> StatsRequestParams still exists and inherits from PaginatedReqestParams

Ah, OK, and I see it extends `PaginatedReqestParams` so fine to leave as is.

Issue #325, opened 2019-05-02
JsonReader error on Xamarin


mattheworiordan:
Reported error from customer:

```
Unexpected error :Error reading JObject from JsonReader. Current JsonReader item is not an object: String. Path '', line 1, position 278.
```

I will ask for more details, but understand this was on iOS with Xamarin Forms, .NET lib 1.1.1

santihcc:
We are trying to connect to ably by using a Token.
The token is issued by our own server at a given auth url.
The code to initialise and connect is the following:

            try
            {
                var options = new ClientOptions
                {
                    AuthHeaders = new Dictionary<string, string> { { "Authorization", bearer } },
                    AuthUrl = new Uri("https://<auth-url>/api/ably"),
                };

                _ablyRealtime = new AblyRealtime(options);

                _ablyRealtime.Connection.On(args =>
                {
                    var currentState = args.Current; //Current state the connection transitioned to
                    var previousState = args.Previous; // Previous state
                    AblyMessages.Insert(0, $"Connection State: {currentState}");

                    var error = args.Reason; // If the connection errored the Reason object will be populated.
                    if (error != null)
                        AblyMessages.Insert(0, $"Connection Error: {error.Message}");
                });

                _ablyRealtime.Connection.ConnectionStateChanged += (sender, e) =>
                {
                };
            }
            catch (Exception e)
            {

            }

The AblyRealtime is initialized. Then the connection object changes state to connecting.
Running on both iOS or Andorid is OK.

Then the connection object should try to change state to connected but
On iOS we've got the following error:
`Unexpected error :Error reading JObject from JsonReader. Current JsonReader item is not an object: String. Path '', line 1, position 278.`
On Android the application crashes straight away. No error is caught in the try-catch wrapping up the code either.

The connection with our auth server looks to be fine as well as data is returned back. We also tested the server by calling it from a different application and the data is returned as expected.

mattheworiordan:
@santihcc I believe we may have a fix for this. We'll be in touch soon.

withakay:
This issue was down to an incorrectly formatted TokenRequest string be return from an ASP.Net MVC Controller

```

withakay:
@santihcc Could you let us know if this fix worked for you? Thanks

withakay:
Assuming fixed as no response from OP. Please reopen if issue is still present.

PR #328, opened 2019-05-03
Add InnerException to ErrorInfo proposal

withakay:
This is a proposal for an improvement to ErrorInfo.

I have added an InnerException property to the ErrorInfo class for situations when an ErrorInfo is created from an Exception this can be assigned, hopefully making it easier to track down issues.

I have  have also updated AuthURL error handling to make use of the InnerException property and added a test to demonstrate it, this in in response to issue #325

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -253,6 +253,7 @@ namespace IO.Ably
             }
             else if (authOptions.AuthUrl.IsNotEmpty())
             {
+                var responseText = String.Empty;
                 try
                 {
                     var response = await CallAuthUrl(authOptions, tokenParams);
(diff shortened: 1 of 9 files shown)


withakay:
`ErrorInfo` could inherit from `Exception` and then we down cast the `innerException` to an `ErrorInfo` and assign it to `cause` (well, `Cause` in c#), technically I think that would work (although iirc there might be some edge cases around serialisation I would have to double check). 

Semantically though, it *might* be a little odd as `ErrorInfo` is now an `Exception` sub class that never gets thrown and we already have `AblyException` for actually throwing Exceptions. 🤷‍♂

mattheworiordan:
> Semantically though, it might be a little odd as ErrorInfo is now an Exception sub class that never gets thrown and we already have AblyException for actually throwing Exceptions. 

Well not sure what's best to be honest.  Feels like you need to work out what is best for C# and our SDK, and something we're willing to sticky by moving forwards.

withakay:
@mattheworiordan 

I think `ErrorInfo` should not inherit from `Exception`, we should a `Cause` property of type `ErrorInfo` per the spec and where useful add another property to hold a reference to a 'source' `Exception`. I am not sure what the most appropriate name for that property would be though...

withakay:
@mattheworiordan 
> what needs to change to a) ensure the cause and inner exception are used

I think it is a case of manually updating the code in the places where it is relevant. I would think we need to decide if that is a proactive or reactive process.

> b) is this definitely visible in logs sent to STDOUT? Do we, for example, need to define a serializer that ensures when this ErrorInfo is shown it includes that info?

Good catch. I have pushed a commit to address this, see https://github.com/ably/ably-dotnet/pull/328/commits/3255a1bf25cd1610592f4883231a0149412d7214

Issue #332, opened 2019-05-13
Resuming after Windows has been asleep (.Net Framework)


withakay:
A follow on to #331 and related to an issue a user is seeing.

In my testing when a program is running that is connected to a channel with presence entered, when windows is put to sleep for several minutes and then resumed the ably connection is not resumed, there is nothing logged the realtime library simply does nothing. The app is still running in the console and can respond to input 

There is no xplat way to detect sleep/resume or start/stop, but for .NET Framework (Windows) there is `SystemEvents.PowerModeChanged`. Wiring up the Suspend and Resume events to then tell ably the network off and on line causes the program to resume as one might expect.

```
switch (eventArgs.Mode)
{
    case PowerModes.Suspend:
        Connection.NotifyOperatingSystemNetworkState(NetworkState.Offline);
        break;
    case PowerModes.Resume:
        {
            if (NetworkInterface.GetIsNetworkAvailable())
            {
                Connection.NotifyOperatingSystemNetworkState(NetworkState.Online);
            }

            break;
        }
}
```

akaviny:
This issue is a little more involved, unfortunately. 

On a Windows Service, `SystemEvents.PowerModeChanged` won't be raised, unless a message pump is available (hidden Window, or manually setting the **Allow service to interact with desktop** on the service log on tab).

The recommended way for services is via `ServiceBase.OnPowerEvent`. 

So, I'd suggest allowing clients to hook up their own handler, and exposing a way of notifying of network changes.

withakay:
> So, I'd suggest allowing clients to hook up their own handler, and exposing a way of notifying of network changes.

This may be an option, but we are aim to have the library 'do the right thing' without needing to be specifically aware of the power status off the OS

akaviny:
An option would be to expose a property on the connection, e.g., an interface, `IPowerEventSource`, with a default implementation, `DesktopPowerEventSource`, using the `SystemEvents.PowerModeChanged`.

This can then be replaced by a client running in a different environment, e.g., `WindowsServicePowerEventSource`. 

It would also give you the benefit of switching in a unit testable implementation.

withakay:
I agree with your point, but the basic premise that the library needs to be aware of the power status I am not sure is necessarily correct, as far as the library is concerned if a system is in a paused power state then that can just be seen as a disconnection.

tomczoink:
[...]

Issue #331, opened 2019-05-13
Detecting network availability changes


withakay:
Currently the dotnet library does not detect network changes in places where it could. There is an internal method `NotifyOperatingSystemNetworkState` on the `Connection` class that is only referenced by Unit Tests.

This method should be wired up to network change events where possible. 

There is a static event for this `System.Net.NetworkInformation.NetworkChange.NetworkAvailabilityChanged` that is supported in all our targets except UWP and .NET Standard 1.4. UWP there is a similar event that can be hooked into `NetworkInformation.NetworkStatusChanged`, for .NET Standard 1.4 there is no option built in. Since Xamarin and .Net Core 2 both support .NET Standard 2.0  the lack of an option for .NET Standard is a minor issue that I think we can ignore IMO.

For each platform version we target we have project that imports the shared code (IO.Ably.Shared) and additionally contains a `Platform.cs`, I propose that we add a static constructor to this class to wire up the Network Change events

```
static Platform()
{
    NetworkChange.NetworkAvailabilityChanged += (sender, eventArgs) =>
        Connection.NotifyOperatingSystemNetworkState(eventArgs.IsAvailable ? NetworkState.Online : NetworkState.Offline);
}
```

marto83:
Closing as it has been implemented.

PR #333, opened 2019-05-14
Network changed events

withakay:
This PR is a solution to the issues raised in #331 and #332.

Regarding #332, I have added sleep/resume handling as if that were a network outage to the .NET Framework (see src/IO.Ably.NETFramework/Platform.cs below).

Instinctively I feel like this shouldn't be required for the client lib to resume from a system sleep but without it when a (Windows) system comes back from sleep the connection doesn't recover. So, perhaps this is a bit of a hack but it makes the library work as expected, so the pragmatist in me says that is OK.

It is possible that similar solutions could be added for iOS and Android, but right now I am not sure what the best options are and no one is complaining of having issues with Xamarin in that regard and testing those solutions is harder, I prefer to push that to another work item.

I should probably have split this up into 2 PRs, but the changes are quite small so hopefully whoever reviews this can overlook that slight faux pas!

### DIFF
--- a/src/IO.Ably.Android/Platform.cs
+++ b/src/IO.Ably.Android/Platform.cs
@@ -1,4 +1,6 @@
 ﻿using IO.Ably.Transport;
+using System.Net.NetworkInformation;
+using IO.Ably.Realtime;
 
 namespace IO.Ably
 {
@@ -6,5 +8,11 @@ namespace IO.Ably
     {
         public string PlatformId => "xamarin-android";
         public ITransportFactory TransportFactory => null;
+
+        static Platform()
+        {
+            NetworkChange.NetworkAvailabilityChanged += (sender, eventArgs) =>
+                Connection.NotifyOperatingSystemNetworkState(eventArgs.IsAvailable ? NetworkState.Online : NetworkState.Offline);
+        }
     }
 }
--- a/src/IO.Ably.NETFramework/Platform.cs
+++ b/src/IO.Ably.NETFramework/Platform.cs
@@ -1,4 +1,8 @@
 ﻿using IO.Ably.Transport;
+using System;
+using System.Net.NetworkInformation;
+using IO.Ably.Realtime;
+using Microsoft.Win32;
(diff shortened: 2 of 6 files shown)


withakay:
I do have a couple of concerns with this PR. 

Firstly, this solution means we get different behaviours on different platforms and even different behaviours on the same platforms depending on how the library is consumed, which is not great. 

Secondly, there is no solution for .NET Standard and this is now the preferred way to consume the library when using Xamarin and .NET Core. At some point in the not too distant future the Android and iOS versions will become legacy.

It might be better to expose hooks and have users wire up the events for each platform themselves

withakay:
@paddybyers 
> How much testing have you been able to do with various network conditions?

I have tested in simulators and a Windows VM, so... not enough.

Is there a provision in the spec for enabling experimental features, perhaps via ClientOptions, because I am confident the solution here does resolve a customer issue on Windows but less confident about the impact on other platforms (specifically Xamarin), so allowing them to enable that might give us some extra breathing room to come up with a more complete solution, WDYT?

paddybyers:
> Is there a provision in the spec for enabling experimental features, perhaps via ClientOptions, because I am confident the solution here does resolve a customer issue on Windows but less confident about the impact on other platforms (specifically Xamarin), so allowing them to enable that might give us some extra breathing room to come up with a more complete solution, WDYT?

I'd be open to enabling a custom client option if we're really not able to establish the stabilty/correctness of this ourselves. I can't help thinking that we're going to need to bite the bullet sooner or later and do the implementation/testing for platform powerstate and network events on the other platforms.

@mattheworiordan wdyt?

mattheworiordan:
> I'd be open to enabling a custom client option if we're really not able to establish the stabilty/correctness of this ourselves. 

Sure, nothing wrong with that, so long as it's undocumented from official docs and has an indicator that it's an experiemental API in the IDE intellisense stuff

withakay:
merging for planned 1.17-beta release

Issue #334, opened 2019-05-15
NRE in IO.Ably.Transport.MsWebSocketTransport.ConnectAndStartListening and SemaphoreSlim.Wait


softsan:
We are getting lots of crash in our hockey app with following exception (on android) 
we are using ably 1.1.2

**Crashes per Day**
2290
Last 24 hours
6548
Last 7 days
6548
Last 30 days

```
MsWebSocketTransport.ConnectAndStartListening() - 2290 times in 24 hours
android.runtime.JavaProxyThrowable: System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Object reference not set to an instance of an object) ---> System.NullReferenceException: Object reference not set to an instance of an object
2a7c3906b6e14e42ae7028a3b0bcd816

Xamarin caused by: android.runtime.JavaProxyThrowable: System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Object reference not set to an instance of an object) ---> System.NullReferenceException: Object reference not set to an instance of an object
  at IO.Ably.Transport.MsWebSocketTransport.ConnectAndStartListening () [0x00189] in <2a7c3906b6e14e42ae7028a3b0bcd816>:0 
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.NullReferenceException: Object reference not set to an instance of an object
  at IO.Ably.Transport.MsWebSocketTransport.ConnectAndStartListening () [0x00189] in <2a7c3906b6e14e42ae7028a3b0bcd816>:0 <---
```

and this:
**Crashes per Day**
1251
Last 24 hours
3180
Last 7 days
3180
Last 30 days
 

```
SemaphoreSlim.Wait(System.Int32 millisecondsTimeout, System.Threading.CancellationToken cancellationToken)
android.runtime.JavaProxyThrowable: System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Value cannot be null.
cfcec5514a31434aae323caf1152dcac

[...]

PR #335, opened 2019-05-17
Fix for Issue 334, incorrectly disposing of a blocking collection

withakay:
This PR resolves a regression introduced with v1.1.1 that is covered by #334.

In implementing the Dispose pattern more thoroughly to solve an issue with a thread leak I had called `Dispose` on a `BlockingCollection` which is not thread safe. The `BlockingCollection`, when access after being disposed with throw a `ObjectDisposedException`, because the `BlockingCollection` is being access from a worker thread this was unhandled and could cause the application to crash.

The recommended technique with a `BlockingCollection` is to call `CompleteAdding()` to prevent more items being added to the collection and make use of a `TokenCancellationSource` passing it to `GetConsumingEnumerable()` to allow cancellation.

`Cancel`ing the `TokenCancellationSource` will cause an `OperationCanceledException` to be thrown which must be handled to prevent a potential crash, but no further action is required (although I have added a debug log message).

Additionally I have added some error logging around adding items to the `BlockingCollection` as a precaution

### DIFF
--- a/src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
+++ b/src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
@@ -66,20 +66,37 @@ namespace IO.Ably.Transport
 
         private void StartSenderQueueConsumer()
         {
-            var queueEnumerable = _sendQueue.GetConsumingEnumerable();
             Task.Run(
                 async () =>
-                {
-                    foreach (var tuple in queueEnumerable)
                     {
-                        await Send(tuple.Item1, tuple.Item2, _tokenSource.Token);
-                    }
-                }, _tokenSource.Token).ConfigureAwait(false);
+                        try
+                        {
+                            if (_sendQueue != null)
+                            {
+                                foreach (var tuple in _sendQueue?.GetConsumingEnumerable(_tokenSource.Token))
+                                {
+                                    await Send(tuple.Item1, tuple.Item2, _tokenSource.Token);
+                                }
+                            }
+                        }
+                        catch (OperationCanceledException e)
+                        {
+                            if (Logger.IsDebug)
+                            {
+                                Logger.Debug(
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
```diff
@@ -66,20 +66,37 @@ public async Task StartConnectionAsync()
...
-                    foreach (var tuple in queueEnumerable)
                     {
-                        await Send(tuple.Item1, tuple.Item2, _tokenSource.Token);
-                    }
-                }, _tokenSource.Token).ConfigureAwait(false);
+                        try
+                        {
+                            if (_sendQueue != null)
+                            {
+                                foreach (var tuple in _sendQueue?.GetConsumingEnumerable(_tokenSource.Token))
```
NVentimiglia: Is it possible the collection changes while the async operation takes place?

caseyniemann:
@withakay we are still having issues with 1.1.3

NVentimiglia:
Looks like this introduced new errors, failure to connect.

src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
withakay: IMO BlockingCollection is the correct construct to be using here, it is specifically designed for these types of situations (the Producer/Consumer pattern). It is designed so that it can be added to in a thread safe manner (via TryAdd) and when it has items `GetConsumingEnumerable` yields them, when there are no items it blocks effectively waiting for an item to be added.

withakay:
@caseyniemann & @NVentimiglia can you elaborate? (preferably on the issue #334 as this PR has now been merged and released).

caseyniemann:
@withakay @NVentimiglia @softsan see comments https://github.com/ably/ably-dotnet/issues/334#issuecomment-493567599

Issue #314, opened 2019-04-05
Investigate report of PublishAsync hanging/timing out


withakay:
We have had a Report of PublishAsync failing to send and timing out when using Xamarin.iOS

karen-hikesupport:
we have same issue publishAsync Timeout failed usiing .net Core, below the log error 

{"IsSuccess":false,"Error":{"code":0,"statusCode":null,"message":"PublishAsync timeout expired. Message was not confirmed by the server","href":null,"IsUnAuthorizedError":false,"IsForbiddenError":false,"IsTokenError":false},"IsFailure":true}

also we have logged loggerSink data : 

10:41:47.462 xx Changing state from Initialized => Connecting. SkipAttach = False.
10:42:06.630 ConnectionManager thread created
10:42:19.984 xx Connecting: BeforeTransition
10:42:28.456 xx Connecting: BeforeTransition end
10:42:39.439 Connection notifying subscribers for state change `Connecting`
10:44:16.203 Recieving message with type: Text
10:44:28.404 Websocket message received. Raw: {"action":0}
10:44:37.037 Proccessing queued message: { action=Heartbeat }
10:44:48.025 Recieving message with type: Text
10:44:56.842 Websocket message received. Raw: {"action":0}
10:45:05.978 Proccessing queued message: { action=Heartbeat }
10:45:37.744 Recieving message with type: Close
10:45:59.168 Transport State: Closing. Error is empty.
10:46:14.811 Closing websocket. Close status: NormalClosure, Description:

withakay:
Currently reported on Xamarin iOS and ASP.Net MVC Core 2

withakay:
@karen-hikesupport 
Hi, I have only be able to get `PublishAsync` to hang (actually a deadlock) when it is used in unexpected ways, either by failing to `await` or by calling `Result` (`PublishAsync().Result`).
I have a update to mitigate this, but it would be really useful to know exactly how you are calling `PublishAsync` so I can make sure I have not missed anything.

withakay:
I am closing this due to inactivity, please re-open if this is still an issue

PR #341, opened 2019-05-19
PublishAsync improvement

withakay:
Related to #314 

Whilst I have not been able to repeat this issue, after some investigation it was apparent that `PublishAsync` had the potential to deadlock due to a returning a `Task.Result`.

We should instead be `await`ing the `Task`, additionally as we are writing library code we can make use of `ConfigureAwait(false)` as we do not need the calling context.  

For more details of why this is so see https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

### DIFF
--- a/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
+++ b/src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
@@ -359,7 +359,7 @@ namespace IO.Ably.Realtime
                 tw.SetException(ex);
             }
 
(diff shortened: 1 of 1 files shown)


withakay:
I have now reproduced a situation where the old code can deadlock, when either the call to `PublishAsync` is not `await`ed or when `.Result` is called on the returned `Task`, causing to block. 
Both of these usages are valid, although probably not what the user that reported this error intended.

Eitherway, the changes in the PR mitigate that problem with no measurable side effects, so I am confident we can release this.

PR #343, opened 2019-05-22
Transient Publishing

withakay:
- (RTL6c1) If the connection is CONNECTED and the channel is INITIALIZED, ATTACHED, DETACHED, ATTACHING, or DETACHING then the messages are published immediately
- (RTL6c2) If the connection is INITIALIZED, CONNECTING or DISCONNECTED, and ClientOptions#queueMessages has not been explicitly set to false, then the message will be queued and delivered as soon as the connection is CONNECTED and the channel is in a state in which publishing is permitted per RTL6c1
- (RTL6c4) If the connection is SUSPENDED, CLOSING, CLOSED, or FAILED, or the channel is SUSPENDED or FAILED, the operation will result in an error

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Connection.cs
+++ b/src/IO.Ably.Shared/Realtime/Connection.cs
@@ -70,6 +70,13 @@ namespace IO.Ably.Realtime
 
         internal Func<DateTimeOffset> Now { get; set; }
 
+        internal bool CanPublishMessages =>
+            State == Realtime.ConnectionState.Connected
+            || ((State == Realtime.ConnectionState.Initialized
+                 || State == Realtime.ConnectionState.Connecting
+                 || State == Realtime.ConnectionState.Disconnected)
+                && RealtimeClient.Options.QueueMessages);
+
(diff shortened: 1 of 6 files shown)


withakay:
@paddybyers @SimonWoolf 
This is a work in progress, could you please sanity check my work so far?

SimonWoolf:
looks fine to me. You're not missing anything, transient publishing pretty is just as simple as 'remove the state==attached check'.

You can probably also get rid of `channel.QueuedMessages` entirely (unless you're also using that queue for presence, didn't check)

withakay:
@SimonWoolf 

>You can probably also get rid of channel.QueuedMessages entirely (unless you're also using that queue for presence, didn't check)

I actually did that on my first pass but it broke some tests, so I reinstated it.
I have now realised that the failing tests cover RTL11 which have not been updated for 1.1 anyway, so I will have another look at removing that queue.

withakay:
@paddybyers 

I have added the test you mentioned and the additional test for transient publishing found in the Java lib and fixed up some test regression, mainly for RTL11.

PR #350, opened 2019-06-15
Presence not re-entering after connect disconnect fix

withakay:
This a fix for an issue where by presence was not re-entered after a connection became disconnected (but remained in the local presence map), which is related to issue #332.

This problem was caused by some changes that were added to support the now defunct RTP3 being incorrectly implemented such that upon every connection state change a `SYNC` message was sent to the Realtime service, this had the effect of putting the Realtime service into an unexpected state (An issue has been raised internally about this by @SimonWoolf).

As RTP3 is no longer required the solution was to remove the changes relating to that (no public facing methods are affected, so this is not a breaking change) and then fix some regressions, most notably for RTP19, which seemed to just be an incorrect test. I have refactored and updated the RTP19 test based on the Java implementation and it now passes.  

I have added a test for the bug, derived from the console application I made for debugging, possible there is a spec item that this test covers, but I have not identified it as yet (any help appreciated!).

Additionally I have made some minor improvements, such as logging the content of outbound messages when the `LogLevel==Debug`

### DIFF
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -602,7 +602,7 @@ namespace IO.Ably.Transport
 
                         SetState(new ConnectionDisconnectedState(this, new ErrorInfo("Connection closed due to Operating system network going offline", 80017), Logger)
                         {
-                            RetryInstantly = true
+                            RetryInstantly = false
                         });
                     }
 
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -600,9 +600,10 @@ namespace IO.Ably.Transport
                             Logger.Debug("Network state is Offline. Moving to disconnected.");
                         }
 
-                        SetState(new ConnectionDisconnectedState(this, new ErrorInfo("Connection closed due to Operating system network going offline", 80017), Logger)
+                        // RTA20a
+                        SetState(new ConnectionDisconnectedState(this, new ErrorInfo("Connection disconnected due to Operating system network going offline", 80017), Logger)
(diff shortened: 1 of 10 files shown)


src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -600,7 +600,8 @@ public void HandleNetworkStateChange(NetworkState state)
                             Logger.Debug("Network state is Offline. Moving to disconnected.");
                         }
 
-                        SetState(new ConnectionDisconnectedState(this, new ErrorInfo("Connection closed due to Operating system network going offline", 80017), Logger)
+                        // RTA20a
```
paddybyers: Typo: RTN20a

src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -326,6 +327,12 @@ public void Send(ProtocolMessage message, Action<bool, ErrorInfo> callback = nul
                 Logger.Debug($"Current state: {Connection.State}. Sending message: {message}");
             }
 
+            if (message.ConnectionId.IsNotEmpty())
+            {
+                Logger.Warning("Setting ConnectionId to null. ConnectionId should never be included in an outbound message on a realtime connection, it’s always implicit");
```
SimonWoolf: Feels a bit odd to have this as a runtime check. If this warning is ever triggered that means there's a bug in the library, which the user can't do anything about. Do you anticipate that this will ever be called?

withakay: > Do you anticipate that this will ever be called?

I would hope not, but that said, if it had been there then I wouldn't have been so confused for so long. 

I don't have a strong feeling about this, I could remove it if that is preferred?

Issue #330, opened 2019-05-09
Using a clientId should no longer be forcing token auth in the 1.1 spec


SimonWoolf:
cf ably/docs#345

PR #351, opened 2019-06-15
Using a clientId should no longer be forcing token auth

withakay:
Fix for issue #330


paddybyers:
1) I don't understand how this fixes the issue - can you explain pls?
2) I think there should be a test.

SimonWoolf:
Like paddy I'm a bit confused - the logic to use token auth when a clientId is set appears to be https://github.com/ably/ably-dotnet/blob/8f849f8c9754c44d427e71bfffe71566e0ba9b28/src/IO.Ably.Shared/ClientOptions.cs#L222 , which isn't touched by this pr?

withakay:
Killing this PR and starting again

Issue #352, opened 2019-06-30
history.NextAsync() throws 40000 exception


ndastur:
Using ably 1.1.8 SDK

Steps to reproduce

1) Publish 100+ messages on a channel
2) Use this code to get a history
```
var historyParams = new PaginatedRequestParams()
{
   Direction = QueryDirection.Forwards,
   Start = Settings.MessageGatewayLastMessageInTimestamp
                                        .FromUnixTimeInMilliseconds()           // get the DateTimeOffset
                                        .AddMinutes(-2)                        // go back sometime to capture all
};

   PaginatedResult<Message> history;
   var specific_incoming_chat_channel = RealtimeObj.Channels.Get(specificChannelName);
   history = await specific_incoming_chat_channel.HistoryAsync(historyParams);

   // Loop over the history
   if (history == null)
       return null;

   do
                    {
                        Debug.WriteLine($"{Tag} got history object with {history.Items.Count} items");

                        if (cancellationToken.IsCancellationRequested)
                            cancellationToken.ThrowIfCancellationRequested();

                        if (!history.HasNext)
                            break;

                        history = await history.NextAsync();
                    }
   while (true);
```

This will throw an exception at ```history = await history.NextAsync();```
Exception: [ErrorInfo Reason: Invalid mode specification. (See https://help.ably.io/error/40000 for help.); Code: 40000; StatusCode: 400 (BadRequest); Href: https://help.ably.io/error/40000]

paddybyers:
This is the request we're receiving:

```
"GET /channels/chat%3AR%3Ahistory_test_1ac4e3722459c/messages?start=946684680000&end=1561853924942&direction=forwards&limit=100&cont=true&format=json&firstStart=946684680000&fromDate=946684680000&mode=all%3E%3B+rel%3D%22first%22%2C+%3C.%2Fmessages"
```

ndastur:
Glad the problem is identified. At the moment this blocks our production app getting me than 100 messages from the history.

paddybyers:
> At the moment this blocks our production app getting me than 100 messages from the history.

If it helps, you can specify a `limit` param to be any value up to 1000 and get that number of messages in the first page.

ndastur:
Okay. Thanks

withakay:
This will be fixed in the 1.1.20 release due today

ndastur:
Thanks @withakay . In our test app the fix works for Android and iOS

PR #353, opened 2019-07-03
split link header values on comma to handle platform implementation d…

withakay:
On the Xamarin (Mono) runtime headers with the same name have their values concatenated to appear as one header with the values as a comma separated string.

```
<./messages?start=1561532338629&end=1562137138951&limit=100&direction=forwards&cont=true&format=json&firstStart=1561532338629&fromDate=1561532338629&mode=all>; rel="first", 
<./messages?start=1561532338629&end=1562137138951&limit=100&direction=forwards&cont=true&format=json&firstStart=1561532338629&fromDate=1561532338629&mode=all>; rel="current", 
<./messages?fromSerial=e7dRsa5kQAZOfO43314764%3A100&mode=live&limit=100&direction=forwards&cont=true&format=json&firstStart=1561532338629&end=1562137138951>; rel="next"
```

(line breaks added for clarity)

This is valid per rfc2616 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), but is inconsistent across various .NET runtimes, (on .NET Core there would be three headers, each with one value).

The applied fix is to split on a comma and loop over the resulting array. If there is no comma an array with one element is created.

### DIFF
--- a/src/IO.Ably.Shared/PaginatedRequestParams.cs
+++ b/src/IO.Ably.Shared/PaginatedRequestParams.cs
@@ -202,16 +202,24 @@ namespace IO.Ably
         internal static PaginatedRequestParams GetLinkQuery(HttpHeaders headers, string link)
         {
             const string linkPattern = "\\s*<(.*)>;\\s*rel=\"(.*)\"";
+
+            // There can be multiple headers for Link (first, next, current)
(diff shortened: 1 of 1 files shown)


withakay:
This addresses Issue #352

PR #367, opened 2019-08-19
RunInBackground bug

withakay:
An error was made when using `TaskUtils.RunInBackground`.
The original call was made like this
```
TaskUtils.RunInBackground(async () => await Context.RetryAuthentication(updateState: false), e => Logger.Warning(e.Message));
```

the method signature was of `RunInBackground` was
```
RunInBackground(Action action, Action<Exception> handler = null)
```

This is wrong and internally this created a Task 'wrapped' in a Task which meant the continuation (the `hander`) was applied to the incorrect (outer) Task allowing the exception on the inner Task to go unhandled.

The simplest solution I could come up with was to create and overload
```
RunInBackground(Task task, Action<Exception> handler = null)
```

and update the way this overload is invoked, thus

```
TaskUtils.RunInBackground(Context.RetryAuthentication(updateState: false), e => Logger.Warning(e.Message));
```

It should be noted that this new overload could still legally be called in the old manner and it would fail in exactly the same way, but the new style invocation would not be legal without the new overload.

### DIFF
--- a/src/IO.Ably.Tests.Shared/TaskUtilsTests.cs
+++ b/src/IO.Ably.Tests.Shared/TaskUtilsTests.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using IO.Ably.Tests.Infrastructure;
+using Xunit;
+
+namespace IO.Ably.Tests
+{
+    public class TaskUtilsTests
+    {
+        class MockBagroundService
+        {
+            public static async Task FailingTask()
+            {
+                await Task.Delay(100);
+                throw new Exception();
+            }
+        }
+
+        [Fact]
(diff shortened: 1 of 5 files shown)


src/IO.Ably.Tests.Shared/TaskUtilsTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                tsc.SetCompleted();
+            });
+
+            var result = await tsc.Task;
+            result.Should().BeTrue();
+        }
+
+        [Fact]
+        public async Task FailingTaskWRappingAsyncWillNotHaveExceptionHandled()
```
paddybyers: typo? `WRapping` -> `Wrapping`

withakay: typo fixed by https://github.com/ably/ably-dotnet/pull/367/commits/21173dbf3f632ad6dc12fefaf21cbac5041b3d58

PR #370, opened 2019-09-19
Add fromEncoded and fromEncodedArray that accept json string

marto83:
This updates the Message class to include overloads that are needed
when we have encoded messages in json format.

It also adds a check on the CypherEncode to return better error message
when the type of Data is not byte[]


src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
```diff
@@ -33,6 +33,9 @@ public override Result Decode(IMessage payload, ChannelOptions options)
             var cipher = Crypto.GetCipher(options.CipherParams);
             try
             {
+                if((payload.Data is byte[]) == false)
+                    return Result.Fail("Expected data to be byte[] but received " + payload.Data.GetType());
+
```
withakay: stylecop should not be allowing this style of `if`, there is a rule to always require curly braces.

src/IO.Ably.Tests.Shared/Rest/TM3Spec.cs
```diff
@@ -0,0 +1,183 @@
...
+                fromEncoded[0].Encoding.Should().BeNullOrEmpty();
+
+                fromEncoded[1].Name.Should().Be("name2");
+                fromEncoded[1].Data.Should().Be("some-data2");
+                fromEncoded[1].Encoding.Should().BeNullOrEmpty();
+            }
+        }
+
+    }
+}
```
withakay: There is a rule that there should always be a carriage return at the end of a file (sorry to nit pick..)

withakay:
Looks good to me, aside from the minor formatting issues.

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
marto83: I've updated it. For some reason the full build didn't run so Stylecop didn't pick it up.

src/IO.Ably.Tests.Shared/Rest/TM3Spec.cs
marto83: Updated 🛩

PR #372, opened 2019-09-26
Poc/channels

marto83:
This is a large pull request and I'll be happy to take any reviewers through it. 

Summary of teh changes

Created a RealtimeWorkflow class that makes sure state transitions and received messages are executed in order 
to ensure we get less threading issues. 
To order the actions - I've created a number of commands which are the units of executions. Examples of commands are:
SetConnectedStateCommand, SetFailedState, RetryAuthCommand. 

The above changes meant that I can simplify a lot of the locks inside the Connection Manager. 
There is no longer a sync method for running tests. We don't need to run things in the background and actions can be just queued.
I've removed the internal Statechange event handlers so everything goes through the eventEmitter class. This is to simplify the different ways notifications are sent internally.

I've also added the beginning of a state class to centralise the state of the library. Currently the state is spread among the different classes and the goal here is to make 
diagnosing and debugging the library much easier. It will be great if in the future we can get to a point where we can dump the whole state of the library.

Streamlined onAuthUpdated and also started to simplify the we do the state transitions. Now the logic is triggered inside the workflow class where we can guarantee a 
single thread execution means we can greatly simplify the logic.


src/IO.Ably.Shared/AblyRealtime.cs
```diff
@@ -1,19 +1,24 @@
...
+    public class AblyRealtime : IRealtimeClient, IDisposable
     {
         internal ILogger Logger { get; private set; }
 
+        internal RealtimeWorkflow Workflow { get; private set; }
+
         private SynchronizationContext _synchronizationContext;
 
+        internal volatile bool Disposed = false;
+
```
codemerx: Shouldn't the setter for this be private? Maybe even the getter, too? `ObjectDisposedException` is pretty standard thing to expect if using disposed object I guess

src/IO.Ably.Shared/Realtime/Workflows/ChannelCommands.cs
```diff
@@ -0,0 +1,36 @@
+namespace IO.Ably.Realtime.Workflow
+{
+    internal class ChannelCommand : RealtimeCommand
+    {
+        public string ChannelName { get; set; }
+
+        public RealtimeCommand Command { get; set; }
```
codemerx: It looks like the setters here should not be public by design. Does it make sense for these to be set externally?

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
```diff
@@ -0,0 +1,389 @@
...
+
+
+    internal class ListCommand : RealtimeCommand
+    {
+        public IEnumerable<RealtimeCommand> Commands { get; }
+
+        public ListCommand(IEnumerable<RealtimeCommand> commands)
+        {
+            Commands = commands.ToList();
+        }
```
codemerx: Why `ToList` when assinging `IEnumerable<RealtimeCommand>` to `IEnumerable<RealtimeCommand>`?

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
```diff
@@ -0,0 +1,389 @@
+using System;
```
codemerx: Why both factory method and public constructor for each command?

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
```diff
@@ -0,0 +1,400 @@
...
+
+        protected override string ExplainData()
+        {
+            var message = (Error != null) ? " Error: " + Error : string.Empty;
+            message += " ClearConnectionKey:" + ClearConnectionKey;
+            return message;
+        }
+    }
+
+
```
withakay: extra blank line here

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
```diff
@@ -0,0 +1,400 @@
...
+        private EmptyCommand() {}
+
+
+        protected override string ExplainData()
+        {
+            return string.Empty;
+        }
+    }
+
+
```
withakay: nit pick, but extra blank line here is inconsistent

src/IO.Ably.Shared/AblyRealtime.cs
marto83: Good point. This is not finished yet. I realise I need to move the object into a disposing state before setting it as disposed because it will need to tell the server to close the connection. I've got it on my list to finish

src/IO.Ably.Shared/Realtime/Workflows/ChannelCommands.cs
marto83: Nope they should never be set externally. I kept them like that to start with because I wasn't sure if I needed to serialise the commands. Now I know I don't need to so I'll make them all private

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
marto83: This is so it can copy the Enumerable at that point in time because it is potentially going to be evaluated on a different thread

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
marto83: I wasn't sure which way I liked in the beginning. I like the factory methods better so I'll make all the constructors private

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
marto83: thanks. There are about 230 warnings 😱 I need to fix. I better crack on 😅

src/IO.Ably.Shared/Realtime/Workflows/RealtimeCommands.cs
marto83: 👷 ⛏ 🔨 🛠

Issue #374, opened 2019-10-03
Token Expiry error when QueryTime is set to true


marto83:
The library is throwing an error about an expired token straight after receiving a new token from the server: 

Error: IO.Ably.AblyException: [ErrorInfo Reason: Token has expired: Token: ******, Expires: 03/10/2019 15:07:10 +00:00, Issued: 03/10/2019 14:07:10 +00:00, Capability: {"chat:*":["history","publish","subscribe"],"org:*":["history","publish","subscribe"]}, ClientId: MiQk3BuXP0 (See https://help.ably.io/error/40142); Code: 40142; StatusCode: 401 (Unauthorized); Href: https://help.ably.io/error/40142]

PR #376, opened 2019-10-03
Fix issue where incorrect time was used to check Token validity

marto83:
Even when the QueryTime time was on the library was using the default
time to verify the token returned.

Now we are using the correct time using the offset we calculated after
getting server time

Please ignore the TODO. I left it there so I can fix it in my poc branch. Otherwise there will be too many conflicts

Addresses Issue: #374


marto83:
Thanks Simon. I wanted to get this fixed and also I didn't want to do too much refactoring in master because it's going to cause a headache when I merge the poc/channels branch.

PR #386, opened 2019-11-18
Feature/delta support

marto83:
Initial changes for the delta infrastructure. 
I've added the new interfaces. Some of the names will need updating. 

We already had some infrastructure to support multiple encoders. This builds on top of that infrastracture and makes it work with supporting external codecs. 

I've made sure the commits represent logical changes so this is best reviewed by through the commits. Please note some of the early changes are superseded by later commits. For example initially we had to support an external dependency which was responsible for the delta decoding.


src/IO.Ably.Shared/MessageEncoder.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <param name="payload">the payload that needs to be encoded.</param>
+        /// <param name="encodingContext">the context needed for the operation.</param>
+        /// <returns><see cref="EncodingResult"/>.</returns>
+        EncodingResult Encode(object payload, EncodingDecodingContext encodingContext);
+    }
+
+    /// <summary>
+    /// Result from an encoding operation.
+    /// </summary>
+    public class EncodingResult
```
QuintinWillison: Properties should be read-only. This class should be immutable.

src/IO.Ably.Shared/Types/MessageExtras.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Extra message properties relating to the delta's functionality.
+    /// </summary>
+    public class DeltaExtras
+    {
+        /// <summary>
+        /// Format.
+        /// </summary>
+        public string Fromat { get; set; }
```
QuintinWillison: typo - should be "Format"

src/IO.Ably.Shared/MessageEncoders/VCDiffEncoder.cs
```diff
@@ -0,0 +1,60 @@
...
+                context.Encoding = RemoveCurrentEncodingPart(payload);
+                return Result.Ok(new ProcessedPayload()
+                {
+                    Data = result.AsByteArray(),
+                    Encoding = RemoveCurrentEncodingPart(payload),
+                });
+            }
+            catch (Exception e)
+            {
+                // TODO: Rethrow an error to indicate the vcdiff failed and things need to happen.
```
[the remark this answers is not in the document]
marto83: That's my next task. I've got a failing test and will add the extra recovery logic.

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -8,15 +9,19 @@
 using IO.Ably.Types;
 using IO.Ably.Utils;
 using Newtonsoft.Json.Linq;
+using IO.Ably.MessageEncoders;
 
 namespace IO.Ably.Realtime
 {
+    [DebuggerDisplay("{Name}. State = {_state}. Error = {ErrorReason} ")]
```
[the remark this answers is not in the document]
marto83: thanks :)

src/IO.Ably.Shared/Types/MessageExtras.cs
marto83: Good spot. Thanks

marto83:
@QuintinWillison the builds are failing because the `delta-codec-dotnet` repo is private and the ci can't get access to it. Should we make it public?

src/IO.Ably.Shared/Result.cs
```diff
@@ -2,6 +2,66 @@
 
 namespace IO.Ably
 {
+    /// <summary>
+    /// A unit type is a type that allows only one value (and thus can hold no information)
+    /// Original source
+    /// https://github.com/louthy/language-ext/blob/0eb922bf9ca33944a5aa0745d791255321a4b351/LanguageExt.Core/DataTypes/Unit/Unit.cs.
+    /// </summary>
+    [Serializable]
+    public struct Unit : IEquatable<Unit>, IComparable<Unit>
```
codemerx: What is the purpose of this type? It looks like it can be seamlessly removed. Could it be a remainder of some old infrastructure?

src/IO.Ably.Shared/MessageEncoders/VCDiffEncoder.cs
```diff
@@ -0,0 +1,64 @@
...
+                return Result.Ok(new ProcessedPayload
+                {
+                    Data = result.AsByteArray(),
+                    Encoding = RemoveCurrentEncodingPart(payload),
+                });
+            }
+            catch (Exception ex)
+            {
+                var error =
+                    $"Payload Encoding: {context.PreviousPayload?.Encoding}. Payload: {context.PreviousPayload?.GetBytes().Length} bytes";
```
codemerx: Why is this reporting the previous message and not the current one?

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
```diff
@@ -1,51 +1,57 @@
+using System;
 using IO.Ably;
 using IO.Ably.Encryption;
 
 namespace IO.Ably.MessageEncoders
 {
     internal class CipherEncoder : MessageEncoder
     {
-        public override string EncodingName => "cipher";
+        public const string EncodingNameStr = "cipher";
```
codemerx: Maybe private?

src/IO.Ably.Shared/MessageEncoders/JsonEncoder.cs
```diff
@@ -9,48 +9,60 @@ namespace IO.Ably.MessageEncoders
 {
     internal class JsonEncoder : MessageEncoder
     {
-        public override string EncodingName => "json";
+        public const string EncodingNameStr = "json";
```
codemerx: Maybe private.

src/IO.Ably.Shared/MessageEncoders/MessageEncoder.cs
```diff
@@ -6,38 +6,56 @@ namespace IO.Ably.MessageEncoders
 {
     internal abstract class MessageEncoder
     {
-        internal ILogger Logger { get; set; }
+        public class ProcessedPayload : IPayload
```
codemerx: Why is this nested?

src/IO.Ably.Shared/MessageEncoders/MessageEncoder.cs
```diff
@@ -6,38 +6,56 @@ namespace IO.Ably.MessageEncoders
...
+        {
+            public object Data { get; set; }
+
+            public string Encoding { get; set; }
+
+            public ProcessedPayload()
```
codemerx: Do we need this constructor? Does it make any sense to allow an instance of this class to exist with no data in it?

src/IO.Ably.Shared/MessageEncoders/Base64Encoder.cs
```diff
@@ -5,39 +5,45 @@ namespace IO.Ably.MessageEncoders
...
-                RemoveCurrentEncodingPart(payload);
+                return Result.Ok(new ProcessedPayload()
+                {
+                    Data = data.FromBase64(),
+                    Encoding = RemoveCurrentEncodingPart(payload),
+                });
             }
 
-            return Result.Ok();
+            return Result.Ok(new ProcessedPayload(payload));
```
codemerx: Why teh copying here. Can't this one return `IPayload`?

src/IO.Ably.Shared/MessageEncoders/DecodingContext.cs
```diff
@@ -0,0 +1,53 @@
...
+        public DecodingContext(ChannelOptions options)
+        {
+            ChannelOptions = options ?? new ChannelOptions();
+        }
+    }
+
+    /// <summary>
+    /// Helpers methods for <see cref="DecodingContext"/>.
+    /// </summary>
+    internal static class EncodingDecodingContextHelpers
```
codemerx: Shouldn't this be `ChannelOptionsHelper`?

src/IO.Ably.Shared/MessageEncoders/DecodingContext.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        /// <summary>
+        /// Creates a new <see cref="DecodingContext"/> from the provided <see cref="ChannelOptions"/>.
+        /// </summary>
+        /// <param name="options">the <see cref="ChannelOptions"/> used in the new context.</param>
+        /// <returns><see cref="DecodingContext"/> created with passed Channel options.</returns>
+        public static DecodingContext ToEncodingDecodingContext(this ChannelOptions options)
```
codemerx: `ToDecodingContext`?

src/IO.Ably.Shared/Realtime/ChannelMessageProcessor.cs
```diff
@@ -69,29 +72,91 @@ public Task<bool> MessageReceived(ProtocolMessage protocolMessage, RealtimeState
...
+
                     break;
             }
 
             return Task.FromResult(true);
         }
+
+        private bool ValidateIfDeltaItHasCorrectPreviousMessageId(ProtocolMessage protocolMessage, LastMessageIds channelSuccessfulMessageIds)
+        {
+            if (protocolMessage.Messages == null || protocolMessage.Messages.Length == 0)
```
codemerx: Maybe this should be split in two methods - `isDelta()` and `HasCorrectPreviousMessageId`

src/IO.Ably.Shared/Result.cs
marto83: I use it in one place. I needed a type that has no value so I don't return void. 
image

src/IO.Ably.Shared/MessageEncoders/VCDiffEncoder.cs
marto83: Good point. I'll update it.

src/IO.Ably.Shared/MessageEncoders/MessageEncoder.cs
marto83: We do need the empty one. It is just in case the payload is `null`
image

src/IO.Ably.Shared/MessageEncoders/JsonEncoder.cs
marto83: Done!

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
marto83: Done!

src/IO.Ably.Shared/MessageEncoders/MessageEncoder.cs
marto83: It started there but I can't remember why. Moved it out.

src/IO.Ably.Shared/MessageEncoders/Base64Encoder.cs
marto83: that's how it was originally and I modified the payload as I went along. However at the end of the encoding processed because the wrapper doesn't know about the different steps I need to keep the original message because I might need to save it as the Last payload.

src/IO.Ably.Shared/MessageEncoders/DecodingContext.cs
marto83: Done

src/IO.Ably.Shared/MessageEncoders/DecodingContext.cs
marto83: Done

src/IO.Ably.Shared/Realtime/ChannelMessageProcessor.cs
marto83: the isDelta part needs to return the delta messageId. That's why I kept it as one method. In this class we won't separate the two anyway.

src/IO.Ably.Shared/Result.cs
codemerx: Yeah but why do that? you could have gone for returning `Result` only, right?

marto83: I see what you mean. If you return Result you can't mix it easily with Result<T>. That's the main reason I added Unit there. I make all the calls I need to check return Result<T> which makes it easier to handle.

src/IO.Ably.Shared/MessageEncoders/MessageEncoder.cs
codemerx: Wouldn't it have encoding if the data is null?

Issue #371, opened 2019-09-25
Error thrown when trying to initialize Ably object in Unity


Srushtika:
wrt the customer who reached out via support: https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/all/conversations/23788071585

He is seeing the following error on initializing Ably:

```
<code>
at (wrapper managed-to-native) System.Net.NetworkInformation.LinuxNetworkChange.CreateNLSocket()
at System.Net.NetworkInformation.LinuxNetworkChange.EnsureSocket () [0x00021] in <9f8c4786222746238eb929db40bf8dd1>:0
at System.Net.NetworkInformation.LinuxNetworkChange.Register (System.Net.NetworkInformation.NetworkAvailabilityChangedEventHandler d) [0x00000] in <9f8c4786222746238eb929db40bf8dd1>:0
at System.Net.NetworkInformation.LinuxNetworkChange.add_NetworkAvailabilityChanged (System.Net.NetworkInformation.NetworkAvailabilityChangedEventHandler value) [0x00000] in <9f8c4786222746238eb929db40bf8dd1>:0
at System.Net.NetworkInformation.NetworkChange.add_NetworkAvailabilityChanged (System.Net.NetworkInformation.NetworkAvailabilityChangedEventHandler value) [0x00021] in <9f8c4786222746238eb929db40bf8dd1>:0
at IO.Ably.Platform..cctor () [0x00001] in <84be4f219e3d4f0c84704765d3ad7590>:0
</code>
```

Some config details:

Windows 10, Io,Ably.dll NetStandard 2.0, ver 1.1.13

marto83:
I have tried to create a unity but I'm having issues with getting Unity working. 
I dug in to see what could be causing the error. 
It is caused by the initialisation of the library where it registers for Operating System events. 
It doesn't seem it's supported on the platform you are deploying to. 
One way to fix the issue is to pull the source of the library and remove line 15 and 16 from this file https://github.com/ably/ably-dotnet/blob/master/src/IO.Ably.NETStandard20/Platform.cs#L15

It help me find a better way you please provide a sample unity project that shows the error.

PR #392, opened 2020-01-10
Issue/380 allow to disable networkstate monitoring

### LINKED ISSUE #380: Expose a custom way to notify the library about OperatingSystemNetworkState
This is a placeholder ticket to look at different ways we can expose this. Customers have reported issues with edge cases on different platforms like. #371 and #332

marto83:
Fixes #380 

Originally we subscribed to Operating system events in the static constructor of the Platform class. However this proved problematic for Unity (#371) and developers have asked to control the notifications (#332). 
This PR addresses the issue. We still automatically subscribe for OS NetworkState change events but there is a new option `ClientOptions.AutomaticNetworkStateMonitoring` which can be switched off. 

If switched off developers will need to notify Ably about changes in the network state. Otherwise the library will not behave as expected until we also implement (https://docs.ably.io/client-lib-development-guide/features/#RTN23). To do that one can call `Connection.NotifyOperatingSystemNetworkState`.

### DIFF
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -297,6 +297,15 @@ namespace IO.Ably
         [Obsolete("We will no longer support the SynchronizationContext in the library. This property will be removed in future versions")]
         public SynchronizationContext CustomContext { get; set; }
 
+        /// <summary>
+        /// Allows developers to disable Automatic network state monitoring. When set to `true` the library will not subscribe to
+        /// `NetworkChange.NetworkAvailabilityChanged` events. Developers can manually notify the Connection for changes by calling
+        /// `Connection.NotifyOperatingSystemNetworkState(NetworkState.Online or NetworkState.Offline)`
+        /// Unity and some Mono environments don't have the feature implemented and the library cannot initialize.
+        /// Default: false.
+        /// </summary>
+        public bool DisableAutomaticNetworkStateMonitoring { get; set; } = false;
+
         [JsonIgnore]
         internal Func<DateTimeOffset> NowFunc
         {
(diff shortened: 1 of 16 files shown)


QuintinWillison:
@marto83, can you describe in more detail how this fixes #380 as that issue describes exposing a means to notify the library about state but your change seems to expose a means to prevent the library from listening to state... or am I missing something?

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -297,6 +297,15 @@ public bool UseBinaryProtocol
...
         public SynchronizationContext CustomContext { get; set; }
 
+        /// <summary>
+        /// Allows developers to disable Automatic network state monitoring. When set to `true` the library will not subscribe to
+        /// `NetworkChange.NetworkAvailabilityChanged` events. Developers can manually notify the Connection for changes by calling
+        /// `Connection.NotifyOperatingSystemNetworkState(NetworkState.Online or NetworkState.Offline)`
+        /// Unity and some Mono environments don't have the feature implemented and the library cannot initialize.
+        /// Default: false.
+        /// </summary>
+        public bool DisableAutomaticNetworkStateMonitoring { get; set; } = false;
```
QuintinWillison: Assuming this name is not explicitly specified elsewhere then I would prefer to drop the `Disable` prefix (default would become `true`, obviously).

src/IO.Ably.Shared/IPlatform.cs
```diff
@@ -11,5 +11,7 @@ internal interface IPlatform
         string PlatformId { get; }
 
         ITransportFactory TransportFactory { get; }
+
+        void RegisterOsNetworkStateChanged();
```
QuintinWillison: Could you document when this gets called and why? Also when it will not get called and/or its relationship to the boolean in `ClientOptions`.

PR #399, opened 2020-02-07
Feature/channel params

marto83:
This branch builds on top of the delta branch and addes ChannelParams and ChannelModes.

I just realised I still need to remove the extension methods for Rewind I started adding and also there is one failing test with deltas which I think is to do with the test payloads I'm using rather than the logic.


src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
```diff
@@ -23,6 +23,7 @@ internal class MessageHandler
             new JsonEncoder(), new Utf8Encoder(), new CipherEncoder(), new VCDiffEncoder(), Base64Encoder,
         };
 
+        public List<MessageEncoder> Encoders = new List<MessageEncoder>();
```
codemerx: Is this used anywhere?

src/IO.Ably.Shared/Rest/ChannelOptions.cs
```diff
@@ -20,6 +26,27 @@ public class ChannelOptions
...
 
+        /// <summary>
+        /// Params allows custom parameters to be passed to the server when attaching the channel.
+        /// In that list are 'delta' and 'rewind'. For more information about channel params visit
+        /// https://www.ably.io/documentation/realtime/channel-params.
+        /// </summary>
+        public ChannelParams Params
+        {
+            get => _params;
+            set => _params = value ?? new ChannelParams();
```
codemerx: Why public setter?

src/IO.Ably.Shared/Rest/ChannelOptions.cs
```diff
@@ -20,6 +26,27 @@ public class ChannelOptions
...
+        }
+
+        /// <summary>
+        /// Channel Modes like Params are passed to the server when attaching the channel.
+        /// They let specify how the channel will behave and what is allowed. <see cref="ChannelMode"/>.
+        /// </summary>
+        public ChannelModes Modes
+        {
+            get => _modes;
+            set => _modes = value ?? new ChannelModes();
```
codemerx: Why public setter?

src/IO.Ably.Shared/Rest/ChannelOptions.cs
```diff
@@ -88,4 +128,73 @@ public override int GetHashCode()
             }
         }
     }
+
+    /// <summary>
+    /// Helper methods to make adding channel options easier.
+    /// </summary>
+    public static class ChannelOptionsExtensions
```
codemerx: Maybe `WithDelta` method would make sense too.

src/IO.Ably.Shared/Rest/ChannelOptions.cs
```diff
@@ -88,4 +128,73 @@ public override int GetHashCode()
...
+        /// <param name="options">the <see cref="ChannelOptions"/> that will be modified.</param>
+        /// <param name="rewindBy">by how much should we rewind.</param>
+        /// <param name="limit">the max number of messages that should be returned. Max: 100.</param>
+        /// <returns>the same ChannelOptions object so other calls can be chained.</returns>
+        public static ChannelOptions WithRewind(this ChannelOptions options, TimeSpan rewindBy, int? limit)
+        {
+            var result = options.WithParam("rewind", rewindBy.TotalSeconds + "s");
+            if (limit.HasValue && limit > 0)
+            {
+                var l = Math.Max(limit.Value, 100);
```
codemerx: Should be `Math.Min`. But should we enforce this on the client side?

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: to make it easier to set. Also it let's do you dict initialisation expressions like
new ChannelOptions() { 
Params = new ChannelParams { { "key": "value"}, {"key2": "value"}}
}

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: I need to remove these. I like them but based on the spec and conversation with Simon and Quintin we are not adding any helper methods and keeping the implementation just based on the Channel Params dictionary

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: good spot. I'll need to remove this.

QuintinWillison:
I see what you did with https://github.com/ably/ably-dotnet/pull/399/commits/f1fff9008524ef92547aa33562feb74ba60e40c7, @marto83 - a fair compromise. Thanks!

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: Done!

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: Done!

src/IO.Ably.Shared/Rest/ChannelOptions.cs
marto83: Just to give the user more options. Same as ChannelParams

src/IO.Ably.Shared/MessageEncoders/MessageHandler.cs
marto83: Nope. I removed it.

src/IO.Ably.Shared/Types/ChannelModes.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                yield return ChannelMode.PresenceSubscribe;
+            }
+        }
+    }
+
+    /// <summary>
+    /// Set of Channel modes. It's used inside ChannelOptions.
+    /// </summary>
+    public class ChannelModes : HashSet<ChannelMode>
```
QuintinWillison: I'm not really sure what the presence of this class adds. Why can't you just use `HashSet<ChannelMode>` where needed? Same goes for the read-only variants. The fact that there are no fields, the constructors are empty and the constructors only call base is a sign to me that these classes just add noise.

src/IO.Ably.Shared/Defaults.cs
```diff
@@ -8,8 +9,7 @@ namespace IO.Ably
 {
     internal static class Defaults
     {
-        internal const int ProtocolMajorVersion = 1;
-        internal const int ProtocolMinorVersion = 1;
+        internal static float ProtocolVersionNumber = 1.1F;
```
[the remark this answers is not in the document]
marto83: I did start with System.Version but then figured out all I use it for is to compare it to 1.2 and print it as a string so `float` seemed more than adequate :)

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -21,6 +22,12 @@ internal class RealtimeChannel : EventEmitter<ChannelEvent, ChannelStateChange>,
         private ChannelOptions _options;
         private ChannelState _state;
 
+        /// <summary>
+        /// True when the channel moves to the @ATTACHED@ state, and False
+        /// when the channel moves to the @DETACHING@ or @FAILED@ states.
+        /// </summary>
+        internal bool AttachResume { get; set; } = false;
```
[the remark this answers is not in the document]
marto83: It's more explicit when you have a variable and also makes testing much easier. I can just set it to test AttachResume and not worry about the channel state. 
Also there is an edge case that is more difficult if you don't have a variable. If the channel goes to `Attaching` after `Attached` you have it as `true` but if the channel goes to `Attaching` after `Initialized` you need to have it as `false`. That's why it's better to keep the state separate.

src/IO.Ably.Shared/Types/ChannelModes.cs
marto83: Here I followed how it was done in Java. It did make some things a bit easier. I don't think I use the enumerable constructor but the other one that takes params ChannelMode[] is quite handy. 
I'm happy to remove it if you think it's unnecessary.

QuintinWillison: I think my main mistake has been reviewing this stuff without looking at the Java alongside. Leave it as is I think.

src/IO.Ably.Shared/Defaults.cs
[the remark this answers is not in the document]
QuintinWillison: I mean, it works. So, yeah, leave it if you're happy. It's internal at least.

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -278,8 +278,10 @@ public bool UseBinaryProtocol
 
         /// <summary>
         /// Switched on IdempotentRest publishing. Currently switched off by default in libraries with version less than 1.2.
+        /// Default before 1.2: false.
+        /// Default after 1.2: true.
```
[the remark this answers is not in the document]
marto83: Done!

src/IO.Ably.Shared/Realtime/LastMessageIds.cs
```diff
@@ -7,10 +7,15 @@ internal class LastMessageIds
     {
         public static LastMessageIds Empty = new LastMessageIds();
 
-        public string LastMessageId { get; }
+        public string LastMessageId { get; internal set; } // To make testing easier.
```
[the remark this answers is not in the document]
marto83: I've updated it.

Issue #384, opened 2019-11-11
Could not load type of field IO.Ably.Transport.ConnectionManager


mattheworiordan:
Customer reported:

> When trying to install your NuGet package in a NET Standard class library which I'm linking from a Xamarin Android project... I'm getting the following exception when I try to instantiate an AblyRealtime. I've tried everything I can think of; nuked packages/bin/obj, uninstalled/reinstalled the package, etc... but I'm stuck. Any help would be appreciated! Thanks

> Could not load type of field 'IO.Ably.Transport.Connection
Manager:<PendingMessages>k__BackingField' (2) due to: Could not resolve type with token 010000a6 from typeref (expected class 'System.Collections.Generic.Queue`1' in assembly 'mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e') assembly:mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e type:System.Collections.Generic.Queue`1 member:(null)

The customer did resolve the issue though:

> Thanks for getting back to me. It seems like it was related to your NuGet package being built with Visual Studio 2019 and my app being built with Visual Studio 2017...
> Here's the SO post that pointed me in the right direction: https://stackoverflow.com/a/58064929/2012597

Internal conversation: https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/conversation/24499044632

marto83:
I've updated the readme to alert people and let them know how to work around the issue. I have 2 options:
1. To remove references to Queue and replace them with a something that is not problematic.
2. Go back to building the library with the older MsBuild. I'll have to check if this is not going to create new issues.

mattheworiordan:
Ok, thanks, keep me / this issue posted.

PR #401, opened 2020-02-21
Issue/384

marto83:
I've updated the library so we no longer use Queue<T> (there was only 1 place) and it will fix the issue for customers building Xamarin projects with VS2017

Fixes: #384


marto83:
> Just one question out of curiosity... what's going on with all the script files being removed from project files? related or simultaneous house keeping?

These were files I had removed ages ago. However I noticed there were still references in the solution files. I like to clean up as we go along.

PR #403, opened 2020-02-26
Issues/402

### LINKED ISSUE #402: Connection moves to Failed when there is no internet connection
A customer reported the following:
> when i disconnect my computer from the internet, the ably library switches to Failed state with a BadeGateway exception, while giving me this link(https://support.ably.io/support/solutions/articles/3000092531). According to docs, or atleast that's how I understand it, Failed state should only occur on major errors, but a user loosing its internet connection is a totally valid scenario. 

After doing some manual tests, I managed to recreate the issue which happens when the socket connection is initialised.

marto83:
Fixes #402

### DIFF
--- a/src/IO.Ably.Shared/AblyRest.cs
+++ b/src/IO.Ably.Shared/AblyRest.cs
@@ -363,7 +363,8 @@ namespace IO.Ably
 
             try
             {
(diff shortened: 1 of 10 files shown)


marto83:
I've tested this branch in a console app. If I block the internet connection with the firewall the library successfully goes to Disconnected and then transitions to Suspended.

PR #407, opened 2020-03-11
Version 1.2

QuintinWillison:


build.fsx
```diff
@@ -117,8 +117,10 @@ let nugetRestore solutionFile =
   |> Proc.run // start with the above configuration
 
 Target.create "Restore" (fun _ ->
-    DotNet.restore id "src/IO.Ably.sln"
+    //DotNet.restore id "src/IO.Ably.sln"
```
QuintinWillison, line 120: @marto83 Is there any reason we can't just remove this line entirely? It's commented out without explanation so this'll just end up hanging around here forever! So if you don't want to remove it then can you please comment why it's left hanging around? Thanks.

marto83: I'm happy to remove it. I wasn't sure whether it would work on the build server. I'll push a commit to get rid of it.

QuintinWillison: removed https://github.com/ably/ably-dotnet/pull/407/commits/9b9586ecc5549af9b519fedcc1605d99bbd0a47a

Issue #409, opened 2020-03-20
Unexpected error: The current state [Suspended] does not allow messages to be sent


QuintinWillison:
We've had a report come in from a customer ([internal link](https://app.intercom.com/a/apps/ua39m1ld/inbox/inbox/conversation/26237569850)) who is seeing this error message (Ably [error code 500](https://help.ably.io/error/500)) being reported periodically by devices running their app in the wild.

The customer states:

> I believe the device has an internet connection because the rest of the app is able to connect to communicate with our web app server. Also, the error occurred either: 1. During the creation of the AblyRealtime object 2. When calling "Channels.Get" or 3. when Subscribing to a channel.

To which @paddybyers has already responded to them with:

> You should only reach the suspended state after trying continuously to connect for 2 minutes, so it shouldn't be possible for it to be your case (1) - if that's happening then it's definitely a bug.
Similarly, your case (2) shouldn't give this error - that's just instancing a channel object.
So I would guess that it's during a subscribe attempt, your case (3). The error message looks like it's misleading if that's the case - we'll have a look at that. It does send a message under the hood in order to subscribe, but the error message looks confusing.

The customer has also told us:

> I am using the Xamarin Nuget library v1.1.18. This morning's incident was on iOS 13.3.1.

marto83:
The creation of ably or Channels.Get can't error. The only possibility is Subscribing to a new channel because that tries to send an Attach message. 

I had a look at the code and it is definitely the case. It's a bug which is fixed in the upcoming 1.2 release but as I'm not sure when this is coming I'll get it fixed and release 1.1.19.

In the meanwhile, you can catch the error and call `client.Connect()`. This will transition the and then try a second time to Subscribe to the channel. Something like 👇 

    var client = new AblyRealtime();
    var channel = client.Channels.Get("myChannel");
    try {
         channel.Subscribe(handler);
    } catch (AblyException ex) { 
        client.Connection.Once(ConnectionEvent.Connecting, _ => channel.Subscribe(handler));
        client.Connect();
    }

marto83:
Fixed in PR #410

PR #410, opened 2020-03-22
Close issue #409. Call error callback instead of through exception wh…

marto83:


QuintinWillison:
Once you've merged this in to `master` then probably priority to get the integration branch rebased so any conflicts are dealt with now while it's fresh in your head.

Issue #419, opened 2020-06-12
Failed to resolve IO.Ably.DeltaCodec


marto83:
Release 1.2.0 nuget package is missing IO.Ably.DeltaCodec for the following Platforms:

- .Net framework
- Xamarin.iOS
- Xamarin.Android

You will get the following error:
`Failed to resolve assembly: 'IO.Ably.DeltaCodec, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'`

This will be fixed with a new release very soon.

PR #420, opened 2020-06-12
Update build to copy all required dlls when packaging

marto83:
Fixes: #419
The problem was that for Xamarin and .net framework we have an extra step that merges newtonsoft json and it didn't copy all the required dlls.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -327,6 +327,11 @@ Target.create "Package - Merge json.net" (fun _ ->
               |> Seq.map (Path.combine "src")
               |> Seq.map (fun path -> sprintf "%s/bin/%s" path buildMode)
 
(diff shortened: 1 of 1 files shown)


QuintinWillison:
ok, I've answered my first question - which simply makes me realise that `paths` is not an obvious name, given what it represents. Perhaps consider renaming that to make it clearer - e.g. `pathsToMergeTo` or something like that.

My second question still stands re naming.

marto83:
We are packaging the json.net library because with the old framework and xamarin it used to create many issues when people were using a different version of the json library. We merge the Json.net dll into IO.Ably.dll and the merge rewrites all the json.net namespaces to make them internal.

PR #432, opened 2020-08-01
Fix Property Typo

### LINKED ISSUE #427: Typo in property name for log handler
Currently [the property](https://github.com/ably/ably-dotnet/blob/1140a3c6a47e285adade291ac6269fa93abc393a/src/IO.Ably.Shared/ClientOptions.cs#L108) on ClientOptions is called `LogHander`.

It should be `LogHandler`, per [TO3c](https://docs.ably.io/client-lib-development-guide/features/#TO3c).

The current, incorrectly named property should be deprecated for now.

sacOO7:
- Added new property LogHandler to ClientOptions.
- Marked LogHander property as obsolete, delegated implementation to LogHandler.
- Refactored the change in relevant property references.
- Fixes #427.

### DIFF
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -105,7 +105,7 @@ namespace IO.Ably
         /// Default: null.
         /// </summary>
         [JsonIgnore]
-        public ILoggerSink LogHandler { get; set; }
+        public ILoggerSink LogHander { get; set; }
 
         /// <summary>
         /// For development environments only. Allows a non default host for the rest service.
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -105,8 +105,20 @@ namespace IO.Ably
         /// Default: null.
         /// </summary>
         [JsonIgnore]
-        [ObsoleteAttribute("This property is obsolete/deprecated", false)]
-        public ILoggerSink LogHander { get; set; }
+        [Obsolete("LogHander will be removed in future versions, please use LogHandler instead.")]
+        public ILoggerSink LogHander
+        {
+            get => LogHandler;
+
+            set => LogHandler = value;
+        }
+
+        /// <summary>
+        /// Log handler: allows the client to intercept log messages and handle them in a client-specific way.
+        /// Default: null.
+        /// </summary>
+        [JsonIgnore]
+        public ILoggerSink LogHandler { get; set; }
 
         /// <summary>
         /// For development environments only. Allows a non default host for the rest service.
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -105,6 +105,7 @@ namespace IO.Ably
         /// Default: null.
         /// </summary>
         [JsonIgnore]
+        [ObsoleteAttribute("This property is obsolete/deprecated", false)]
         public ILoggerSink LogHander { get; set; }
 
         /// <summary>
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -105,7 +105,7 @@ namespace IO.Ably
         /// Default: null.
         /// </summary>
         [JsonIgnore]
-        public ILoggerSink LogHander { get; set; }
+        public ILoggerSink LogHandler { get; set; }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


QuintinWillison (changes requested):
The old interface should be deprecated, not removed, per my commentary in #427. Thanks.

sacOO7:
Hi @QuintinWillison , I have marked current properly as deprecated. Compiler will show warning if it's used anywhere in the code. (We can make it to show as error if required). I am not sure if we need to introduce new property ```LogHandler``` as a part of this PR.

QuintinWillison:
Yes, the introduction of the new property is essential for this fix!

What I was pointing out is that simultaneously the deprecation needs to be introduced. Thus, overall, the fix is atomic in that new users of this API can use the correct name and those with existing codebases will not completely break as they will get a deprecation warning.

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -105,6 +105,7 @@ internal string GetClientId()
         /// Default: null.
         /// </summary>
         [JsonIgnore]
+        [ObsoleteAttribute("This property is obsolete/deprecated", false)]
```
QuintinWillison, line 108: The deprecation message has to point the user to the property name they should be using, please.

QuintinWillison (changes requested):
As stated above this PR also needs to include the new property interface and implementation. The deprecated property implementation should internally delegate to the new implementation.

src/IO.Ably.Tests.Shared/ClientOptionsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Build.Logging;
+using Xunit;
+
+namespace IO.Ably.Tests.Shared
+{
+    public class ClientOptionsTests
+    {
+        [Fact]
+        [Obsolete]
+        public void Options_WithLogHanderReturns_LogHandlerWithSameLoggerSink()
```
QuintinWillison, line 14: I'm not sure what the naming convention here is but wouldn't this be more logical?
```suggestion
        public void Options_WithLogHander_ReturnsLogHandlerWithSameLoggerSink()
```

src/IO.Ably.Tests.Shared/ClientOptionsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            var testLoggerSink = new TestLoggerSink();
+            var options = new ClientOptions()
+            {
+                LogHander = testLoggerSink
+            };
+
+            Assert.Same(options.LogHander, testLoggerSink);
```
QuintinWillison, line 22: This feels over the top in terms of validation. Wouldn't `Assert.Equal` achieve what you really need to check? I'm assuming the `LogHandler` property is not designed to copy the provided instance, therefore a simple check of the reference should suffice and makes less assumptions about the implementation of the object instances being verified?

QuintinWillison (changes requested):
Thanks for adding tests but that has opened up another couple of small queries from me. 😁

src/IO.Ably.Tests.Shared/ClientOptionsTests.cs
sacOO7: Sure, I guess I was referring to the test naming convention from https://github.com/ably/ably-dotnet/blob/344d1d2741daee45ca37b411c07cfc35c4e88a8b/src/IO.Ably.Tests.Shared/ApiKeyTests.cs#L17.
But surely, makes sense to change it to the above name.

src/IO.Ably.Tests.Shared/ClientOptionsTests.cs
sacOO7: I think equal checks if two objects are equal, Assert.Same checks if they belong to the same instance.

sacOO7: I think Assert.Equal method can have typical checks for instance check (uses overridden equals method internally), hashcode check and may check for internal data. Assert.Same is more simple IMO.

sacOO7: You can say, ```Assert.Same``` checks if references are the same.

QuintinWillison: Yup, I was wrong. Sorry. Read an incorrect blog post somewhere. 🤦

sacOO7: Yeah no worries : )

PR #439, opened 2020-09-21
Fix auth token clock-skew

### LINKED ISSUE #303: Improve handling of clock skew
See `RSA4b1` as part of 1.2 spec changes in https://github.com/ably/docs/pull/622/files

Given this impacts users now with significant clock skew, we aim to release a fix for this now.

sacOO7:
- Removed token expiry check from auth flow
- Fixes #303

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -21,13 +21,15 @@ namespace IO.Ably
             _rest = rest;
             Logger = options.Logger;
             ServerTime = () => _rest.TimeAsync();
-            ServerTimeOffset = () => null;
+            ServerNow = () => null;
             Initialise();
         }
 
         protected Func<Task<DateTimeOffset>> ServerTime { get; set; }
 
-        protected Func<DateTimeOffset?> ServerTimeOffset { get; private set; }
+        private bool HasServerTime => ServerNow().HasValue;
+
+        protected Func<DateTimeOffset?> ServerNow { get; private set; }
 
         internal Func<DateTimeOffset> Now { get; set; }
 
@@ -91,10 +93,23 @@ namespace IO.Ably
             LogCurrentAuthenticationMethod();
         }
 
-        private async Task SetServerTimeOffset()
+        internal async Task SetServerTime()
         {
-            TimeSpan diff = Now() - await ServerTime();
-            ServerTimeOffset = () => Now() - diff;
+            var serverTime = await ServerTime();
+            if (Logger.IsDebug)
+            {
+                Logger.Debug("Received server time: " + serverTime);
+            }
+
+            TimeSpan diff = Now() - serverTime;
+
+            if (Logger.IsDebug)
+            {
+                Logger.Debug("Server time differs from device Now by: " + diff);
+            }
+
+            // TODO: Save servertime and check if now has drastically changed
+            ServerNow = () => Now() - diff;
         }
 
         private AuthMethod CheckAndGetAuthMethod()
@@ -161,6 +176,11 @@ namespace IO.Ably
             }
             else
             {
+                if (HasServerTime == false && Options.QueryTime == true)
+                {
+                    await SetServerTime();
+                }
+
                 var currentValidToken = await GetCurrentValidTokenAndRenewIfNecessaryAsync();
                 if (currentValidToken == null)
                 {
@@ -178,7 +198,12 @@ namespace IO.Ably
                 throw new AblyException("AuthMethod is set to Auth so there is no current valid token.");
             }
 
-            return await Task.FromResult(CurrentToken);
+            if (CurrentToken.IsValidToken(ServerNow()))
+            {
+                return CurrentToken;
+            }
+
+            return await RenewToken();
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -91,10 +93,23 @@ internal void Initialise()
...
+
+            TimeSpan diff = Now() - serverTime;
+
+            if (Logger.IsDebug)
+            {
+                Logger.Debug("Server time differs from device Now by: " + diff);
+            }
+
+            // TODO: Save servertime and check if now has drastically changed
+            ServerNow = () => Now() - diff;
```
sacOO7, line 112: Can we move the definition for this at class level, we can store the diff in a variable that can be accessed from there.

marto83: We can but I don't see any benefit at the moment. We can do it when we implement the check whether the local time has drifted.

marto83:
I updated the PR implementing RSA4b1 so we only check the expiry of the time if we have server time and the token has an expiry field. I had to fix some tests on the way because they were expecting the check to happen automatically.

sacOO7:
Its easy to understand that way. I think the confusion was caused because
of ServerNow implementation. I thought it's a class variable. Tbh, I am new
to this kind of pattern ( where global class fun. definition is dynamically
set in a local function )

src/IO.Ably.Shared/TokenDetails.cs
```diff
@@ -90,9 +81,20 @@ public TokenDetails(string token, Func<DateTimeOffset> nowFunc)
...
+            Expire(Defaults.NowFunc()());
         }
 
+        /// <summary>
+        /// Expires the current token.
+        /// </summary>
+        /// <param name="now">Current date and time.</param>
+        public void Expire(DateTimeOffset now)
+        {
+            Expires = now.AddDays(-1);
```
QuintinWillison, line 93: This feels a little bit arbitrarily 'magic'. Would it not be safer to have a discrete boolean field to indicate explicitly expired? This mutation is effectively meaningless and could actually be misleading in some circumstances.

src/IO.Ably.Shared/TokenDetails.cs
```diff
@@ -155,21 +157,23 @@ public static class TokenDetailsExtensions
...
-            if (token.Expires == DateTimeOffset.MinValue)
+            // (RSA4b1) We can only check validity of the token if it is a full TokenDetails object
+            // and we already have the server time
+            if (serverTime is null || token.CanBeUsedToCheckExpiry == false)
             {
                 return true;
             }
 
-            return !token.IsExpired(now);
+            return token.IsExpired(serverTime.Value) == false;
```
QuintinWillison, line 176: Not sure why the explicit `== false` here, as I've commented on elsewhere. Why not?:

```suggestion
            return !token.IsExpired(serverTime.Value);
```

QuintinWillison (changes requested):
It's difficult to review this change at a high level because there's a lot going on and I don't have enough knowledge of the inner workings of this library to assess impact, therefore have to trust you in terms of general architectural fit.

Hence, my comments are more smaller nit picks. Sorry!

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -161,6 +176,11 @@ internal async Task AddAuthHeader(AblyRequest request)
             }
             else
             {
+                if (HasServerTime == false && Options.QueryTime == true)
```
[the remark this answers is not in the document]
marto83: 😂 I love boolean logic discussions. 
for `Options.QueryTime` I have to explicitly compare to `true` because it is nullable and it complains if you just leave it. 
For `!HasServerTime` vs `HasServerTime == false` it's just a personal preference because it makes it easier to see. I don't mind changing it if there is a convention in Ably otherwise I'd prefer to leave it `== false`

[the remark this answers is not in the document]
QuintinWillison: Understood if coercion involved, but when comparing fields where the assigned type is boolean then no need.

src/IO.Ably.Shared/TokenDetails.cs
marto83: Agree. I was not really fixing that function but just needed to be able to pass the date. I'll check where it is used and whether I can remove it.

marto83: I've removed the method completely. It was only used internally.

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: Its easy to understand that way. I think the confusion was caused because
of ServerNow implementation. I thought it's a class variable. Tbh, I am new
to this kind of pattern ( where global class fun. definition is dynamically
set in a local function )

sacOO7:
@marto83 I took a pull and try running the tests, seens ```RealTimeWithAuthUrl_WhenTokenExpired_ShouldRetry_WhenRetryFails_ShouldSetError``` and 
```Auth_WithRealtimeClient_WhenAuthFails_ShouldTransitionToOrRemainInTheCorrectState``` is failing

I was able to make the first one work after increasing the awaiter timeout. Not sure about the second one.

File: src/IO.Ably.Shared/AblyAuth.cs (hunk 1 above)
Comment (marto83) [reply src/IO.Ably.Shared/AblyAuth.cs:107]: no problem. For sure, it's not the most obvious way of doing it and it could be done much better. Could you please add a issue about it.

File: src/IO.Ably.Shared/AblyAuth.cs [hunk 5]
```diff
@@ -91,10 +93,23 @@ internal void Initialise()
...
+            }
+
+            TimeSpan diff = Now() - serverTime;
+
+            if (Logger.IsDebug)
+            {
+                Logger.Debug("Server time differs from device Now by: " + diff);
+            }
+
+            // TODO: Save servertime and check if now has drastically changed
```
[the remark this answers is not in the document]
marto83: Create issue #441 441

marto83 (approved):
Looks good

marto83:
@sacOO7 regarding your message about failing tests. I've created another branch to look at all the test separately and make sure our current CI builds pass before making the next release.

src/IO.Ably.Shared/AblyAuth.cs
QuintinWillison: @sacOO7 we're going to get this PR merged but I would appreciate it if you could please resolve this conversation once you are happy and/or have created this issue. Thanks.

QuintinWillison (approved):

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: will fix as a part of https://github.com/ably/ably-dotnet/issues/441 as per conversation

PR #440, opened 2020-09-22
Issue/437

### LINKED ISSUE #437: Failing internet check
The client checks for internet connection availability by querying https://internet-up.ably-realtime.com/is-the-internet-up.txt and appears to be failing checks due to a newline being present in the response.

```
2020-09-02 18:07:02.1274440 :          : Ably.IO #1 : Warning:06:07:02.127 Cannot get a valid internet connection. Response: yes

```
https://github.com/ably/ably-dotnet/blob/main/src/IO.Ably.Shared/AblyRest.cs#L370

The response probably needs to be trimmed before comparison to let this work. I don't quite understand why this wouldn't have been an issue before as the response from this endpoint hasn't changed since 2015.

marto83:
Fixes #437 

I've trimmed the result and added tests

### DIFF
--- a/src/IO.Ably.Shared/Extensions/StringExtensions.cs
+++ b/src/IO.Ably.Shared/Extensions/StringExtensions.cs
@@ -29,6 +29,16 @@ namespace IO.Ably
                    || (input.StartsWith("[") && input.EndsWith("]"));
         }
 
+        public static string SafeTrim(this string input)
+        {
+            if (input.IsEmpty())
+            {
+                return input;
+            }
+
+            return input.Trim();
+        }
+
         public static string JoinStrings(this IEnumerable<string> input, string delimiter = ", ")
         {
             if (input == null)
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -1273,7 +1273,6 @@ namespace IO.Ably.Tests.Realtime
         {
             var client = await GetRealtimeClient(Protocol.Json, (options, _) =>
             {
-                options.SkipInternetCheck = false;
                 options.FallbackHosts = new string[] { };
             });
 
@@ -1301,7 +1300,7 @@ namespace IO.Ably.Tests.Realtime
         [Trait("issue", "437")]
         public async Task CanConnectToAbly_ShouldReturnTrue()
         {
-            var restClient = await GetRestClient(Protocol.Json, options => options.SkipInternetCheck = false);
+            var restClient = await GetRestClient(Protocol.Json);
 
             var result = await restClient.CanConnectToAbly();
             result.Should().BeTrue();
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -1267,6 +1267,46 @@ namespace IO.Ably.Tests.Realtime
             await client.WaitForState(ConnectionState.Suspended, TimeSpan.FromSeconds(10));
         }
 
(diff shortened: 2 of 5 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -1267,6 +1267,46 @@ public async Task WhenInternetConnectionIsLost_WithoutOSNotification_ShouldBehav
             await client.WaitForState(ConnectionState.Suspended, TimeSpan.FromSeconds(10));
         }
 
+        [Fact]
+        [Trait("issue", "437")]
+        public async Task WhenTimeoutMessageReceived_ShouldReconnectSuccessfully()
+        {
+            var client = await GetRealtimeClient(Protocol.Json, (options, _) =>
+            {
+                options.SkipInternetCheck = false;
```
QuintinWillison, line 1276: Is this not the default?

src/IO.Ably.Shared/Extensions/StringExtensions.cs
```diff
@@ -29,6 +29,16 @@ public static bool IsJson(this string input)
                    || (input.StartsWith("[") && input.EndsWith("]"));
         }
 
+        public static string SafeTrim(this string input)
+        {
+            if (input.IsEmpty())
```
QuintinWillison, line 34: Looking at the docs for `IsEmpty()` and `Trim()`, I'm not sure why you need this step. Why is this method declared 'Safe'?

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
marto83: You are right. I'll remove it.

src/IO.Ably.Shared/Extensions/StringExtensions.cs
marto83: Usually if you call `Trim` when you have a `null` string you will get a null reference exceptions. This way I make sure that cannot happen because `IsEmpty` is also an extension method and deals with nulls already. I can call `SafeTrim` on any string without worrying about unexpected values.

QuintinWillison: I don't follow you. If `input` is `null` then there is no instance to call `IsEmpty()` on, surely? I would expect you to be using the static [`IsNullOrEmpty(String)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=netcore-3.1) method if this is your concern.

This is a classic example of where I would like to see a unit test focussed on validating this method. Have you got one that checks its handling of `null` for `input`?

QuintinWillison: Perhaps I get your point regarding extension method, given you can only call the extension if you have an instance, but then that draws me to the calling site where presumably you're still not guarding against `response.TextResponse` being `null`... so I'm still really confused as to what's safe at all about this method! 😕

marto83: I've added a test case to prove it works. `IsEmpty()` is an extension method so the compiler turns it into `StringExtensions.IsEmpty(input)` and that's why you can safely call it on null strings.

marto83: Have a look at https://github.com/ably/ably-dotnet/pull/440/files#diff-b05e864eb6dc1d5b02c6399dffd932a4R19

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
marto83: I've removed calls to SkipInternetCheck and tests work 👍

QuintinWillison (approved):

Issue #442, opened 2020-09-24
Migrate CI to Github actions


marto83:
Move all build to github actions:
* Windows net framework
* Windows net standard
* Linux net standard

Fix unit tests for Linux net standard.

PR #443, opened 2020-09-24
Issue/435 fix build and tests

### LINKED ISSUE #435: [CI] Fix build and tests
- From https://github.com/ably/ably-dotnet/pull/431#issuecomment-672181808.
- Need to fix build on local and CI (Also make sure, it works across different platforms) 
- Make sure every test runs successfully even if run in isolation.

-  Failing tests - https://ably-real-time.slack.com/archives/C0174D4AP4P/p1596120980020500
-  CI - https://ably-real-time.slack.com/archives/C0174D4AP4P/p1596122321028600
-  CI - https://ably-real-time.slack.com/archives/C0174D4AP4P/p1596187625031700

marto83:
The aim of this is to fully fix #435 but I need to spend some more time because 1 particular unit tests always fails in linux and I suspect it is because it's accessing embedded resources. For that reason I created #442.

- Here I did the following (please review a commit at a time):
- Update SendMessage command to handle failed message send
- Simplify tests and remove redundant client.Close() calls
- Add some logging to the idempotent publishing test
- Improve ShouldDisconnectIfConnectionIsNotEstablishedWithInDefaultTimeout
- Change linux build to use the same build script as windows build


src/IO.Ably.Tests.DotNetCore20/IO.Ably.Tests.DotNetCore20.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
QuintinWillison, line 4: Given the name of this assembly / project, this looks a bit suspect. Intentional?

.circleci/config.yml
```diff
@@ -3,13 +3,13 @@ jobs:
...
+      - image: mcr.microsoft.com/dotnet/core/sdk:3.1
     environment:
       DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
     steps:
       - checkout
       - run: git submodule update --init --recursive
-      - run: chmod +x .circleci/run_tests.sh
-      - run: cd .circleci && ./run_tests.sh
+      - run: chmod +x ./build.sh
```
QuintinWillison, line 13: I realise you're simply copying the pattern from previous, but I'm surprised that Git couldn't carry that attribute through for us. Somewhat automatic on macOS but if you commit from Windows I think you can achieve the same [using Git's CLI](https://git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---chmod-x).

QuintinWillison (approved):
Only a couple of minor observations, otherwise I don't really understand what I'm looking at, nor have I tried to test (via Amazon Workspace, or otherwise).

.circleci/config.yml
marto83: Cool. I've setup a linux machine to run the test and made the change there. 👍🏼. I learnt something new today

src/IO.Ably.Shared/Transport/MsWebSocketTransport.cs
```diff
@@ -127,8 +127,13 @@ public void Close(bool suppressClosedEvent = true)
...
+        public Result Send(RealtimeTransportData data)
         {
+            if (_socket is null)
+            {
+                return Result.Fail($"Cannot send message. Socket instance is null. Transport state is: {State}");
+            }
+
             if (BinaryProtocol)
             {
                 _socket?.SendData(data.Data);
```
sacOO7, line 139: if the socket is non-null, no need to have optional right?

src/IO.Ably.Tests.Shared/Infrastructure/FakeTransport.cs
```diff
@@ -69,10 +69,11 @@ public void Close(bool suppressClosedEvent = true)
             }
         }
 
-        public void Send(RealtimeTransportData data)
+        public Result Send(RealtimeTransportData data)
         {
```
sacOO7, line 73: Do we have a test where this generates exception while sending data and error captured & the message is queued? ( Not sure, if we already have tests but not present in the current PR )

src/IO.Ably.Shared/Transport/MsWebSocketTransport.cs
marto83: You are right. I didn't look at the existing code to update that. However I'd like to keep the optional check just in case. `_socket` is set to null on another thread so hypothetically it can still become null during the execution of this method. If you are happy, I'd like to leave it for now.

src/IO.Ably.Tests.Shared/Infrastructure/FakeTransport.cs
marto83: Yes we do. Have a look at `WithChannelInDetachingState_WhenTransportIsDisconnected_ShouldResendDetachMessageOnConnectionResumed`. This was the test that was throwing errors intermittently.

sacOO7: Great : )

src/IO.Ably.Shared/Transport/MsWebSocketTransport.cs
sacOO7: Yeah, I am happy to have it there since it's a small code snippet having an optional check. It's just that, redundant optional checks shouldn't be passed recursively if nulls are already handled ( it can also lead to redundant recursive null handling ).
Resharper/ Rider null inspector might help (Not sure if we have one :p, we have it in other IntelliJ IDE's)

sacOO7: I am marking the convo. as resolved for now : )

PR #445, opened 2020-09-28
Github workflow

marto83:
In this pull request:
* Remove Azure DevOps build
* Remove Circle CI build
* Add Windows github actions to test both the net framework and net standard versions
* Add Linux github action to test the net standard version of the library

In the process I had to tweak some test to get them passing on Linux consistently.


src/IO.Ably.Tests.Shared/Realtime/DeltaSpecs.cs
```diff
@@ -132,9 +132,11 @@ public async Task
 
         [Fact]
         [Trait("spec", "RSL6f")]
+        [Trait("linux", "skip")]
```
QuintinWillison, line 135: I'm interested as to what this does and why it's there?

src/IO.Ably.Tests/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -475,7 +475,7 @@ public async Task
 
             var channel = client.Channels.Get("test-channel");
             channel.Once(ChannelState.Detaching, change => client.GetTestTransport().Close(false));
-            channel.Attach();
+            await channel.AttachAsync();
```
QuintinWillison, line 478: Interesting. What was wrong with the blocking variant? Is that fixed elsewhere?

QuintinWillison (approved):
Looks sensible to me, thanks. I think I might add @sacOO7 as a reviewer to get a second opinion.

src/IO.Ably.Tests.Shared/Realtime/DeltaSpecs.cs
marto83: This is because the test delta files were generated on Windows with different line endings so on Linux they don't match and the delta encoding library. The label makes sure the test is not run on Linux because it will never pass.

src/IO.Ably.Tests/Realtime/ConnectionSandBoxSpecs.cs
marto83: actually the first variant was non blocking. It just sends the attach message but doesn't wait for the confirmation. For this specific test it was causing in issue too often on Linux. So I made sure I wait for the attach to complete.

.github/workflows/build-and-test-linux.yml
```diff
@@ -0,0 +1,25 @@
...
+        with: 
+          submodules: 'recursive'
+      - name: Download dotnet framework
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.x'
+      - name: Download fake-cli
+        run: dotnet tool install fake-cli --tool-path .
+      - name: Unit tests
+        run: ./build.sh Test.NetStandard.Unit.WithRetry
```
sacOO7, line 22: We have integration tests with retry right 🤔

.github/workflows/build-and-test-windows.yml
```diff
@@ -0,0 +1,29 @@
...
+        with: 
+          submodules: 'recursive'
+      - name: Download framework
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: '3.1.x'
+      - name: Download fake-cli
+        run: dotnet tool install fake-cli --tool-path .
+      - name: Unit tests
+        run: ./build.cmd ${{ matrix.target }}.Unit
```
sacOO7, line 26: Aren't we using retry here too

src/IO.Ably.Tests.Shared/Realtime/DeltaSpecs.cs
```diff
@@ -132,9 +132,11 @@ public async Task
...
         public async Task
             WhenMultipleMessagesWithAMixOfNormalAndDeltaMessages_ShouldDecodeCorrectly()
         {
+            using var logging = EnableDebugLogging();
```
sacOO7, line 139: Aren't we using code block here 🤔. Not sure if it auto-dispose without code block too 😀

src/IO.Ably.Tests/Realtime/ConnectionSandBoxSpecs.cs
sacOO7: Nicely implemented. ( I wish there was something like this in Java, I don't find futures that powerful in Java haha)

.github/workflows/build-and-test-linux.yml
marto83: we do. Usually the integration tests always retry but I found when run in linux about 3 - 4 tests always failed when run together with the rest. So just for Linux I created a new target for the unit tests. It will be good to have them passing full time but I spent a full day and still managed to get some that failed intermittently while in windows they always passed 🤔. That's why the new target

src/IO.Ably.Tests.Shared/Realtime/DeltaSpecs.cs
marto83: this is new with c#8 it automatically disposes at the end of the code block. It's pretty neat 😉

src/IO.Ably.Tests/Realtime/ConnectionSandBoxSpecs.cs
QuintinWillison: Thanks for the explanation. 😁

Issue #446, opened 2020-10-01
Websocket failures during high load


marto83:
The issue manifests itself when sending over messages at a rate of over 100 per second. The messages are of varying sizes with some over 48kb in size.

What happens is the content of the message is truncated after 49139 characters and the rest of the content of the message consist of random characters. 

![image (4)](https://user-images.githubusercontent.com/150508/94788440-74ef0d00-03cb-11eb-8285-14b60ecb2781.png)

Also the frame types change from Text or Binary to Reservedx4 and the server kills the connection.

marto83:
I've done further testing and the issue appears in .net Framework 4.7.1 up to 4.8. There is no issue with .net core. 

I've setup a small server that mimics the Ably service and hosted it in Azure and managed to replicate the issue. It's definitely not related to the ably service and it's somewhere in the System.net.websocket implementation. 

When testing the issue with the ably service we get the error quicker but it is always the same error. 
The payload gets truncated usually after 16371 bytes or 49139. 

Next steps will be to create a very simple console app that simulates the same issue and report it to Microsoft in the .net runtime repo: https://github.com/dotnet/runtime/issues/new?assignees=&labels=&template=01_bug_report.md&title=

I've asked a question on stack overflow but haven't heard anything yet.

marto83:
After further testing I found that by increasing the send and receive buffers to the maximum 64kb the issue does not appear. This is after an hour of testing and 15k 80kb messages sent. 
I will do some more testing over the coming days to come up with what size buffer is needed.

marto83:
@QuintinWillison I need a way let consumers of the library increase the default buffers for the websocket connection which fixes the issue with large message sizes. The default buffer sizes are 8192 bytes for both the sending and receiving buffers. The maximum size is 64kb which fixes the issues I've been experiencing above. 

I'm not sure how to best add it to the library. Currently the default Websocket implementation is internal to the library. So I'm thinking of 2 options:

1. Add to 2 new properties to ClientOptions called `WebsocketSendBufferSize` and `WebsocketReceiveBufferSize` and we can document what they do. 
2. I can make the Websocket implementation public document how to set parameters for it. 

Option 1 seems more straight forward, however, I don't like the fact that it's only necessary for one platform and the 8kb default buffer size is perfectly fine for most use cases. 

What do you think?

sacOO7:
Just a thought.... it there a way to provide  access to raw websocket
through ably instance and then set required options on top of it.

[...]

PR #451, opened 2020-10-15
Issue/446 websocket trouble

marto83:
I've made MsWebsocketConnection and MsWebsocketTransport public and added a new options class that can be passed to them which allows users to set the websocket send and receive limits. 

Closes: #446 

Unfortunately I found a bug in the way we handle Presence. It can't have been around for some time because a week ago all tests passed but the 250 messages presence test no longer passes. Need to check if something has changed with presence.


src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
```diff
@@ -2,60 +2,134 @@
...
+        public MsWebSocketConnection(Uri uri, MsWebSocketOptions socketOptions)
         {
-            Logger = logger;
             _uri = uri;
             ClientWebSocket = new ClientWebSocket();
+
+            if (socketOptions.SendBufferInBytes.HasValue || socketOptions.ReceiveBufferInBytes.HasValue)
+            {
+                var receiveBuffer = socketOptions.ReceiveBufferInBytes ?? (8 * 1024);
+                var sendBuffer = socketOptions.SendBufferInBytes ?? (8 * 1024);
```
sacOO7, line 97: Can we store the constant value somewhere else? maybe in defaults or just refer it from native defaults from the WebSocket lib

src/IO.Ably.Shared/Transport/MsWebSocketConnection.cs
```diff
@@ -2,60 +2,134 @@
...
+        private readonly Channel<MessageToSend> _sendChannel = Channel.CreateUnbounded<MessageToSend>(
+            new UnboundedChannelOptions()
+            {
+                SingleReader = true,
+                SingleWriter = true,
+            });
 
         internal ClientWebSocket ClientWebSocket { get; set; }
 
         private CancellationTokenSource _tokenSource = new CancellationTokenSource();
```
sacOO7, line 82: Aren't we using the cancel method on the source from the controller/tokenprovider class?
- I am not sure if it's in the code but not in the PR or if it's being called by respective token consumers.
- Or we don't need it in the first place, this is just a hunch though :p

sacOO7: Ok, we already are doing it in a dispose :+1:  Just ignore :p

sacOO7: Maybe you can look for places where we might accidentally miss it

QuintinWillison:
@sacOO7 Can you please evolve this:

1. Check that we're only exposing the minimum required to provide these override points. I'm a little concerned this might grow the public API for the client library in areas it does not need. Focus on the original issue for scope analysis.
2. Add documentation to the readme describing how to use the new interfaces / overrides, and why a customer might need to use them.

Thanks.

sacOO7:
Also, @marto83 would you like to look into public interfaces/documentation or should I take it up?

marto83:
@sacOO7 I think for this one it will be easier for me to do it. I'll add a section in readme about sending large messages at a higher rate. I've documented all the classes anyway. We can then merge it.

QuintinWillison (approved):

PR #453, opened 2020-10-15
Feature/452 transport params

marto83:
Spec align. Add custom Transport params dictionary to ClientOptions

### DIFF
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -299,7 +299,7 @@ namespace IO.Ably
         /// <summary>
         /// Additional parameters to be sent in the querystring when initiating a realtime connection.
         /// </summary>
-        public Dictionary<string, string> TransportParams { get; set; } = new Dictionary<string, string>();
+        public Dictionary<string, object> TransportParams { get; set; } = new Dictionary<string, object>();
 
         /// <summary>
         /// [Obsolete] Tells the library whether to capture the current SynchronizationContext and use it when triggering handlers and emitters
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -297,6 +297,11 @@ namespace IO.Ably
         public bool IdempotentRestPublishing { get; set; } = Defaults.ProtocolVersionNumber >= 1.2;
 
         /// <summary>
+        /// Additional parameters to be sent in the querystring when initiating a realtime connection.
+        /// </summary>
+        public Dictionary<string, string> TransportParams { get; set; } = new Dictionary<string, string>();
+
+        /// <summary>
         /// [Obsolete] Tells the library whether to capture the current SynchronizationContext and use it when triggering handlers and emitters
         /// The default has changed from `true` to `false`
         /// It will be removed in the next version of the library.
--- a/src/IO.Ably.Shared/Transport/TransportParams.cs
+++ b/src/IO.Ably.Shared/Transport/TransportParams.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
@@ -72,6 +73,11 @@ namespace IO.Ably.Transport
         /// </summary>
         public bool EchoMessages { get; private set; }
 
+        /// <summary>
+        /// Additional parameters coming from ClientOptions.
+        /// </summary>
+        public Dictionary<string, string> AdditionalParameters { get; set; }
+
         private TransportParams()
         {
         }
@@ -106,6 +112,7 @@ namespace IO.Ably.Transport
             result.UseBinaryProtocol = options.UseBinaryProtocol;
             result.RecoverValue = options.Recover;
             result.Logger = logger ?? options.Logger;
+            result.AdditionalParameters = options.TransportParams;
             return result;
         }
 
@@ -166,6 +173,11 @@ namespace IO.Ably.Transport
             if (ClientId.IsNotEmpty())
             {
                 result["clientId"] = ClientId;
+            }
+
+            if (AdditionalParameters?.Any() ?? false)
+            {
+                return AdditionalParameters.Merge(result);
             }
 
             return result;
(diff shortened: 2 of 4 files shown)


src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -296,6 +296,11 @@ public bool UseBinaryProtocol
         /// </summary>
         public bool IdempotentRestPublishing { get; set; } = Defaults.ProtocolVersionNumber >= 1.2;
 
+        /// <summary>
+        /// Additional parameters to be sent in the querystring when initiating a realtime connection.
+        /// </summary>
+        public Dictionary<string, string> TransportParams { get; set; } = new Dictionary<string, string>();
```
QuintinWillison, line 302: [RTC1f](https://docs.ably.io/client-lib-development-guide/features/#RTC1f) states that these values should simply be `Stringifiable` (spec. wording, not mine) so perhaps the values here should be `Object` and your implementation then uses [`ToString()`](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=netcore-3.1)?

QuintinWillison (changes requested):

src/IO.Ably.Shared/ClientOptions.cs
marto83: I did see that but didn't think it added much value as these will be used for query parameters. I'm happy to change it and call `ToString()` if you think it's required.

sacOO7: I agree with @QuintinWillison [RTC1f](https://docs.ably.io/client-lib-development-guide/features/#RTC1f), it states If values are unidiomatic to the language, the implementer may consider this as equivalent to String. Since dotnet is a typed language, it will be great if we can leverage that feature and be specific with values stored in the map. I am not sure if the current implementation works without being specific with values, I am considering usability wrt passed values in ClientOptions.

sacOO7: Maybe use something like ```where T : struct, string``` for value

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs
```diff
@@ -137,6 +138,48 @@ public async Task ShouldSetTransportLibVersionParameter()
...
         }
 
+        [Fact]
+        [Trait("spec", "RTC1f")]
+        public async Task WithNullTransportParamsInOptions_ShouldNotThrow()
+        {
+            var client = await GetConnectedClient(options => options.TransportParams = null);
+
+            var uri = LastCreatedTransport.Parameters.GetUri().ToString();
+            Assert.True(true);
```
sacOO7, line 148: 😵😱

src/IO.Ably.Shared/Transport/TransportParams.cs
```diff
@@ -168,6 +175,11 @@ public Uri GetUri()
                 result["clientId"] = ClientId;
             }
 
+            if (AdditionalParameters?.Any() ?? false)
```
sacOO7, line 178: Any returns a bool value, can we remove ```?? false``

sacOO7, line 178: Can we make it more simple to read
```if ( param is not null && param.any)```
Took me some time to understand why ```false``` was needed

File: src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs (hunk 2 above)
Comment (marto83) [reply src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs:148]: The test just makes sure `GetUri()` doesn't throw an exception.

Comment (sacOO7) [reply src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs:148]: Can we use ```RecordException``` to be explicit about assertions 
https://github.com/xunit/xunit/blob/master/test/test.xunit.assert/Asserts/RecordTests.cs

File: src/IO.Ably.Shared/ClientOptions.cs (hunk 1 above)
Comment (marto83) [reply src/IO.Ably.Shared/ClientOptions.cs]: @QuintinWillison I've updated the params to accept `object` for the values. 
@sacOO7 as it's not a very mainstream feature I'm going to keep it simple. :)

File: src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs (hunk 2 above)
Comment (marto83) [reply src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionParameterSpecs.cs]: Thanks Sachin. Great suggestion. I updated the code.

File: src/IO.Ably.Shared/Transport/TransportParams.cs (hunk 3 above)
Comment (marto83) [reply src/IO.Ably.Shared/Transport/TransportParams.cs:209]: I haven't upgraded to c# 9 yet. Once we switch we can use `is not null`. Currently it is not available.

Comment (sacOO7) [reply src/IO.Ably.Shared/Transport/TransportParams.cs:209]: what about ``` ( param != null ) ``` 😉

Comment (marto83) [reply src/IO.Ably.Shared/Transport/TransportParams.cs:209]: I think we fall in the lang of personal preference 😁

Comment (sacOO7) [reply src/IO.Ably.Shared/Transport/TransportParams.cs:209]: Yeah, but tbh, this syntax ( sugar ) is new to me : )

Comment (sacOO7) [reply src/IO.Ably.Shared/Transport/TransportParams.cs:209]: See, if you can change it, I am just looking at it from a readability perspective

**Review by sacOO7 — APPROVED** [review]

File: src/IO.Ably.Shared/Transport/TransportParams.cs [hunk 4]
```diff
@@ -106,7 +112,39 @@ internal static async Task<TransportParams> Create(string host, AblyAuth auth, C
             result.UseBinaryProtocol = options.UseBinaryProtocol;
             result.RecoverValue = options.Recover;
             result.Logger = logger ?? options.Logger;
+            result.AdditionalParameters = StringifyParameters(options.TransportParams);
             return result;
+
+            Dictionary<string, string> StringifyParameters(Dictionary<string, object> originalParams)
```
QuintinWillison, line 118: I'm not going to insist on it, however my preference would be to lean less on system tests to validate and instead put in place some unit tests specifically for methods like this. It's very unit testable, especially in respect of specific corner cases you had in mind (as is clearly the case looking at the implementation). Perhaps one to think more on for future.

QuintinWillison (approved):

PR #454, opened 2020-10-19
Generate environment fallbacks

### LINKED ISSUE #438: Defaults: Generate environment fallbacks
Per `RSC15g`, introduced in https://github.com/ably/docs/pull/965.

See https://github.com/ably/ably-js/pull/682 and https://github.com/ably/ably-ruby/pull/196 for reference implementations.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-106) by [Unito](https://www.unito.io/learn-more)

sacOO7:
- Refactored ClientOptions
- Added environment fallbacks method to Defaults
- Renamed IsLiveEnvironment to IsProductionEnvironment  based on spec
- Fixes #438

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -50,7 +50,7 @@ namespace IO.Ably
         public string ClientId => ConnectionClientId
             ?? CurrentToken?.ClientId
             ?? CurrentTokenParams?.ClientId
-            ?? Options.GetClientId();
+            ?? Options.ClientId;
 
         public bool TokenRenewable => TokenCreatedExternally || HasApiKey;
 
@@ -66,7 +66,7 @@ namespace IO.Ably
             CurrentTokenParams = Options.DefaultTokenParams;
             if (CurrentTokenParams != null)
             {
-                CurrentTokenParams.ClientId = Options.GetClientId(); // Ensure the correct ClientId is set in AblyAuth
+                CurrentTokenParams.ClientId = Options.ClientId; // Ensure the correct ClientId is set in AblyAuth
             }
 
             if (AuthMethod == AuthMethod.Basic)
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -50,7 +50,7 @@ namespace IO.Ably
         public string ClientId => ConnectionClientId
             ?? CurrentToken?.ClientId
             ?? CurrentTokenParams?.ClientId
-            ?? Options.ClientId;
+            ?? Options.GetClientId();
 
         public bool TokenRenewable => TokenCreatedExternally || HasApiKey;
 
@@ -66,7 +66,7 @@ namespace IO.Ably
             CurrentTokenParams = Options.DefaultTokenParams;
             if (CurrentTokenParams != null)
             {
-                CurrentTokenParams.ClientId = Options.ClientId; // Ensure the correct ClientId is set in AblyAuth
+                CurrentTokenParams.ClientId = Options.GetClientId(); // Ensure the correct ClientId is set in AblyAuth
             }
 
             if (AuthMethod == AuthMethod.Basic)
--- a/src/IO.Ably.Shared/ClientOptions.cs
+++ b/src/IO.Ably.Shared/ClientOptions.cs
@@ -195,8 +195,9 @@ namespace IO.Ably
                     else
                     {
                         Logger.Warning("Deprecated fallbackHostsUseDefault : fallbackHosts: Ably.Defaults.FALLBACK_HOSTS");
-                        return Defaults.FallbackHosts;
                     }
+
+                    return Defaults.FallbackHosts;
                 }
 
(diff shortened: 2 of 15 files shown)


sacOO7:
@marto83 I was going through the ClientOptions, I see a lot of refactoring scope as a part of fixing env. fallbacks
 1. ClientOption properties duplicated with accessor methods
- Following properties are directly accessible, although separate access methods are provided. I believe business logic can be moved inside the get accessor itself (otherwise need to remove get from the property)

marto83:
@sacOO7. I had a look at some of the methods you mentions `GetClientId` ... they are quite handy and simplify some of the code. 

How are you thinking of simplifying / removing them?

marto83:
So far the PR looks good. Where are you going to use the `GetEnvironmentFallbackHosts()` method?

sacOO7:
> GetEnvironmentFallbackHosts

I will put the logic inside ```FallbackHosts { get { } }``` itself. I don't think we need to define it anywhere else.

sacOO7:
> @sacOO7. I had a look at some of the methods you mentions `GetClientId` ... they are quite handy and simplify some of the code.
> 
> How are you thinking of simplifying / removing them?

I will create the PR itself, then it will be more clear and you can suggest me changes, if any : )

sacOO7:
> @sacOO7. I had a look at some of the methods you mentions `GetClientId` ... they are quite handy and simplify some of the code.
> 
> How are you thinking of simplifying / removing them?

@marto83 you can take a look at implementation and let me know about your thoughts : )

lmars (approved):
I'm unfamiliar with .NET, but the logic looks good to me 👍

marto83:
The logic is ok. 
By convention in this project I don't put much logic in .net properties. Properties should have simple getters and setting with a minimum amount of logic. 
Can you convert the Properties you refactored and also can you create a method for Fallback hosts and call that. This way we can continue with the 'Properties are simple' convention.

sacOO7:
- I didn't see much complex logic in ```RestHost``` and ```RealtimeHost```. Just a few if-else conditions. 
- I can refactor ```FallbackHosts```. makes sense to move out the logic. Not sure if it can be kept internal and then routed through ```get``` ( instead of keeping ```get``` private)

marto83 (changes requested):
It's not about the amount of code but the principle about keeping the properties very simple. It is much better to have the `ClientSettings` object plain so when you ask for the debug info you can get everything the client has set and not calculated properties. 
For example if the client hasn't entered fallback hosts you will get `null` but if they have you will get exactly what they have entered. If the property is calculated you will get some default hosts. Of course if you knew the logic you can figure it out. The point is to keep things obvious.
Can you please remove all logic from properties in Client

sacOO7:
@marto83 I have updated the code, you can review it now

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -124,7 +121,48 @@ public ILoggerSink LogHander
         /// For development environments only. Allows a non default host for the rest service.
         /// Default: null.
         /// </summary>
-        public string RestHost { get; set; }
+        public string RestHost
```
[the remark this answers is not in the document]
sacOO7: I partially agree with you. But, we already have a property container which can be used for debugging the passed value. Using properties without any logic ( when we can),  kind of makes it look like any other publicly available variable.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -50,7 +50,7 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
         public string ClientId => ConnectionClientId
             ?? CurrentToken?.ClientId
             ?? CurrentTokenParams?.ClientId
-            ?? Options.GetClientId();
+            ?? Options.ClientId;
```
marto83, line 53: I think you forgot to revert this change. 
Everything else looks good.

marto83 (commented):
Looks good. Thanks for reverting the changes. 
Can you just double check the Options.ClientId. I'm happy to approve once that is done.

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: I have marked GetClientId() as obsolete. That's why I used ClientId here

marto83: GetClientId() is internal. No need to mark it as obsolete. Regardless could you turn `ClientId` back to a simple property and either user `GetClientId()` or even better remove it from ClientOptions and create a private helper methed inside AblyAuth that implements the logic :)

sacOO7: Okay

marto83 (approved):

kavalerov:
@QuintinWillison / @owenpearson  why is the linked issue close, but this PR is open - did we actually merge/implement this?

QuintinWillison:
The linked issue [has been re-opened](https://github.com/ably/ably-dotnet/issues/438#issuecomment-820315945).

PR #456, opened 2020-10-27
Fix client time drift

### LINKED ISSUE #441: Handle case where client time has drifted further than 500ms after we have saved the server time offset
It's a continuation of spec item ([RSA4b1](https://docs.ably.io/client-lib-development-guide/features/#RSA4b1)). Once we have saved the server time we should check if the client time hasn't drifted further. In this case we should call and get Server time again

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-567) by [Unito](https://www.unito.io)

sacOO7:
- Refactored ServerNow implementation
- References #441

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -21,15 +21,20 @@ namespace IO.Ably
             _rest = rest;
             Logger = options.Logger;
             ServerTime = () => _rest.TimeAsync();
-            ServerNow = () => null;
             Initialise();
         }
 
         protected Func<Task<DateTimeOffset>> ServerTime { get; set; }
 
-        private bool HasServerTime => ServerNow().HasValue;
+        private TimeSpan? _serverTimeDiff;
 
-        protected Func<DateTimeOffset?> ServerNow { get; private set; }
+        private bool HasServerTime => ServerNow.HasValue;
+
+        protected DateTimeOffset? ServerNow
+        {
+            get => _serverTimeDiff != null ? Now() - _serverTimeDiff : null;
+            set => _serverTimeDiff = Now() - value;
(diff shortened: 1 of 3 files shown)


sacOO7:
Actually, I am a bit skeptical about the second option ( starting a custom stopwatch counter to measure server time). I came up with the following snippet to calculate ServerTime.

```public class ServerTime
{
    public static DateTime Now
    {
        get
        {
            if (baseTime != null)
                return baseTime.Add(swatch.Elapsed);
            else
                return DateTime.Now;
        }
        set
        {
            baseTime = value;
            swatch.Reset();
            swatch.Start();
        }
    }

    private static DateTime baseTime;
    private static Stopwatch swatch = new Stopwatch();

}
```
So, server time is now independent of client time, I would like to know your thoughts before I proceed on this one @marto83

marto83:
@sacOO7 I thought about this and there isn't an easy way to achieve this and I don't know if using the stopwatch will fix this. I wonder if we should just get server time every time we request a token. It will definitely make the logic much easier.

sacOO7:
Well, @marto83 I think current logic should work just fine ( considering the client does not mess up with the clock).
We are thinking about too much of a critical scenario, which might never happen. 
I believe we don't need to introduce new code, for now 😌

sacOO7:
And if it gives a valid token for an expired token, the server will anyway return an invalid token error. So, I don't think we need to introduce complex logic.

marto83:
I agree. Let's park this.

sacOO7:
Btw, as a part of PR, I just did some refactoring ( Did not implement actual solution). I ran relevant tests, everything seems to be working. I will convert it to PR, can you check

marto83 (approved):

PR #459, opened 2020-11-15
Fail connection on forbidden authz

### LINKED ISSUE #448: Fail connection immediately if authorize() called and 403 returned
The first task here is to validate whether this client library behaves per `RSA4d` and `RSA4d1`, as added in https://github.com/ably/docs/pull/961 with reference implementation in https://github.com/ably/ably-js/pull/680.

If not, then implement correct behaviour with appropriate testing.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-126) by [Unito](https://www.unito.io)

sacOO7:
- Fixed flaky test behavior 
- Fixes #448


sacOO7:
@marto83 I have made some cross-domain changes ( Had no other choice). Please go through the code, and let me know if you feel, there is a better way to do it ( To publish ConnectionFailed event).

marto83 (commented):
@sacOO7. I had a look through the code and also the ably-js repository. 

You are right in putting the logic in the Authorize method and you are right to queue a command to be processed. I think we should make a few changes to keep the code clean. 

1.Let's not pass the whole connectioncontext interface to AblyAuth. Instead let's create a property that holds an Action<RealtimeCommand> that has an empty default behaviour. Like the code below
`        internal Action<RealtimeCommand> ExecuteCommand {get; set;} = (cmd) => { }; `  
This creates the property and initializes it to an empty action. So you don't need to do a null check and can always trust it will be `noop` if somebody is only using AblyRest.

2. Create a new command called something like `HandleAblyAuthoriseError`
3. The code in ablyAuth will change to

```
            try
            {
                CurrentToken = await RequestTokenAsync(tokenParams, authOptions);
            }
            catch (AblyException ex)
            {
                ExecuteCommand(HandleAblyAuthorizeError.Create(ex.ErrorInfo));
                throw;
            }
```

4. The place to set this ExecuteCommand Property will be when AblyRealtime is created after line 61
`RestClient.AblyAuth.ExecuteCommand = cmd => Workflow.QueueCommand(cmd);`

5. Finally you need to implement the logic inside RealtimeWorkflow. 

This way in the logs you will clearly be able to see when this specific case is handled.

sacOO7:
Hi @marto83 , I have made the necessary changes. Regarding the test, it works fine in debug mode but gives a problem when run directly as a part of the test suite. I tried debugging the exact issue but I was not able to get close to the issue. I think the problem might be related to RealtimeWorkflow. Your guidance will be really helpful.

sacOO7:
It seems, some issue with ```Volatile.Read(ref ConnectionStateChanged);``` in ```Connection.cs: L279```

marto83 (approved):
I've added an extra check to make sure we don't fail the connection when the error doesn't have status 403 (Forbidden).

The rest looks good.

PR #461, opened 2020-11-20
Console app documentation

### LINKED ISSUE #455: SDK has no documented way to block on initialisation.
In a simple console app using the example code, the library fails to connect or send anything before exiting. This includes checking the API key is correct.
To make the application work I added some sleeps to the main thread and it published correctly.
I believe this is because the SDK was not able to set up and send before exiting due to the SDK being asynchronous by default, and being no documented way to await completion of setup or connection. If there is a way to do this I was not able to find it.

Tested with visual studios 17 and the latest 3 versions of the SDK from Nuget.

sacOO7:
- Added console app documentation
- Fixes #455

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,51 @@ options.TransportFactory = new MsWebSocketTransport.TransportFactory(websocketOp
 var realtime = new AblyRealtime(options);
 ```
 
+### Examples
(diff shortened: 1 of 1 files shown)


marto83 (approved):
Looks good to me.

PR #492, opened 2020-12-30
Update github workflow

sacOO7:
- Removed old workflow files
- Created separate workflow for unit and integration tests
- Added cross-platform OS matrix with .net standard runtime


.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  Run unit tests:
+    runs-on: ${{ matrix.os }}
+    env:
+      DOTNET_NOLOGO: true
+
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        target: [ "Test.NetStandard", "Test.NetFramework" ]
```
marto83, line 14: You can't run NetFramework tests on MacOs and ubuntu.

marto83 (commented):
Sachin, what is your thinking behind splitting the tests to run separate unit and integration tests.

tomkirbygreen:
Spoke to Martin about this and we're going to close this for now as we're in the process of adding Mac OS to the build scripts and GitHub Actions as part of the Push Notifications work.

Issue #499, opened 2021-04-09
connectionKey property missing from Message class


jasonmsmith:
See [https://docs.ably.io/client-lib-development-guide/features/#TM2h](https://docs.ably.io/client-lib-development-guide/features/#TM2h)

`Message` class should have a `connectionKey` property which is used by a publishing client when publishing on behalf of another client. Currently, I receive error 40006 when attempting to publish on behalf of another client using only the `connectionId`.

See [Message.cs](https://github.com/ably/ably-dotnet/blob/main/src/IO.Ably.Shared/Types/Message.cs) and [IMessage](https://github.com/ably/ably-dotnet/blob/main/src/IO.Ably.Shared/Types/IEncodedMessage.cs#L16). The property is missing.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-690) by [Unito](https://www.unito.io/learn-more)

PR #500, opened 2021-04-14
Add ConnectionKey property to messages

jasonmsmith:
Proposed fix for Issue #499. Added ConnectionKey property to Message class, PresenceMessage class, and IMessage interface. With these modifications, I have successfully published Ably messages via REST using the connectionKey of an active real-time client, which is the desired behavior.


marto83 (approved):
Looks good. Thank you @jasonmsmith

PR #521, opened 2021-06-14
Remove legacy tests

tomkirbygreen:
Remove the legacy tests which are now superceded by
`IO.Ably.Tests.Shared`.  Martin confirms that leaving these
behind somehow slipped through the review process at the
time.

The project `GUID` is not refered to by any other `.sln`.


owenpearson (approved):
LGTM, thanks

PR #535, opened 2021-06-21
Use 'static' where possible

tomkirbygreen:
`static` methods get us one tiny step closer to pseudo pure
functions in C#.  Like `readonly` and `records` its one of those
things that in the absence of strong `const` we should leverage
as much as possible.


marto83 (approved):
Looks good.

PR #537, opened 2021-06-21
Feature/push/push admin

marto83:
Created PushAdmin class and Api interfaces. I've added an internal property to the rest and realtime clients and documentation where required. The implementation still lacks logging and error handling.

### DIFF
--- a/src/IO.Ably.Shared/IRealtimeClient.cs
+++ b/src/IO.Ably.Shared/IRealtimeClient.cs
@@ -29,10 +29,10 @@ namespace IO.Ably
         /// <summary>A reference to the connection object for this library instance.</summary>
         Connection Connection { get; }
 
-        /// <summary>
-        /// Methods for managing Push notifications subscriptions and device registrations.
-        /// </summary>
-        PushRest Push { get; }
+        // /// <summary>
+        // /// Methods for managing Push notifications subscriptions and device registrations.
+        // /// </summary>
+        // PushRest Push { get; }
 
         /// <summary>
         /// Current client id.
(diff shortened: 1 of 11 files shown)


src/IO.Ably.Shared/IRealtimeClient.cs
```diff
@@ -28,6 +29,11 @@ public interface IRealtimeClient : IStatsCommands
         /// <summary>A reference to the connection object for this library instance.</summary>
         Connection Connection { get; }
 
+        // /// <summary>
```
tomkirbygreen, line 32: Can we not just remove this or do you plan to restore it later?

src/IO.Ably.Shared/IRestClient.cs
```diff
@@ -12,6 +13,11 @@ public interface IRestClient : IStatsCommands
         /// <summary>Authentication methods.</summary>
         IAblyAuth Auth { get; }
 
+        // /// <summary>
```
tomkirbygreen, line 16: Can we not just remove this or do you plan to restore it later?

tomkirbygreen (approved):
Looks good to me. Just a point about commented-out code.

src/IO.Ably.Shared/IRealtimeClient.cs
marto83: I added them first but you can't make them internal and I wanted to keep the public facing API the same for now so we can do another release in case we need to.

src/IO.Ably.Shared/IRestClient.cs
marto83: Just to keep the surface of the external API the same

PR #542, opened 2021-06-22
Activation State Machine - NotActivated State and Transition

marto83:
I've ported the logic for NotActivated state but have left stubs for `GetRegistrationToken` and `HandleEvent` which will come in the next 2 PRs.


tomkirbygreen (commented):
Just re-starting CI to see if I can get it to go green before Approving.

PR #543, opened 2021-06-22
Port the code for HandleEvent from the java sdk

marto83:


src/IO.Ably.Shared/Push/ActivationStateMachine.cs
```diff
@@ -61,9 +68,73 @@ private async Task ValidateRegistration()
...
+            finally
+            {
+                _handlingEvent = false;
+            }
+        }
+
+        private void PersistState()
         {
-            throw new NotImplementedException();
+            if (CurrentState != null && CurrentState.Persist)
```
tomkirbygreen, line 127: Consider: Is it worth having something like:

```
var sharedName = PersistKeys.StateMachine.SharedName;
```

At the top level of the function so we that can simplify the three use cases and avoid a bit if reference-chasing?

tomkirbygreen (commented):
Just a minor point, otherwise good.

src/IO.Ably.Shared/Push/ActivationStateMachine.cs
marto83: I'll get it updated in the next PR if you don't mind :)

src/IO.Ably.Shared/Push/ActivationStateMachine.cs
```diff
@@ -61,9 +68,73 @@ private async Task ValidateRegistration()
             }
         }
 
-        private async Task HandleEvent(Event @event)
+        public async Task HandleEvent(Event @event)
+        {
+            lock (_handleEventsLock)
```
sacOO7, line 73: @marto83 I thought we were not using any locks in the code

marto83: fair point. I'm currently just porting the code from Java to get it working and when I get both Android and iOS working, I'll refactor it to use our workflow process. 
I'm currently porting and learning how it works.

PR #545, opened 2021-06-23
Make 'MessageToSend' struct 'readonly'

tomkirbygreen:
Making the `MessageToSend` struct `readonly`.  Less is more.


marto83 (approved):
Nice. I didn't even know you could create `readonly` structs

PR #547, opened 2021-06-23
Add Transition logic for WaitingForPushDeviceDetails State

marto83:


src/IO.Ably.Shared/Push/PushAdmin.cs
```diff
@@ -30,6 +30,27 @@ internal PushAdmin(AblyRest restClient, ILogger logger)
...
+        {
+            // TODO: Add validation.
+            // TODO: Add fullwait parameter.
+            var request = _restClient.CreateRequest("/push/deviceRegistrations/", HttpMethod.Post);
+            request.PostData = details;
+            var result = await _restClient.ExecuteRequest<LocalDevice>(request);
+
+            if (result == null)
+            {
+                throw new AblyException("Failed to register device", 40000);
```
tomkirbygreen, line 48: Magic number 40,000. Can we have a symbolic?

src/IO.Ably.Shared/Push/ActivationStateMachine.States.cs
```diff
@@ -80,6 +81,76 @@ public WaitingForPushDeviceDetails(ActivationStateMachine machine)
...
+                        var ably = Machine._restClient; // TODO: Check if there is an instance when Ably is not set. In which case Java set queues GettingDeviceRegistrationFailed
+
+                        try
+                        {
+                            var registeredDevice = await ably.Push.Admin.RegisterDevice(device);
+                            var deviceIdentityToken = registeredDevice.DeviceIdentityToken;
+                            if (deviceIdentityToken.IsEmpty())
+                            {
+                                // TODO: Log
+                                _ = Machine.HandleEvent(new GettingDeviceRegistrationFailed(new ErrorInfo("Invalid deviceIdentityToken in response", 40000, HttpStatusCode.BadRequest)));
```
tomkirbygreen, line 109: Magic number 40,000. Consider using a symbolic?

tomkirbygreen (approved):
Just wondering about the `40000` magic number ...

src/IO.Ably.Shared/Push/PushAdmin.cs
paddybyers: If the HTTP request failed, then we need this error response to contain information about that error, not just a fixed, non-specific error code

src/IO.Ably.Shared/Push/ActivationStateMachine.States.cs
paddybyers: So if I'm reading this correctly, we made a successful device registration request, but the response didn't include a `deviceIdentityToken`. I think that's an internal error - ie something that we think would only happen as a result of something being broken. In that case, it's a 5xx error, not 4xx. And I think there should be a specific error code for it.

src/IO.Ably.Shared/Push/PushAdmin.cs
marto83: You both are right. I'm only hooking things up to get to a point where I can complete a happy path integration. Thus the numerous TODOs and fixed error codes. I'm going to add a TODO label here as well to make sure it doesn't get dropped.

src/IO.Ably.Shared/Push/ActivationStateMachine.States.cs
marto83: Good point Paddy. I'm porting the code from the java library. Do you want me to create issues and create PRs there as well?

PR #551, opened 2021-06-25
Waiting for Registration Sync State - Add Transition logic

marto83:


src/IO.Ably.Shared/Push/ActivationStateMachine.cs
```diff
@@ -305,5 +305,10 @@ private void ResetDevice()
             _mobileDevice.ClearPreferences(PersistKeys.Device.SharedName);
             LocalDevice = new LocalDevice();
         }
+
+        private void CallSyncRegistrationFailedCallback(ErrorInfo reason)
+        {
+            SendErrorIntent("PUSH_UPDATE_FAILED", reason);
```
tomkirbygreen, line 311: Do we need to define this string somewhere else, say for testing?

tomkirbygreen (approved):
Just a thought about string literals.

PR #576, opened 2021-06-29
Use string interpolation

tomkirbygreen:


sacOO7:
Why closed without review

PR #578, opened 2021-06-29
Fix class is never instantiated.

tomkirbygreen:


src/IO.Ably.Tests.Shared/Helpers/JAssert.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace IO.Ably.Tests.Shared.Helpers
 {
-    internal class JAssert
+    internal static class JAssert
```
sacOO7, line 6: Since, we have merged `FluentAssertions` upgrade PR, I think we can remove this `JAssert` utils class, since it seems we can do deepEquals from new version of `FluentAssertions`

sacOO7: Seems we need to install one more dependency `https://github.com/fluentassertions/fluentassertions.json` to make it work, along with `newtonSoft` upgrade to 10 +, for now we can keep it as is !!

marto83 (approved):

PR #580, opened 2021-06-29
Specify string culture explicitly

tomkirbygreen:


marto83:
I know you are not a fan of extension methods. In this case it will be easier to get it right. Something like `ToStringWithInvariantCulture()` :)

tomkirbygreen:
> I know you are not a fan of extension methods. In this case it will be easier to get it right. Something like `ToStringWithInvariantCulture()` :)

I don't think we can do this with a one-stop extension method for `this object` since `Object.ToString` does not take any parameters and therefore we can't specify 'CultureInfo.InvariantCulture'.

marto83 (approved):

marto83:
Fair point :)

PR #582, opened 2021-06-30
Fix missing 'readonly'-ness

tomkirbygreen:
Fixing missing `readonly`ness and associated secondary issues.


marto83 (approved):
Looks good.

PR #583, opened 2021-06-30
Fix 'MockBackGroundService' class not used.

tomkirbygreen:


marto83 (approved):
Oh. I see what you have done. I got confused by the title. All good.

PR #586, opened 2021-07-01
Delete project reference to deleted 'nuspec' file.

tomkirbygreen:
We removed the `io.ably_msgpack.nuspec`, but also need to remove
the various project references to it.


sacOO7:
About this one, I would question, if we need to add msgpack in the near future, is it related to that one? Although we are not using msgpack right now, it was high priority issue sometime back

marto83:
Hi @sacOO7. The Msgpack nuspec file was very out of date. When we add it back we should just create a new one.

PR #587, opened 2021-07-01
Remove unused fakeconnectioncontext properties

tomkirbygreen:


sacOO7:
About this one, I would question, if we need to add msgpack in the near future, or is it related to that one? Although we are not using message pack right now, it was high priority issue sometime back

sacOO7:
About this one, I would question, if we need to add msgpack in the near future, or is it related to that one? Although we are not using message pack right now, it was high priority issue sometime back

tomkirbygreen:
Need to refactor.

PR #589, opened 2021-07-01
Address possible 'NullReferenceException's.

tomkirbygreen:


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -1222,6 +1224,7 @@ public async Task WhenChannelBecomesFailedOrDetached_QueuedPresenceMessagesShoul
                     await client.WaitForState();
 
                     var channel = client.Channels.Get("RTP5a".AddRandomSuffix()) as RealtimeChannel;
+                    Assert.NotNull(channel);
```
marto83, line 1227: For consistency, could you use the `Should()` extensions, for example, `channel.Should().NotBeNull()`

marto83 (approved):

PR #591, opened 2021-07-02
Release/1.2.5

marto83:
This release focuses on adding:

- Heartbeat support for detecting dropped connections
- Numerous code quality fixes

It comes relatively quickly after `1.2.4` since we want to get the above out there. `1.2.5` paves the way for eventually adding Push Notifications support, which will have more structural impact.


QuintinWillison (changes requested):
I recognise that the contents of the change log has been auto-generated, but from a customer's perspective this list could be more than a little bit daunting.

Can we please thin it down, removing entries which don't have an obvious customer impact (e.g. internal refactors, test fixes, etc..).

There's always the "Full Changelog" link that a customer can click if they really want to see every change made.

QuintinWillison (approved):

PR #602, opened 2021-07-10
Migrate to 'Fluent Assertions'

tomkirbygreen:
Migrate the our `Assert.` based assertions over to `Fluent
Assertions`.  This way we get better assertion messages and
move towards having a single API for assertion testing.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ChannelsSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ChannelsSpecs.cs
@@ -227,7 +227,7 @@ namespace IO.Ably.Tests.Realtime
 
             // Assert
             Assert.Same(channel, enumerator.Current);
-            Assert.False(enumerator.MoveNext());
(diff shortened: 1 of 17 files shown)


src/IO.Ably.Tests.Shared/Realtime/ChannelsSpecs.cs
```diff
@@ -348,7 +348,7 @@ public async Task PublishShouldNotAlterChannelOptions()
             await client.ProcessCommands();
 
             Assert.Equal(options.ToJson(), channel2.Options.ToJson());
-            Assert.True(options.CipherParams.Equals(cipherParams));
+            options.CipherParams.Equals(cipherParams).Should().BeTrue();
```
marto83, line 351: this could be simplified to `options.CipherParams.Should().Be(cipherParams)`

marto83 (approved):
looks good. Just one more improvement you can make :)

PR #605, opened 2021-07-12
Refactor 'PresenceMap'

tomkirbygreen:
Removed unnecessary virtuals, properties and layout type along .NET BCL FDG lines.


marto83 (approved):
Looks good.

PR #607, opened 2021-07-14
Ably agent header

tomkirbygreen:


src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
```diff
@@ -80,6 +80,33 @@ public async Task WhenCallingUrl_AddsRequestIdIfSetTrue()
...
+            {
+                "ably-dotnet/",
+                "os-platform/",
+                "runtime/",
+            };
+
+            agentValues.Length.Should().Be(keys.Length);
+            for (int i = 0; i < keys.Length; ++i)
+            {
+                agentValues[i].StartsWith(keys[i]).Should().BeTrue();
```
marto83, line 106: You can do `agentValues[i].Should().StartWith(keys[i])` then you get better error messages if it fails. :)

src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
```diff
@@ -80,6 +80,33 @@ public async Task WhenCallingUrl_AddsRequestIdIfSetTrue()
...
+        [Trait("spec", "RSC7d")]
+        public async Task WhenCallingUrl_AddsDefaultAblyAgentHeader()
+        {
+            var response = new HttpResponseMessage(HttpStatusCode.Accepted) { Content = new StringContent("Success") };
+            var handler = new FakeHttpMessageHandler(response);
+            var client = new AblyHttpClient(new AblyHttpOptions(), handler);
+
+            await client.Execute(new AblyRequest("/test", HttpMethod.Get));
+            string[] values = handler.LastRequest.Headers.GetValues("Ably-Agent").ToArray();
+            values.Length.Should().Be(1);
```
marto83, line 93: you can change this to `values.Should().HaveCount(1)`

src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
sacOO7: Simplest assertion would be to add string `contains` some value

PR #613, opened 2021-07-14
Push/push admin publish method

marto83:
Add validation to `PushAdmin.PublishAsync` method and add 2 unit tests to make sure we are using the correct url and payload.


tomkirbygreen (approved):
There are some GitHub Actions warnings about `(cast)` being followed by whitespace, i.e. `(cast) thing` rather than `(cast)thing`.

PR #633, opened 2021-07-16
Add 'SemanticVersion' type

tomkirbygreen:
We need a `SemanticVersion` type that follows the https://semver.org/ standard.


marto83 (approved):
Looks good. What are we going to use this class for?

PR #638, opened 2021-07-17
Remove unused 'RealtimeClients' property

tomkirbygreen:
I think the 'RealtimeClients' property may be unused. It's appears to always be an empty collection and I can't find any unit test that exercises it.  Check with Martin...


marto83:
You are right. It's never used. I used to add all realtime clients created during a test and dispose them at the end of each test. 
image

That is how it was and I removed the code on 15 Nov 2019. Now we have a passing test suite I think we should introduce that back so we clean up connection properly and see if it breaks anything.

PR #656, opened 2021-07-18
Push/notactivated handle gotpushdevicedetails

marto83:


src/IO.Ably.Shared/Push/RegistrationToken.cs
```diff
@@ -1,22 +1,37 @@
...
+        /// <summary>
+        /// Token value.
+        /// </summary>
         public string Token { get; set; }
 
+        /// <summary>
+        /// Constructs a new registration token instance.
+        /// </summary>
+        /// <param name="type">Token type.</param>
+        /// <param name="token">Token value.</param>
```
tomkirbygreen, line 24: We could always call these `tokenType` and `tokenValue` - but we can address that later I guess :-)

tomkirbygreen (approved):

PR #669, opened 2021-07-19
Push/waitingforregistrationsync handle registrationsynced

marto83:


tomkirbygreen (approved):

src/IO.Ably.Tests.Shared/Push/FakeMobileDevice.cs
```diff
@@ -22,7 +23,12 @@ public void SetPreference(string key, string value, string groupName)
 
         public string GetPreference(string key, string groupName)
         {
-            return Settings[$"{groupName}:{key}"];
+            if (Settings.ContainsKey($"{groupName}:{key}"))
+            {
+                return Settings[$"{groupName}:{key}"];
```
tomkirbygreen, line 28: Maybe factor this out into something like:

```
string compoundKey = $"{groupName}:{key}";
```

and then reuse it?

tomkirbygreen (approved):

PR #675, opened 2021-07-24
.NET types do not using screaming caps

tomkirbygreen:
Following the BCL FDG's.


marto83:
I've already changed that in an upcoming PR. Do you mind if we close it so it doesn't create merge conflicts

PR #680, opened 2021-07-25
'RealtimeWorkflow' remove redundant 'ToString'

tomkirbygreen:


marto83 (approved):

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -166,7 +166,7 @@ private async Task Consume()
                                     }
                                     catch (Exception e)
                                     {
-                                        Logger.Error("Error Processing command: " + cmdsToExecute.ToString(), e);
+                                        Logger.Error("Error Processing command: " + cmdsToExecute, e);
```
marto83, line 169: while you are changing it, should we not move to using string interpolation and change it to `$"Error Processing command: {cmdsToExecute}"`

tomkirbygreen: Yeah I can do that. Generally speaking I try to do one kind of change per PR but seeing as we touched the line anyway :-)

PR #686, opened 2021-07-26
Push/handle event

marto83:


src/IO.Ably.Shared/Push/ActivationStateMachine.cs
```diff
@@ -29,6 +48,118 @@ internal ActivationStateMachine(AblyRest restClient, IMobileDevice mobileDevice,
...
 
+        public async Task HandleEvent(Event @event)
+        {
+            // Handle current event and any consequent events
+            // if current event didn't change state put it in pending queue and return
+            // finally check the pending event queue and process it following the above rules
+            // finally finally - preserve the state
+            Debug($"Handling event ({@event.GetType().Name}. CurrentState: {CurrentState.GetType().Name}");
+
+            var canEnter = await _handleEventsLock.WaitAsync(2000); // Arbitrary number = 2 second
```
tomkirbygreen, line 59: I'm slightly concerned about the phrase `'Arbitrary number'` in source code that customers can see. Can we find some different words perhaps? Where does the magic 2 seconds come from?

tomkirbygreen (changes requested):
My bad: clicked merge on the screaming-caps PR before I saw you comment. I'll be more careful in future. Would you be able to integrate the changes just this once?

I just have one comment about this PR.

src/IO.Ably.Shared/Push/ActivationStateMachine.cs
marto83: I didn't want to make the wait indefinite but also didn't know how short it should be. 2 seconds worked ok in the initial integration testing but I haven't got to that part yet in the PR process. That's why I've put that for now it's arbitrary. Once we have done more real world testing and checked logs it will either prove that 2s works or we will have to tweak it.

marto83:
@tomkirbygreen I've resolved the conflict :)

tomkirbygreen (approved):

PR #694, opened 2021-08-01
Remove interface methods that hide those in the base

tomkirbygreen:
Methods in the `IMessage` interface hide those in the `IPayload` base interface.

Perhaps due to a previously interupted refactoring?


marto83 (approved):
Good spot :)

PR #715, opened 2021-08-05
Fix param name in 'Presence' ctor.

tomkirbygreen:


marto83 (approved):
some of these spelling mistakes are entertaining :)

PR #724, opened 2021-08-06
'CipherEncoder' specify string comparison explicitly

tomkirbygreen:


src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
```diff
@@ -26,7 +26,7 @@ public override Result<ProcessedPayload> Decode(IPayload payload, DecodingContex
             }
 
             var cipherType = GetCipherType(currentEncoding);
-            if (cipherType.ToLower() != options.CipherParams.CipherType.ToLower())
```
marto83, line 29: There is a `string` extension method that does just that. It's called `EqualsTo` and by default it ignores case

marto83 (approved):

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
sacOO7: +1

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
```diff
@@ -26,7 +26,7 @@ public override Result<ProcessedPayload> Decode(IPayload payload, DecodingContex
...
+            if (cipherType.EqualsTo(options.CipherParams.CipherType) == false)
```
sacOO7, line 29: Can we use commonly used syntax 
`if (!cipherType.EqualsTo(options.CipherParams.CipherType))`

marto83: I disagree but if you both twist my hand I will start using `!` instead of `==false`. My eyes tend to skip over `!` where there is no missing `== false` even if you wanted to 😂

marto83: my other suggestion is to move to C# 9 and use the `not` keyword

sacOO7: I have never used it till now, and haven't came across any major codebases that follows this practice. If you ask majority of developers, they will choose to go with `!` . Also, it destroys purpose of boolean evaluation by `if` conditional. Also, checking for `false` value is not optimised solution IMO

sacOO7: @marto83 I did open rider sometime back, if `not` is supported : p. It's supported in python though ... lucky

sacOO7: That said, it will be easy to replace `!` with `not` in the future 😃

marto83: > Also, checking for false value is not optimised solution IMO
Sachin you will have to show me some data for me to believe you on this one. 

It's just a personal preference and I have avoided many boolean logic bugs this way.

sacOO7: There are two logical comparisons, first is checking if it's false, then again checking if resulting value is true. I am not sure, if there are compiler optimisations for this, but from first look, it feels overcomplicated. 
Not that the optimisation matters much in the case, I am looking from simplicity and from maintenance perspective. While I value there are few contributors as of now, there will be several maintainers in the future, who would love to stick to standard way of doing things.

sacOO7: I would say, it's better start treating `!` as `not` while reading the code inside `if` conditional. It will help reading other codebases too. Well, that problem will eventually be solved when `not` is introduced in dotnet.

sacOO7: I am approving this PR. I would like @tomkirbygreen to take a call on this, since we are following some project specific defaults (can be changed if needed).

sacOO7 (approved):

src/IO.Ably.Shared/MessageEncoders/CipherEncoder.cs
sacOO7: Marking the discussion as resolved

PR #726, opened 2021-08-06
Remove unused 'GettingPushDeviceDetailsFailed' class

tomkirbygreen:
Martin to check...

### DIFF
--- a/src/IO.Ably.Shared/Push/ActivationStateMachine.Events.cs
+++ b/src/IO.Ably.Shared/Push/ActivationStateMachine.Events.cs
@@ -32,14 +32,6 @@
             }
         }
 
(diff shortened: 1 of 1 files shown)


marto83 (commented):
I'll just double check this one on the weekend. I'm not sure why it is not used and whether it should be used anywhere.

tomkirbygreen:
> I'll just double check this one on the weekend. I'm not sure why it is not used and whether it should be used anywhere.

Did you get a chance to look into this Martin? Just want to know if we can commit for now - we can always restore the type if you subsequently think we're missing something.

marto83:
Go for it. I haven't look but as you say we can always get it back.

marto83 (approved):

PR #735, opened 2021-08-09
'HttpUtility' Fix Possible 'NullReferenceException' warning

tomkirbygreen:
Re-express `HttpUtility` to remove the possible `NullReferenceException` warning.

### DIFF
--- a/src/IO.Ably.Shared/HttpUtility.cs
+++ b/src/IO.Ably.Shared/HttpUtility.cs
@@ -116,10 +116,15 @@ namespace IO.Ably
             }
         }
 
-        private void FillFromString(string query, bool urlencoded)
+        private void FillFromString(string query, bool urlEncoded)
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Shared/HttpUtility.cs
```diff
@@ -116,10 +116,15 @@ public HttpValueCollection(string query, bool urlencoded)
             }
         }
 
-        private void FillFromString(string query, bool urlencoded)
+        private void FillFromString(string query, bool urlEncoded)
         {
+            if (query == null)
```
marto83, line 121: let's go with `query is null` way of checking for `null`

marto83 (changes requested):
Could you change the null check to `query is null` instead?

marto83 (approved):

PR #749, opened 2021-08-16
'ActivationStateMachine' make member private

tomkirbygreen:
Check with Martin as the implementation is actually throwing a `NotImplementedException`.  Is this by design or does this indicate we've got unfinished business?


marto83:
That part is not finished. This is todo with the actual device integration which is not done yet.

marto83:
Let's close this PR and come back to it when the method is implemented

PR #763, opened 2021-08-19
Pass AuthCallback exception to ErrorInfo to make it easier for debugging

### LINKED ISSUE #762: Unity on iOS: Error 80019
I've put a ticket in with support for this error that we're seeing in Unity on device only - the same code works in the editor. In trying to track it down, I notice that the inner exception ('ex') isn't put into the ErrorInfo constructor so it isn't available via the AblyRealitime.Connection.ErrorReason object which it makes it difficult to see it, as instructed! Is this a mistake?

https://github.com/ably/ably-dotnet/blob/b9ec2dc4e093df06cd54beb65f9cddeb6f5ace7f/src/IO.Ably.Shared/AblyAuth.cs#L291-L303

These exceptions do put it into the ErrorInfo constructor:

https://github.com/ably/ably-dotnet/blob/b9ec2dc4e093df06cd54beb65f9cddeb6f5ace7f/src/IO.Ably.Shared/AblyAuth.cs#L338-L344
https://github.com/ably/ably-dotnet/blob/b9ec2dc4e093df06cd54beb65f9cddeb6f5ace7f/src/IO.Ably.Shared/AblyAuth.cs#L357-L363

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1147) by [Unito](https://www.unito.io)

marto83:
Fixes #762

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -267,7 +267,7 @@ namespace IO.Ably
 
                     if (callbackResult == null)
                     {
-                        throw new AblyException("AuthCallback returned null", 80019);
+                        throw new AblyException("AuthCallback returned null", ErrorCodes.ClientAuthProviderRequestFailed);
                     }
 
                     if (callbackResult is TokenDetails)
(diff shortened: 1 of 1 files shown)


src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -298,9 +298,11 @@ public virtual async Task<TokenDetails> RequestTokenAsync(TokenParams tokenParam
 
                     throw new AblyException(
                         new ErrorInfo(
-                        "Error calling AuthCallback, token request failed. See inner exception for details.",
-                        80019,
-                        statusCode), ex);
+                            "Error calling AuthCallback, token request failed. See inner exception for details.",
+                            80019,
```
tomkirbygreen, line 302: Isn't this magic number actually `ErrorCodes.ClientAuthProviderRequestFailed`?

tomkirbygreen (changes requested):
Just a quick thought...

src/IO.Ably.Shared/AblyAuth.cs
marto83: Fair comment. I've update the magic numbers in AblyAuth with the const

tomkirbygreen (approved):

PR #777, opened 2021-08-22
'JsonMessageSerializerTests' remove redundant type argument specifica…

tomkirbygreen:
'JsonMessageSerializerTests' remove redundant type argument specification.

Note the newly added `Debug.Assert(x.y.HasValue, $"'{nameof(x.y)}' should have a value");` is needed to pacify a further inline code analysis warning, it is not possible to use a `FluentAssertions` `Should()` based approach for this.


marto83 (commented):
Could we convert the `Assert.Equal` to `.Should()...`

marto83 (approved):

PR #802, opened 2021-08-30
'AblyCollectionFixture' fix possible 'NullReferenceException'

tomkirbygreen:


src/IO.Ably.Tests.Shared/Infrastructure/AblyCollectionFixture.cs
```diff
@@ -60,22 +60,21 @@ public TestEnvironmentSettings()
...
         public ClientOptions CreateDefaultOptions(string key = null, string environment = null)
         {
-            environment = environment ?? Environment;
+            environment ??= Environment;
 
-            var env = System.Environment.GetEnvironmentVariable("ABLY_ENV").IsNotEmpty()
-                ? System.Environment.GetEnvironmentVariable("ABLY_ENV").Trim()
-                : environment;
+            var ablyEnv = System.Environment.GetEnvironmentVariable("ABLY_ENV");
+            var env = !string.IsNullOrEmpty(ablyEnv) ? ablyEnv.Trim() : environment;
```
marto83, line 68: I prefer the `.IsNotEmpty()` extension method rather than the double negative

marto83 (approved):

src/IO.Ably.Tests.Shared/Infrastructure/AblyCollectionFixture.cs
tomkirbygreen: The problem is that static code analysis doesn't seem to go down the extension method rabbit hole to check. Also the original version called `System.Environment.GetEnvironmentVariable` twice, which has an associated cost.

PR #807, opened 2021-09-03
Added 'ErrorPolicy' type with 'HandleUnexpected'

tomkirbygreen:
We need to do something more than just sending unexpected exceptions to /dev/nul so at the very least the current implementation now logs that they occured before carrying on.


marto83 (approved):
I like the ErrorPolicy class

PR #812, opened 2021-09-08
'ConnectionFailureSpecs' check 'stateChanges' accumulator

tomkirbygreen:
We go to the trouble of creating and populating the `stateChanges` accumulator so we can at least check it's not empty at the end of a test run.


marto83:
You have left this as a Draft PR. Do you intend to make more changes?

marto83 (approved):

PR #856, opened 2021-09-16
'JsonEncoder' make method private

tomkirbygreen:


marto83 (approved):

src/IO.Ably.Shared/MessageEncoders/JsonEncoder.cs
```diff
@@ -53,7 +53,7 @@ public override Result<ProcessedPayload> Encode(IPayload payload, DecodingContex
             return Result.Ok(new ProcessedPayload(payload));
         }
 
-        public static bool NeedsJsonEncoding(IPayload payload)
+        private static bool NeedsJsonEncoding(IPayload payload)
```
sacOO7, line 56: We can simplify the logic -> 
`(payload.Data is string  || payload.Data is byte[]) == false`

PR #872, opened 2021-09-17
Push/preserve channel order

marto83:
This came into ligth because when I added the Push channel property to both Rest and Realtime channels the order of channels that came out was different than what we had before.


tomkirbygreen (changes requested):
What do you think about moving the container logic (Dictionary, lock and List) out to a separate component with its own unit tests. That way we can encapsulate the locking protocol etc.

marto83:
Are you thinking something like `ChannelConcurrentDictionary` that also exposes an ordered list property?

tomkirbygreen:
Spoke to Martin about this out of band, we're going to let this go through as is. I'll refactor later and move the `OrderedAssociativeContainer` logic out to its own component later.

tomkirbygreen (approved):

PR #880, opened 2021-09-18
'MessageDataConverterTests' fix warning and tidy

tomkirbygreen:


src/IO.Ably.Tests.Shared/CustomSerializers/MessageDataConverterTests.cs
```diff
@@ -1,22 +1,14 @@
...
 namespace IO.Ably.Tests.Shared.CustomSerializers
 {
     public class MessageDataConverterTests
     {
-        private readonly ITestOutputHelper _testOutputHelper;
-        public JsonSerializerSettings JsonSettings = JsonHelper.Settings;
-
-        public MessageDataConverterTests(ITestOutputHelper testOutputHelper)
-        {
-            _testOutputHelper = testOutputHelper;
```
sacOO7, line 18: I am not sure, why this is removed? Actually it was added to log error message containing difference in JSON strings. Or am I missing something here 🤔

tomkirbygreen: I have restored the `private readonly ITestOutputHelper _testOutputHelper` but FYI, there's a code-analysis observation that the field `_testOutputHelper` is assigned to, but never used.

marto83 (approved):

PR #887, opened 2021-09-18
Push/localdevice init

marto83:
The goal of this PR is to make sure we have 1 instance of LocalDevice no matter how many instances of the library we have. 
I had to refactor LocalDevice.cs and move some of the logic from ActivationStateMachine to there. 
I've also updated how we pass MobileDevice (interface that hides the actual physical device)


tomkirbygreen (commented):
I'm having trouble getting this to go green on CI Martin. Could you take a look when you have time. I'll hold off reviewing it until we have a green run.

tomkirbygreen (approved):

PR #891, opened 2021-09-24
'LocalDevice' use collection initializer

tomkirbygreen:


src/IO.Ably.Shared/Push/LocalDevice.cs
```diff
@@ -38,13 +38,18 @@ internal RegistrationToken RegistrationToken
...
+                if (value == null)
                 {
-                    JObject obj = new JObject();
-                    obj.Add("transportType", value.Type);
-                    obj.Add("registrationToken", value.Token);
-                    Push.Recipient = obj;
+                    return;
                 }
+
+                var obj = new JObject
```
marto83, line 46: lovely. Didn't know you could do that?

marto83 (approved):

PR #909, opened 2021-09-24
'ConnectionFailureSpecs' replace lambda with local function

tomkirbygreen:
Unlike lambdas or delegates, local functions do not cause additional overhead because they are essentially regular methods. For example, instantiating and invoking a delegate requires additional members being generated by compiler and causing some memory overhead. Another benefit of local functions is their support for all the syntax elements allowed in regular methods.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailureSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailureSpecs.cs
@@ -250,8 +250,7 @@ namespace IO.Ably.Tests.Realtime.ConnectionSpecs
         [Trait("spec", "RTN14e")]
         public async Task WhenInSuspendedState_ShouldTryAndReconnectAfterSuspendRetryTimeoutIsReached()
         {
-            Func<DateTimeOffset> nowFunc = () => DateTimeOffset.UtcNow;
-            DateTimeOffset NowWrapperFunc() => nowFunc();
+            DateTimeOffset Func() => DateTimeOffset.UtcNow;
 
             FakeTransportFactory.InitialiseFakeTransport =
                 transport => transport.OnConnectChangeStateToConnected = false;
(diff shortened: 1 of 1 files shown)


marto83 (approved):

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailureSpecs.cs
```diff
@@ -262,7 +261,7 @@ public async Task WhenInSuspendedState_ShouldTryAndReconnectAfterSuspendRetryTim
                 opts.AutoConnect = false;
                 opts.DisconnectedRetryTimeout = TimeSpan.FromMilliseconds(10);
                 opts.SuspendedRetryTimeout = TimeSpan.FromMilliseconds(1000);
-                opts.NowFunc = NowWrapperFunc;
+                opts.NowFunc = Func;
```
QuintinWillison, line 264: Perhaps I don't understand the complexity of what I'm looking at, but I'm really curious why you can't just use `DateTimeOffset.UtcNow` here. It's a `static` function - does that make a difference? 🤔 Why do you need the `Func` (bad name, IMO, FWIW) intermediary?

sacOO7 (approved):

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailureSpecs.cs
sacOO7: Seems `DateTimeOffset.UtcNow` is a dotnet property, returns DateTimeOffset object whose date and time are set to the current Coordinated Universal Time (UTC) date.

sacOO7: Though instead of just plain `Func`, we can call it `NowFunc`

marto83: @QuintinWillison we need it wrapped in a function for test purposes. It let's us play with time and fix it for the whole library if necessary. 
+1 for NowFunc()

QuintinWillison: Based on a [basic search of this codebase](https://github.com/ably/ably-dotnet/search?q=NowFunc) it seems there are a few other places where `NowFunc` / `nowFunc` gets assigned to 'plain' `_now` / `Now`. Perhaps worth, perhaps under a separate pull request, addressing consistency there in terms of use of this suffix.

PR #911, opened 2021-09-24
'DocumentationSamples' refactor for consistency

tomkirbygreen:
Various clean-ups since this is meant to be a teaching aid:

1. Replace lambdas with local functions.
2. Consistent style of single line comments.
3. Consistent use of string-interpolation.
4. Stop using now `[Obsolete]` APIs.
5. Make type and members `static` to better convery that each method is isolated.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Samples/DocumentationSamples.cs
+++ b/src/IO.Ably.Tests.Shared/Samples/DocumentationSamples.cs
@@ -8,65 +8,64 @@ using IO.Ably.Realtime;
 
 namespace IO.Ably.Tests.Samples
 {
-    public class DocumentationSamples
+    public static class DocumentationSamples
     {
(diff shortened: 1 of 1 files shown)


QuintinWillison (approved):
Thanks. All good stuff.

The only thing I can see which is perhaps missing from the pull request, as an atomic increment representative of wider change across the codebase, is backup in the form of any linting or other check / analysis to ensure:

1. these changes fix all the contraventions in the scope of code you were considering
2. future modifications to this scope which contravene these consistency rules will break the build in future

I hope I'm making sense on the way I describe that.

I'm not so familiar with the tools available to you with C#, but an example of what I mean and aligning with the spirit of some of your changes in this PR, would be ESLint's [prefer-template](https://eslint.org/docs/rules/prefer-template).

sacOO7 (approved):

PR #912, opened 2021-09-24
Push/localdevice rsh8

marto83:
To be reviewed after https://github.com/ably/ably-dotnet/pull/887


src/IO.Ably.Shared/Push/LocalDevice.cs
```diff
@@ -62,10 +62,94 @@ public static LocalDevice Create(string clientId = null, IMobileDevice mobileDev
...
+
+        internal static void ResetDevice(IMobileDevice mobileDevice)
+        {
+            mobileDevice.ClearPreferences(PersistKeys.Device.SharedName);
+            Instance = null;
+        }
+
+        internal static LocalDevice Instance { get; set; }
+
+        internal static LocalDevice GetInstance(IMobileDevice mobileDevice = null, string clientId = null)
```
tomkirbygreen, line 78: Why are we defaulting this to `null` if it's never correct to call the API when it *is* `null` as it throws?

tomkirbygreen (commented):
Just a question about API design...

src/IO.Ably.Shared/Push/LocalDevice.cs
marto83: Good point. I'll update it. I always pass both parameters so I'll remove the nulls

marto83: I've pushed an update

tomkirbygreen (approved):

PR #915, opened 2021-09-24
Auth/add notify clientid changed

marto83:
Review https://github.com/ably/ably-dotnet/pull/914 first


src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -30,6 +31,9 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
         protected Func<Task<DateTimeOffset>> ServerTime { get; set; }
 
         private TimeSpan? _serverTimeDiff;
+        private TokenDetails _currentToken;
```
sacOO7, line 34: All of these variables can simply be named to 
```
token
tokenParams
clientId
```
there's no previous token variable declared, so it feels odd at first look. Also, clientID is universal for the app (as per domain language, there's no connection specific client ID). We can pass clientID as a param, it can get updated either through requesting a token or setting via connection obj, on first connected message

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -47,18 +51,51 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
...
 
-        internal TokenParams CurrentTokenParams { get; set; }
+        internal TokenParams CurrentTokenParams
+        {
+            get => _currentTokenParams;
+            set
+            {
+                if (value != null)
+                {
+                    NotifyClientIdIfChanged(value.ClientId);
```
sacOO7, line 61: I am not sure about this one, as per spec, tokenParams will be updated once `auth` is successful with newly provided tokenParams, in this case,`clientID` will anyways be updated to a new value. Hence, it will invoke a duplicate `clientID` changed. I am seeing a high chance of duplicate `clientID` changed events being replicated. Maybe a check can be added, only if `clientID` is changed, then send a `clientIDChanged` event.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -47,18 +51,51 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
...
 
-        public TokenDetails CurrentToken { get; set; }
+        public TokenDetails CurrentToken
+        {
+            get => _currentToken;
+            set
+            {
+                if (value != null)
+                {
+                    NotifyClientIdIfChanged(value.ClientId);
```
sacOO7, line 77: Same goes here as per above comment

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -47,18 +51,51 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
...
+        {
+            get => _connectionClientId;
+            set
+            {
+                NotifyClientIdIfChanged(value);
+                _connectionClientId = value;
+            }
+        }
 
+        // TODO: It's really difficult to know when the clientId gets updated
```
sacOO7, line 94: You should remove this todo, I should be able to point to the spec (I have worked on the clientID update story in go, so it's easy to grab the context). In short, clientID is set/updated in the following cases
1. When `clientID` is passed in client options.
2. When implicit/explicit authorize is called, that returns a `tokenDetails` object, so clientID is set to the one present in `tokenDetails` clientID object (though some validations should be performed to make sure its compatible with existing clientID)
3. When connected msg is received, connectionDetails has a `clientID`. So, it should be set after checking compatibility.

sacOO7: Read [RSA7](https://docs.ably.io/client-lib-development-guide/features/#RSA7) for full information regarding `clientID`. 
To know, where clientID set/unset, read [RSA12](https://docs.ably.io/client-lib-development-guide/features/#RSA12) and [RSA7b](https://docs.ably.io/client-lib-development-guide/features/#RSA7b) specifically.
 Though, it's not going to be much different from the cases I mentioned in the above comment.

src/IO.Ably.Shared/AblyAuth.cs
marto83: Have a look at the function. The name implies the function will check it if they are different first before calling the action. The code is here: https://github.com/ably/ably-dotnet/pull/915/files#diff-f382821304556a78f5f9cf153060dd7b1392f314fffe11396d3cba599fc63b2bR441

src/IO.Ably.Shared/AblyAuth.cs
marto83: @sacOO7 this is just me converting Auto properties to ones with a backing store so I can add logic.

src/IO.Ably.Shared/AblyAuth.cs
marto83: @sacOO7 I know the specs and for one reason or another I don't store the clientId in the AblyAuth class. 
There was no need to store the clientId in Auth but just pull it from the various places it can be stored. 
So far I haven't needed to know when the clientId changes for anything. 

I've intentionally flagged it with the TODO because I think we can do better here. However I'm trying to get the Push notifications complete at the same time so I have went with a least resistance approach instead of refactoring how we keep the clientId in the Auth class. 
If @tomkirbygreen thinks it's necessary, let's create an issue and fix it in a branch off of this one.

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: I know, but naming convention sounds confusing ...

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: Yes, I think it solves this issue. 
But again, I am thinking, changed event doesn't make sure `connectionClientId` is updated.

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: It's difficult to grab the exact `TODO` intention, if we take the meaning as is, it feels there's no deterministic way of finding when clientID gets updated.
We are using `clientID` as a part of Auth class in go client. That pretty much solved most of our issues
https://github.com/ably/ably-go/blob/692824c1140f7b7c499134007485924c03560c61/ably/auth.go#L63
Once we add this field, we don't need to fire `notify` event from `tokenparams` or details changed, since we will only be addressing actual `clientID` changed. I feel we should do that since this PR will be dependent on the underlying spec, currently, it feels like being broken to a certain degree.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -47,18 +51,51 @@ internal AblyAuth(ClientOptions options, AblyRest rest)
...
+                {
+                    NotifyClientIdIfChanged(value.ClientId);
+                }
 
-        internal string ConnectionClientId { get; set; }
+                _currentToken = value;
+            }
+        }
+
+        internal string ConnectionClientId
```
sacOO7, line 84: I am not sure about `ConnectionClientId`, clientID is universal & unique across the application.

src/IO.Ably.Shared/AblyAuth.cs
marto83: The changed event is for any update to clientId. I'm going to create an issue to we change how we track the clientId in AblyAuth. 

Also I can see an edge case where the current implementation will notify the wrong clientId. I'll update that as well.

tomkirbygreen (approved):

PR #922, opened 2021-09-26
'ChannelOptionsExtensions' remove unused method

tomkirbygreen:


sacOO7:
@tomkirbygreen This is actually a public method (Supposed to be used by SDK users). Also, the current implementation makes it a lot explicit, contextual, and easier than just setting a plain `rewind` key/value pair. Also, for setting a key/value, you need to explicitly look at rewind documentation, which makes it harder to understand the context.  Adding a link to rewind for reference 
https://ably.com/documentation/realtime/channels/channel-parameters/rewind 
Though I didn't add this method explicitly, I would like to hear your thoughts on this @marto83

tomkirbygreen:
Closing this as both @marto83 and @sacOO7 feel this API constitutes part of our public API surface.

Issue #829, opened 2021-09-15
Skipped Test: TaskUtilsTests.FailingTaskShouldHaveExceptionHandled


tomkirbygreen:
Re-enable the `TaskUtilsTests.FailingTaskShouldHaveExceptionHandled` test.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1184) by [Unito](https://www.unito.io)

PR #927, opened 2021-09-30
Fix skipped test 'TaskUtilsTests.FailingTaskShouldHaveExceptionHandled'

tomkirbygreen:
This corresponds to [Issue 829](https://github.com/ably/ably-dotnet/issues/829)

The fundamental problem is that the test can legitimately return `false` if it
takes longer to execute than the supplied timeout of 500-ms.  Added a `TimeoutExpired`
boolean property that allows the test to detect this and check for the correct value
depending on whether the test completed in time, or the timeout expired.


src/IO.Ably.Tests.Shared/TaskUtilsTests.cs
```diff
@@ -8,26 +8,23 @@ namespace IO.Ably.Tests
...
             }
+
+            var tca = new TaskCompletionAwaiter(500);
+            TaskUtils.RunInBackground(FailingTask(), exception =>
+            {
+                tca.SetCompleted();
+            });
+
+            var result = await tca.Task;
+            result.Should().Be(!tca.TimeoutExpired);
```
sacOO7, line 27: Use `Should().NotBe()` instead

marto83:
@tomkirbygreen, the only reason this test exists is because we have the RunInBackground method with is only used in the `Transport` interface which is really old. 

The `ConnectionManager`. I think should remove TaskUtils completely and use the standard way of running a task in the background with `Task.Run(...)` and not await the task. This way we can wrap the code with a try/catch and the compiler will make sure the generated code is correct. 

So the test can become: 

    try
    {
        _ = Task.Run(FailingTask);
    }
    catch (Exception e)
    {
        tca.SetCompleted();
    }

and we can refactor the Close method to be the same

    try
    {
        _ = _socket.StopConnectionAsync();
    }
    catch (Exception e)
    {
        Logger.Warning(e.Message);
    }

src/IO.Ably.Tests.Shared/TaskUtilsTests.cs
sacOO7: I am not sure why this was closed. Am I missing anything here?

tomkirbygreen: > I am not sure why this was closed. Am I missing anything here?

No my bad. I take your point. Thank you Sachin.

PR #938, opened 2021-10-04
'ListDeviceDetailsRequest' member hides inherited member

tomkirbygreen:
Using the `new` keyword to make this explicit, Martin to confirm please.


marto83 (approved):
Looks good.

PR #946, opened 2021-10-06
'MsWebSocketTransport' explicitly check for null socket.

tomkirbygreen:
Rather than let the runtime throw a `NullReferenceException` I am explicitly checking and delegating to the error handler.

This is an interim step. We have many such `NullReferenceExceptions` being thrown by the runtime during execution. The source of each needs to be investigated to see if it is a latent bug or has somehow become part of our Happy Path through the logic.


tomkirbygreen:
This corresponds to [Issue 949](https://github.com/ably/ably-dotnet/issues/949) .

PR #996, opened 2021-11-06
Fix redundant argument with default value ceremony

tomkirbygreen:


src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
```diff
@@ -674,5 +666,13 @@ private static object DecodeData(string data, string encoding)
                     return data;
             }
         }
+
+        private static ChannelOptions GetOptions(JObject data)
```
sacOO7, line 670: can be renamed to `GetChannelCipherOptions`

src/IO.Ably.NETStandard20/Platform.cs
```diff
@@ -7,7 +7,7 @@ namespace IO.Ably
 {
     internal class Platform : IPlatform
     {
-        private static readonly object _lock = new object();
+        private static readonly object Lock = new object();
```
sacOO7, line 10: I am not sure, why we needed to rename `_lock` to `Lock`?

tomkirbygreen: Because the field is *static* (it's on the type not the instance) and this wins over it's 'field'ness. As a member of the type it gets a Capital and no underscore. It's standard practice and we do it elsewhere in the code base.

src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
tomkirbygreen: Quite possibly, but I was trying to keep the remit of *this* PR limited. I haven't yet looked at how much wiggle-space we have when it comes to improving the names we give our *private* and *internal* methods save for the `WaitFor...` thing.

sacOO7: Great

src/IO.Ably.NETStandard20/Platform.cs
sacOO7: Great

src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
sacOO7: Agreed !! Though my general perception is, it's impossible to keep track of all elements that needs to be changed, so name refactoring can be done at the time it's identified.

sacOO7 (approved):

PR #998, opened 2021-11-07
Fix non-accessed field warning

tomkirbygreen:


src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -292,7 +292,12 @@ public bool UseBinaryProtocol
 #else
             get { return false; }
 #endif
+
+#if MSGPACK
             set { _useBinaryProtocol = value; }
+#else
```
sacOO7, line 298: Do we need `else` statement here? We can just end with `endif`

sacOO7: Also, I am thinking, we can possibly move L297 to under L19. This is optional though

tomkirbygreen: We need the `#else` since we call the setter in code.

sacOO7: Got it. Maybe we could add conditional statement there too. Just that it looks too redundant. I am approving PR for now, you can decide to change it.

sacOO7 (approved):

PR #1001, opened 2021-11-08
Fix 'possible InvalidOperationException' warning

tomkirbygreen:


src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailuresOnceConnectedSpecs.cs
```diff
@@ -196,6 +197,7 @@ public async Task WhenTransportCloses_ShouldResumeConnection()
 
             var firstTransport = LastCreatedTransport;
             var connectionKey = client.Connection.Key;
+            Debug.Assert(client.Connection.Serial.HasValue, "Expected a serial number, got null");
```
sacOO7, line 200: Actually, I am not sure why `fluentAssertion` is not used here.

sacOO7 (approved):

src/IO.Ably.Tests.Shared/Realtime/ConnectionSpecs/ConnectionFailuresOnceConnectedSpecs.cs
tomkirbygreen: I'm using a type from the BCL because it is known to the code analysis (CA) engine. So by indicating intent with `Debug.Assert` we tell we know what we're doing. Unfortunately CA doesn't understand / take-into-account Fluent Assertions.

PR #1004, opened 2021-11-08
Remove unreachable code

tomkirbygreen:


src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -225,7 +225,6 @@ internal async Task<IEnumerable<RealtimeCommand>> ProcessCommand(RealtimeCommand
                             return new RealtimeCommand[] { CompleteWorkflowCommand.Create().TriggeredBy(command) };
                         }
 
-                        break;
```
sacOO7, line 228: This change touches an important part of the code, I would request one more approval from @marto83.

tomkirbygreen: Within this `case` all paths `return` so `break` can not be reached.

sacOO7: makes sense

sacOO7 (approved):

Issue #1054, opened 2021-11-22
RNGCryptoServiceProvider is obsolete


tomkirbygreen:
We need to update our usage of the [obselete](https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0023)  `RNGCryptoServiceProvider`.  For now I will just `pragma around` this but it will need to be properly addressed before we're production ready on .NET 6.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1380) by [Unito](https://www.unito.io)

Issue #1007, opened 2021-11-10
Verify no regressions on .NET 6 RTM


tomkirbygreen:
Now that .NET 6 is released verify we have no regressions on Windows, MacOS and Linux.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1362) by [Unito](https://www.unito.io)

Issue #523, opened 2021-06-15
Upgrade from .NET 3.1 to 6 before December 2021.


tomkirbygreen:
.NET 3.1 (which is a Long Term Support, 'LTS', release) has its support end in [December 2022](https://dotnet.microsoft.com/platform/support/policy/dotnet-core), the next LTS release will be .NET 6 which is due to release on [November 9th 2021](https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-4/).

.NET 5 sees its support end in [February 2022.](https://dotnet.microsoft.com/platform/support/policy/dotnet-core) 

We should consider being ready to move to .NET 6 upon its release, dropping .NET 3.1 and .NET 5 at that point.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-937) by [Unito](https://www.unito.io)

tomkirbygreen:
.NET 6 and Visual Studio 2022 is released this week. Build the .NET *Standard* and *Framework* configurations on a clean machine with Visual Studio 2022 and check there are no test regressions.

tomkirbygreen:
Closing this as we didn't quite manage to do it as it was dependent on our getting control of our versioning scheme. Hopefully this will happen this year as we head into .NET 7 land...

PR #1055, opened 2021-11-22
'Crypto' suppress (for now) .NET 6 build error

tomkirbygreen:
We need to update our usage of the [obselete](https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0023)  `RNGCryptoServiceProvider`.  For now I will just `pragma around` this but it will need to be properly addressed before we're production ready on .NET 6.

This relates to the GitHub Issues: [1007](https://github.com/ably/ably-dotnet/issues/1007), [523](https://github.com/ably/ably-dotnet/issues/523), [1054](https://github.com/ably/ably-dotnet/issues/1054)

### DIFF
--- a/src/IO.Ably.Shared/Encryption/Crypto.cs
+++ b/src/IO.Ably.Shared/Encryption/Crypto.cs
@@ -86,7 +86,9 @@ namespace IO.Ably.Encryption
     /// </summary>
     public static class Crypto
     {
+#pragma warning disable SYSLIB0023
         private static readonly RNGCryptoServiceProvider SecureRandom = new RNGCryptoServiceProvider();
+#pragma warning restore SYSLIB0023
 
         private const int IdempotentGeneratedIdLength = 9;
 


sacOO7:
@tomkirbygreen can you check if it's easy to replace the current implementation with `https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator`. As per official doc, it's plain static method.

tomkirbygreen:
>

tomkirbygreen:
@sacOO7 Yes, the change is trivial but the static overload we want is not present in the various versions of the .NET BCL that we currently support. So for now I'm just pragma'ing this. When we have proper versioning we'll be able to do something better. There is a GitHub issue that captures this.

sacOO7 (approved):
Hoping this temporary change will be taken care by proper library implementation in the future.

PR #1066, opened 2021-11-26
'LockedList' move duplicated implementation to shared type

tomkirbygreen:


marto83 (approved):
Nice. I did wonder where LockedList came from :). It makes it much easier to read.

PR #1072, opened 2021-12-08
'Presence' Fix warning about call to 'SuppressFinalize' being used

tomkirbygreen:
We don't need to call `GC.SuppressFinalize` on a type without a destructor.


marto83 (approved):
I've learnt something new :)

sacOO7 (approved):

PR #1075, opened 2021-12-15
Fix use of async lambda with a delegate that returns void

tomkirbygreen:
Fixes use of async lambdas with delegates that return void.


marto83 (approved):
Looks good :)

PR #1078, opened 2022-01-01
Fixed RTP19

### LINKED ISSUE #840: Skipped Test: PresenceSandboxSpecs.PresenceMap_WithExistingMembers_WhenSync_ShouldRemoveLocalMembers_RTP19
Re-enable the `PresenceSandboxSpecs.PresenceMap_WithExistingMembers_WhenSync_ShouldRemoveLocalMembers_RTP19` test.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1196) by [Unito](https://www.unito.io)

sacOO7:
- Fixes #840

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
@@ -911,7 +911,7 @@ namespace IO.Ably.Tests.Realtime
                 }
             }
 
-            [Theory(Skip = "Keeps failing")]
+            [Theory]
             [ProtocolData]
             [Trait("spec", "RTP19")]
             public async Task


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -911,7 +911,7 @@ public async Task Presence_GetMethodBehaviour(Protocol protocol)
                 }
             }
 
-            [Theory(Skip = "Keeps failing")]
+            [Theory]
```
tomkirbygreen, line 914: _How_ does this fix the test please?

tomkirbygreen (changes requested):
I don't understand how this fixes the test. Are you implying that there have been back end changes and the test _as is_ is correct?

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
sacOO7: Yes, I tried running the test multiple times and it doesn't seem to be failing. So, your hunch is correct, there might have been backend changes and the test is not failing anymore. Also, just an idea, from now on, we can add detailed information in the skip as to why the test is failing or error information related to failed test which can help while resolving test issues.

sacOO7: Also, I feel bad I couldn't find a way to fix the test, since as you can see, the test is not failing in the first place.

tomkirbygreen: Thank you for the clarification Sachin. Please update the original commit message to include:

1. A link the related GitHub issue.
2. Details of the testing protocol, this can be something like "I ran the test multiple times locally and under CI and it passed every time there I conclude that can remove the 'Skip' because...".

sacOO7: Done

tomkirbygreen (approved):

PR #1079, opened 2022-01-02
Presence map behavior

### LINKED ISSUE #837: Skipped Test: PresenceSandboxSpecs.PresenceMapBehaviour_ShouldConformToSpec
Re-enable the `PresenceSandboxSpecs.PresenceMapBehaviour_ShouldConformToSpec` test.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1191) by [Unito](https://www.unito.io)

sacOO7:
- Fixes #837

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
@@ -177,6 +177,46 @@ namespace IO.Ably.Tests.Realtime
                         Timestamp = new DateTimeOffset(2000, 1, 1, 1, 1, 2, default(TimeSpan)),
                         Data = string.Empty,
                     },
+                    /* Should be newer than previous one */
+                    new PresenceMessage
+                    {
+                        Action = PresenceAction.Update,
+                        ClientId = "2",
+                        ConnectionId = "2",
+                        Id = "2:3:1",
+                        Timestamp = new DateTimeOffset(2000, 1, 1, 1, 1, 2, default(TimeSpan)),
+                        Data = string.Empty,
+                    },
+                    /* Should be newer than previous one */
+                    new PresenceMessage
+                    {
+                        Action = PresenceAction.Update,
+                        ClientId = "2",
+                        ConnectionId = "2",
+                        Id = "2:3:2",
+                        Timestamp = new DateTimeOffset(2000, 1, 1, 1, 1, 2, default(TimeSpan)),
+                        Data = string.Empty,
+                    },
+                    /* Should be newer than previous one */
+                    new PresenceMessage
+                    {
+                        Action = PresenceAction.Update,
+                        ClientId = "2",
+                        ConnectionId = "2",
+                        Id = "2:3:3",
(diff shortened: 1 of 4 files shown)


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -130,18 +130,21 @@ public async Task PresenceMapBehaviour_ShouldConformToSpec(Protocol protocol)
...
                 await client.WaitForState(ConnectionState.Connected);
 
+                // Block ably server sent sync action
+                client.BlockActionFromReceiving(ProtocolMessage.MessageAction.Sync);
+
                 var channel = client.Channels.Get(channelName);
                 await channel.AttachAsync();
 
                 channel.State.Should().Be(ChannelState.Attached);
+                List<PresenceMessage> receivedPresenceMsgs = new List<PresenceMessage>();
```
tomkirbygreen, line 140: Use `var` please to avoid repeating the type from the RHS to the LHS.

src/IO.Ably.Tests.Shared/TestExtensions.cs
```diff
@@ -50,6 +50,17 @@ internal static void BlockActionFromSending(this IRealtimeClient client, Protoco
             transport.BlockSendActions.Add(action);
         }
 
+        internal static void BlockActionFromReceiving(this IRealtimeClient client, ProtocolMessage.MessageAction action)
+        {
+            var transport = ((AblyRealtime)client).ConnectionManager.Transport as TestTransportWrapper;
```
tomkirbygreen, line 55: Since this is a hard requirement, why don't we change the `as` cast to a regular 'C style' cast, this will throw a `InvalidCastException` for us.

tomkirbygreen (changes requested):

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -230,15 +230,15 @@ public async Task PresenceMapBehaviour_ShouldConformToSpec(Protocol protocol)
 
                 await client.ProcessCommands();
 
-                int n = 0;
+                var n = 0;
```
tomkirbygreen, line 233: Please revert this. Where `var` gets its value is where the type is clearly stated on the right hand side of the expression so the following are good uses of `var`:

```
var a = new List<Foo>();    // Rather than 'Foo a = new List<Foo>();'
var b = Create<Bar>();    // Rather than 'Bar b = Create<Bar>();'
```

But using `var` in place of a built-in type such as `int`, `float` and `long` etc does not add value. So please avoid replacing:

```
int x = 0;
```
with:
```
var x = 0;
```
Since following that path leads us to (before) `long y = 0;` => `var y = (long)0;`.

sacOO7: Sure, this works

tomkirbygreen (approved):

PR #1086, opened 2022-01-11
Push platform specific packages

marto83:
Added nuspec files for both android and ios. Updated the build script to add a new target for creating packages for the push projects. Add a shell script `package-push.sh` to run the build and output the packages in the root directory.

Things for review:
1. Package names. Our main package is called `ably.io` however I don't like the sound of `ably.io.push.android` so I called it `ably.push.android`. What do you think? I'm happy to fall back to the longer version.
2. Package descriptions. Please review 🙏🏼 
3. The build script will do a separate build for the main library and the push packages. I tried to make it happen all at once but the main build needs `windows` for .net framework and the push packages need `macos` otherwise the ios package doesn't build. 

Once we have the packages the next step will be to setup a myget repo and push them there for an internal test before pushing them to nuget.org


nuget/io.ably.push.ios.nuspec
```diff
@@ -0,0 +1,32 @@
...
+        <projectUrl>https://github.com/ably/ably-dotnet</projectUrl>
+        <releaseNotes>https://github.com/ably/ably-dotnet/releases</releaseNotes>
+        <iconUrl>https://s3.amazonaws.com/files.ably.io/logo-70x70.png</iconUrl>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>Ably .NET library to enable push notifications on iOS devices.</description>
+        <copyright>©2022 Ably</copyright>
+        <language />
+        <dependencies>
+
+          <group>
```
tomkirbygreen, line 19: Do we need this empty group?

tomkirbygreen (changes requested):
With respect to package names, whilst I'm sympathetic when it comes to wanting a package name that 'reads' better I think we should follow the established (ugly?) pattern and go with `ably.io.push.android` and `ably.io.push.ios` respectively.

sacOO7:
> With respect to package names, whilst I'm sympathetic when it comes to wanting a package name that 'reads' better I think we should follow the established (ugly?) pattern and go with `ably.io.push.android` and `ably.io.push.ios` respectively.

Yes, and I would also recommend keeping the package name in sync with a respective namespace in the code.

nuget/io.ably.push.ios.nuspec
marto83: we can remove it. I'll do that shortly.

marto83:
Thanks for the comments guys. I'm going to update the package names.

tomkirbygreen (approved):

sacOO7 (approved):

tomkirbygreen (changes requested):
Please resolve the merge conflicts.

tomkirbygreen (approved):

PR #1092, opened 2022-01-17
Presence leave while syncing

### LINKED ISSUE #839: Skipped Test: PresenceSandboxSpecs.PresenceMap_WhenSyncingLeaveStoredAsAbsentAndDeleted
Re-enable the `PresenceSandboxSpecs.PresenceMap_WhenSyncingLeaveStoredAsAbsentAndDeleted` test.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1197) by [Unito](https://www.unito.io)

sacOO7:
- Fixes #839


src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -672,6 +672,9 @@ public async Task PresenceMap_WhenSyncingLeaveStoredAsAbsentAndDeleted(Protocol
                 var client = await GetRealtimeClient(protocol);
                 await client.WaitForState(ConnectionState.Connected);
 
+                // Block ably server sent sync action
+                client.BlockActionFromReceiving(ProtocolMessage.MessageAction.Sync);
```
tomkirbygreen, line 676: I think for now, given that we only have one use case for this API it would be better packaged as a local function, possibly called 'BlockAblyServerSentSyncAction' (that way we can lose the comment).

tomkirbygreen (changes requested):

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
sacOO7: Done !

tomkirbygreen (approved):
Nice. Personally I would have been tempted to make `BlockAblyServerSentSyncAction` a local / nested function of `GetChannel` thus further reducing its scope.

Issue #1126, opened 2022-02-21
Investigate CI failure


tomkirbygreen:
@sacOO7 reports that CI is now routinely failing. As no code has gone in since it was last green this _may_ be down to a change in a dependency.

Reproduce and analyse.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1609) by [Unito](https://www.unito.io)

tomkirbygreen:
Findings thus far, given a new clone of the repo:

.NET Standard builds fine on Mac OS 12.2.1
.NET Standard builds fine on Windows 10 21H2 build 19044.1466
.NET Framework (all release targets) build fine on Windows 10 21H2 build 19044.1466

sacOO7:
I think we can close this issue now, since it's been fixed. Thanks for the PR.

tomkirbygreen:
Now fixed.

PR #1127, opened 2022-02-21
Increase .NET Framework min ver from 4.6.1 to 4.6.2

tomkirbygreen:
@sacOO7 ran into a CI build issue involving `4.6.1`. After some investigation it turns out we need to move on to `4.6.2` because:

```
The following versions of .NET Framework will reach end-of-support on April 26, 2022: 4.5.2, 4.6, and 4.6.1. After this date, security fixes, updates, and technical support for these versions will no longer be provided.

If you're using .NET Framework 4.5.2, 4.6, or 4.6.1, update your deployed runtime to a more recent version, such as .NET Framework 4.6.2, before April 26, 2022 in order to continue to receive updates and technical support.
```

[ref](https://docs.microsoft.com/en-gb/dotnet/framework/install/guide-for-developers)

[GitHub Issue](https://github.com/ably/ably-dotnet/issues/1126)


sacOO7:
LGTM : ), thanks for the PR

PR #1131, opened 2022-03-01
Update release PR guidance to include senior peers

tomkirbygreen:
Update the release PR guidance to include adding senior peers to the review.

PR #1133, opened 2022-03-02
Refine release PR approvals guidance

QuintinWillison:
We can't be sprinkling links to the GitHub ids of individual team members in documentation (though, change log is a clear exception to that).

Replaces the references added in https://github.com/ably/ably-dotnet/pull/1131 (which I wasn't around to review before it merged) with role titles.


tomkirbygreen (approved):

stmoreau (approved):
Agreed that having roles rather than names sounds like a better approach.

owenpearson (approved):

PR #1139, opened 2022-04-02
Fix realtime logging

sacOO7:
- Fixed #1140 
 
~- [x] Remove `isDebug` flag usage, since it introduces unnecessary complexity, and it's already handled as a part of~

PR #1157, opened 2022-06-09
Add message extras to the Message equality comparison

marto83:
Sachin caught the issue when serialising a `ProtocolMessage` containing Messages with only `MessageExtras` and no other properties. As Message extras was not in the Equality comparison the `OnSerializing` method in `ProtocolMessage` thought all the messages were empty and set the whole `Messages` property to `null` thus causing issues.
I don't think it would have caused any real user issues because having a message with only Extras is a pointless. It did however cause issues when Sachin enabled Debug logging in the RealtimeWorkflow class.

### DIFF
--- a/src/IO.Ably.Shared/Types/Message.cs
+++ b/src/IO.Ably.Shared/Types/Message.cs
@@ -108,7 +108,14 @@ namespace IO.Ably
         /// <returns>true / false..</returns>
         protected bool Equals(Message other)
         {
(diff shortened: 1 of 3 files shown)


sacOO7:
Thanks for raising the PR @marto83 , really helps for enabling debug logs for real-time SDK

tomkirbygreen (approved):

sacOO7 (approved):
LGTM :👍

sacOO7:
Related to #1139

PR #1169, opened 2022-06-24
Remove mention of Visual Studio 2017

tomkirbygreen:
Given that Visual Studio 2019 is effectively end of life at this point we should remove mention of Visual Studio 2017 from the documentation.

A follow-up PR will replace mention of Visual Studio 2022.


AndyNicks (approved):

sacOO7 (approved):
LGTM

PR #1172, opened 2022-06-28
Add channel lifecycle support

tomkirbygreen:
JIRA: 1849

### DIFF
--- a/src/IO.Ably.Shared/Rest/ChannelDetails.cs
+++ b/src/IO.Ably.Shared/Rest/ChannelDetails.cs
@@ -1,4 +1,6 @@
-﻿namespace IO.Ably.Rest
+﻿using Newtonsoft.Json;
+
+namespace IO.Ably.Rest
 {
 #pragma warning disable SA1307
 #pragma warning disable SA1600
@@ -8,12 +10,14 @@
         /// <summary>
         /// The required name of the channel including any qualifier, if any.
         /// </summary>
-        public string channelId;
+        [JsonProperty("channelId")]
+        public string ChannelId { get; set; }
 
         /// <summary>
         /// The status and occupancy stats for the channel.
         /// </summary>
-        public ChannelStatus status;
+        [JsonProperty("status")]
+        public ChannelStatus Status { get; set; }
     }
 
     public class ChannelStatus
@@ -21,12 +25,14 @@
         /// <summary>
         /// Indicates whether the channel that is the subject of the event is active.
         /// </summary>
-        public bool isActive;
+        [JsonProperty("isActive")]
+        public bool IsActive { get; set; }
 
         /// <summary>
         /// Metadata relating to the occupants of the channel.
         /// </summary>
-        public ChannelOccupancy occupancy;
+        [JsonProperty("occupancy")]
+        public ChannelOccupancy Occupancy { get; set; }
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


sacOO7:
We need to update README for the updated public interface. Something like this -> https://github.com/ably/ably-php#getting-the-channel-status

src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
```diff
@@ -654,6 +654,29 @@ public async Task WithEncryptionCipherAlgorithmMismatch_ShouldLeaveMessageEncryp
             message.Encoding.Should().Be("utf-8/cipher+aes-128-cbc");
         }
 
+        [Theory]
+        [ProtocolData]
+        [Trait("spec", "RSL8")]
```
sacOO7, line 659: Optional - You can also add `CHD1` as a part of spec

src/IO.Ably.Shared/Rest/RestChannel.cs
```diff
@@ -224,5 +226,17 @@ public PaginatedResult<Message> History(PaginatedRequestParams query)
         {
             return AsyncHelper.RunSync(() => HistoryAsync(query));
         }
+
+        private async Task<ChannelDetails> StatusAsync()
```
sacOO7, line 230: Abstract documentation can be added to public interface. i.e. following is added to `ably-php`
`     * Retrieves current channel active status with no. of publishers, subscribers, presenceMembers etc
     * @return ChannelDetails`

src/IO.Ably.Shared/Rest/IRestChannel.cs
```diff
@@ -101,6 +101,12 @@ public interface IRestChannel
         /// <param name="query"><see cref="PaginatedRequestParams"/> query.</param>
         /// <returns><see cref="PaginatedResult{T}"/> of Messages.</returns>
         PaginatedResult<Message> History(PaginatedRequestParams query);
+
+        /// <summary>
+        /// Returns channel lifecycle status.
+        /// </summary>
+        /// <returns><see cref="ChannelDetails"/>Channel Details.</returns>
```
sacOO7, line 108: Doc can be made more verbose i.e. following is added to `ably-php`
* Retrieves current channel active status with no. of publishers, subscribers, presenceMembers etc * @return ChannelDetails

src/IO.Ably.Shared/Rest/ChannelDetails.cs
```diff
@@ -0,0 +1,72 @@
+﻿namespace IO.Ably.Rest
```
marto83, line 1: What is the reason to use fields instead of properties? 
Everywhere else we use properties for DTOs :)

tomkirbygreen: Done.

marto83 (approved):
Looks good

sacOO7:
> We need to update README for the updated public interface. Something like this -> https://github.com/ably/ably-php#getting-the-channel-status

Can we update README .. i think it takes a high priority

src/IO.Ably.Tests.Shared/Rest/ChannelSandboxSpecs.cs
tomkirbygreen: Done.

sacOO7 (approved):
LGTM

PR #1183, opened 2022-09-06
Remove connection serial

### LINKED ISSUE #1181: Implement no-connection-serial
Implement the changes listed in https://github.com/ably/docs/pull/1526

sacOO7:
- Fixes #1181


src/IO.Ably.Shared/Realtime/RealtimeChannels.cs
```diff
@@ -246,5 +246,31 @@ private void HandleInitialiseFailedChannelsCommand(RealtimeChannel channel)
...
+
+        // RTN16j, RTL15b
+        internal void SetChannelSerialsFromRecoverOption(IDictionary<string, string> serials)
+        {
+            foreach (var keyValuePair in serials)
+            {
+                var channelName = keyValuePair.Key;
+                var channelSerial = keyValuePair.Value;
+                var channel = (RealtimeChannel)this[channelName];
+                if (channel != null)
```
andydunstall, line 258: Where is the channel serial set for channels attached after the transport is created?

SimonWoolf (approved):
At a quick look through, can't see any concerns (other than the one Andy has raised)

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -358,7 +358,7 @@ public bool UseBinaryProtocol
         /// Default before 1.2: false.
         /// Default from 1.2: true.
         /// </summary>
-        public bool IdempotentRestPublishing { get; set; } = Defaults.ProtocolVersionNumber >= 1.2;
+        public bool IdempotentRestPublishing { get; set; } = Convert.ToDouble(Defaults.ProtocolVersion) >= 1.2;
```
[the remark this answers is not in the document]
sacOO7: Yeah, I think I should probably change it since it's a global constant.

src/IO.Ably.Shared/Realtime/RealtimeChannels.cs
sacOO7: We are setting it in `src/IO.Ably.Shared/Transport/ConnectionManager.cs` . You can check the file. It's right after the transport is created

andydunstall: What about when the channel is attached after the transport is created?

Given its attached by the developer rather than automatically we don't know when they'll attach (and still need to load the channel serial from the recovery key)

sacOO7: I didn't get your question ... the current implementation makes sure before connecting to the ably server, channel serials are properly set.

src/IO.Ably.Shared/ClientOptions.cs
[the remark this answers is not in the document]
sacOO7: Fixed

src/IO.Ably.Shared/Realtime/RealtimeChannels.cs
andydunstall: Yeah but with recover its up to the developer to explicitly reattach the channel right? So we don't know if they will attach their channels before or after we connect to the ably server - though even if they attach a channel 5 minutes after we connect to the server we still need to use the recovered channel serial

So my question was if a channel is attached by the developer after we connect to the ably server - where is the channel serial of that channel set from the recovery key?

sacOO7: We are setting it in file `src/IO.Ably.Shared/Transport/ConnectionManager.cs`

andydunstall (approved):

sacOO7:
Merging the PR, since it's been approved.

PR #1195, opened 2022-12-28
Feature/agent header

### LINKED ISSUE #498: Implement RSC7d (Ably-Agent header)
see [PR](https://github.com/ably/docs/pull/1034) and [spec point](https://ably-docs.herokuapp.com/client-lib-development-guide/features/#RSC7d) for context.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-683) by [Unito](https://www.unito.io)

sacOO7:
- Added basic ably agent header.
- Added test to check for agent header
- Referenced from https://github.com/ably/ably-php/pull/137 and https://github.com/ably/ably-go/pull/360
- Updated ably-common -> https://github.com/ably/ably-common/pull/177
- Updated dotnet-features https://github.com/ably/features/pull/104
- Fixed #498

### DIFF
--- a/src/IO.Ably.Shared/Defaults.cs
+++ b/src/IO.Ably.Shared/Defaults.cs
@@ -126,31 +126,31 @@ namespace IO.Ably
             return string.Empty;
         }
 
-        internal static string Agents(Dictionary<string, string> additionalAgents)
+        internal static string AblyAgentIdentifier(Dictionary<string, string> additionalAgents)
         {
-            var agentsIdentifier = $"{AblySdkIdentifier}";
+            var agentIdentifier = $"{AblySdkIdentifier}";
 
             var ablyDotnetRuntimeIdentifier = DotnetRuntimeIdentifier();
             if (!string.IsNullOrEmpty(ablyDotnetRuntimeIdentifier))
             {
-                agentsIdentifier += $" {ablyDotnetRuntimeIdentifier}";
+                agentIdentifier += $" {ablyDotnetRuntimeIdentifier}";
             }
 
             if (additionalAgents == null)
             {
-                return agentsIdentifier;
+                return agentIdentifier;
             }
 
             foreach (var (agentId, agentVersion) in additionalAgents.Where(agent => !string.IsNullOrEmpty(agent.Key)))
             {
-                agentsIdentifier += $" {agentId}";
+                agentIdentifier += $" {agentId}";
                 if (!string.IsNullOrEmpty(agentVersion))
                 {
-                    agentsIdentifier += $"/{agentVersion}";
+                    agentIdentifier += $"/{agentVersion}";
                 }
             }
 
-            return agentsIdentifier;
+            return agentIdentifier;
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/IO.Ably.Shared/Defaults.cs
```diff
@@ -109,5 +98,60 @@ internal static string[] GetEnvironmentFallbackHosts(string environment)
...
+                case "xamarin-ios":
+                    return "xamarin-iOS";
+            }
+
+            return string.Empty;
+        }
+
+        internal static string AgentHeaders(Dictionary<string, string> additionalHeaders)
+        {
+            var agentHeaders = $"{AblySdkIdentifier}";
```
owenpearson, line 131: This variable name seems misleading, if it's a set of headers it should be a `Dictionary<string, string>` but in this case it's just a `string` so I think you should call it `agentIdentifier` (as it's called in RSC7d)

src/IO.Ably.Shared/Defaults.cs
```diff
@@ -109,5 +98,60 @@ internal static string[] GetEnvironmentFallbackHosts(string environment)
...
+
+            var ablyDotnetRuntimeIdentifier = DotnetRuntimeIdentifier();
+            if (!string.IsNullOrEmpty(ablyDotnetRuntimeIdentifier))
+            {
+                agentHeaders += $" {ablyDotnetRuntimeIdentifier}";
```
owenpearson, line 136: This logic seems kinda overcomplicated, would it not be simpler to collect the agents into a `String[]` and just return `String.join(" ", agents)`?

owenpearson (changes requested):

src/IO.Ably.Shared/Defaults.cs
sacOO7: It's not complicated, it's a widely used syntax in dotnet. 
It's called `string interpolation` https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

sacOO7: You can see recommended ways to concatenate strings - https://learn.microsoft.com/en-us/dotnet/csharp/how-to/concatenate-multiple-strings.
`string.join` / `string.concat` needs collection/list to operate on.

src/IO.Ably.Shared/Defaults.cs
sacOO7: Fixed!

src/IO.Ably.Shared/Defaults.cs
sacOO7: done!

owenpearson: nice, thanks! the tests are failing now because it expects the agents to be in a specific order, but once they're passing i'll approve 👍

sacOO7: Okay, fixed as a part of https://github.com/ably/ably-dotnet/pull/1195/commits/b40d7454ee3f6552f8d8f7eb62c27787c8ca99f1. I was passing `AblySdkIdentifier` instead of `product` value, so tests were failing

owenpearson (approved):
LGTM, thanks

sacOO7:
Merging the PR since it's approved by @owenpearson : )

PR #1197, opened 2023-01-02
Fix authCallback token string

### LINKED ISSUE #1196: AuthCallback returned a string which can't be converted to TokenRequest
- When token string/JWT string is passed in the authCallback it throws the above error.
- There is a bad check inside, that accepts and tries to cast a string token to TokenRequest without proper conversion.
- Link to internal conversation - https://ably-real-time.slack.com/archives/C0174D4AP4P/p1671637382980649

sacOO7:
- Fixes #1196

### DIFF
--- a/src/IO.Ably.Shared/AblyAuth.cs
+++ b/src/IO.Ably.Shared/AblyAuth.cs
@@ -559,7 +559,7 @@ namespace IO.Ably
         /// <param name="tokenParams"><see cref="TokenParams"/>. If null a token request is generated from options passed when the client was created.</param>
         /// <param name="authOptions"><see cref="AuthOptions"/>. If null the default AuthOptions are used.</param>
         /// <returns>signed token request.</returns>
-        public async Task<TokenRequest> CreateTokenRequestAsync(TokenParams tokenParams, AuthOptions authOptions)
+        public async Task<TokenRequest> CreateTokenRequestObjectAsync(TokenParams tokenParams, AuthOptions authOptions)
         {
             authOptions = authOptions ?? CurrentAuthOptions ?? Options;
             tokenParams = tokenParams ?? CurrentTokenParams ?? TokenParams.WithDefaultsApplied();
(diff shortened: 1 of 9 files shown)


sacOO7:
@owenpearson I will be marking the old interface that returns `string` for `CreateTokenRequestAsync` and `CreateTokenRequest` as deprecated. We need to discuss if this should be made as breaking change or not, since the method signature is totally wrong as per spec -> https://sdk.ably.com/builds/ably/specification/main/features/#RSA9a

sacOO7:
Updated as per discussion, maintained public methods.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -694,6 +669,20 @@ public TokenDetails Authorise(TokenParams tokenParams = null, AuthOptions option
             return AsyncHelper.RunSync(() => AuthorizeAsync(tokenParams, options));
         }
 
+        public TokenRequest CreateTokenRequestObject(TokenParams tokenParams = null, AuthOptions authOptions = null)
+        {
+            return AsyncHelper.RunSync(() => CreateTokenRequestObjectAsync(tokenParams, authOptions));
+        }
+
+        [Obsolete("This method will be removed in the future, please replace with a call to CreateTokenRequestObjectAsync")]
```
owenpearson, line 677:
```suggestion
        [Obsolete("This method will be removed in a future version, please use CreateTokenRequestObjectAsync instead")]
```

src/IO.Ably.Tests.Shared/Infrastructure/TaskCompleterAwaiterTests.cs
```diff
@@ -17,7 +17,7 @@ public void TimeoutElapsedIsInitiallyFalse()
         public async void TimeoutElapsedSignalsOnTimeout()
         {
             var sut = new TaskCompletionAwaiter(10);
-            await Task.Delay(1000);
+            await Task.Delay(1100);
```
owenpearson, line 20: what's the reason for this change?

owenpearson (changes requested):
Looks good, just a couple of minor comments

src/IO.Ably.Tests.Shared/Infrastructure/TaskCompleterAwaiterTests.cs
sacOO7: Actually, the test was failing saying `timeoutExpired` as `false`.

sacOO7: I should be reverting the change since it doesn't really fix the issue lol. Maybe it was a sandbox issue.

src/IO.Ably.Shared/AblyAuth.cs
sacOO7: updated!

src/IO.Ably.Tests.Shared/Infrastructure/TaskCompleterAwaiterTests.cs
sacOO7: removed the change!

src/IO.Ably.Shared/AblyAuth.cs
owenpearson: thanks!

owenpearson (approved):
LGTM. thanks

sacOO7:
Thanks for approval 👍

Issue #1187, opened 2022-11-25
Protocol-v2: readd recoveryKey to make this a non-breaking change


SimonWoolf:
see https://github.com/ably/specification/pull/119

sacOO7:
Closed via #1198

PR #1199, opened 2023-01-18
[SDK-3703] Feature/integration 2.0

### LINKED ISSUE #1244: Review and merge integration 2.0
Related to #1199
- Sync with the main.
- Need to re-review and add missing tests ( if needed ).
- Merge with main.

sacOO7:
- Implemented no connection serial as per https://github.com/ably/specification/pull/88/files 
- Deprecated the recoveryKey support as per https://github.com/ably/ably-dotnet/issues/1187
- For more details -> https://github.com/ably/ably-js/pull/1110
- Fixed #1244


SimonWoolf (approved):
Not very familar with .net, but afaict this looks good to me

src/CommonAssemblyInfo.cs
```diff
@@ -4,13 +4,13 @@
 
 [assembly: AssemblyCompany("Ably")]
 [assembly: AssemblyProduct("Ably .NET Library")]
-[assembly: AssemblyVersion("1.2.9")]
-[assembly: AssemblyFileVersion("1.2.9")]
+[assembly: AssemblyVersion("2.0.0")]
```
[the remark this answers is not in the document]
sacOO7: Seems we don't need to change it. I will update according to minor version change 👍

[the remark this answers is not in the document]
sacOO7: I have removed the changes from the file.

src/IO.Ably.Shared/Realtime/ChannelMessageProcessor.cs
```diff
@@ -63,7 +74,7 @@ public Task<bool> MessageReceived(ProtocolMessage protocolMessage, RealtimeState
                         // RTL12
                         if (!protocolMessage.HasFlag(ProtocolMessage.Flag.Resumed))
                         {
-                            channel.Presence.ChannelAttached(protocolMessage);
+                            channel.Presence.HandleAlreadyAttachedChannel(protocolMessage);
```
sacOO7, line 77: Need to check if the presence needs to start sync only when the channel was already attached. 
As per `ably-go` implementation, this is more simplified. 

This has nothing to do with `spec-2.0`, rather might be a existing bug in the library,

sacOO7: This has been re-implemented : )

src/IO.Ably.Shared/Realtime/PresenceMap.cs
```diff
@@ -121,15 +138,16 @@ public void StartSync()
...
-            if (!IsSyncInProgress)
+            if (!SyncInProgress)
             {
                 lock (_lock)
                 {
-                    _residualMembers = new HashSet<string>(_members.Keys);
-                    IsSyncInProgress = true;
+                    _beforeSyncMembers = new HashSet<string>(_members.Keys); // RTP19
+                    SyncInProgress = true;
+                    SyncCompleted = false;
```
SimonWoolf, line 150: Why do we need both of these? they're always the opposite of each other, right?

SimonWoolf (approved):
Looks good to me, one small comment

src/IO.Ably.Shared/Realtime/PresenceMap.cs
sacOO7: so, I have to maintain an enum stating 3 states. initial, progress and completed

PR #1201, opened 2023-01-23
Release/1.2.10

sacOO7:
- Includes support for ably agent headers.
- Includes authCallback fix for accepting TokenRequest Object/ string token.
- Referred to https://github.com/ably/ably-dotnet/pull/1180


CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
...
+
+[Full Changelog](https://github.com/ably/ably-dotnet/compare/1.2.9...1.2.10)
+
+**Implemented enhancements:**
+
+- Implement RSC7d \(Ably-Agent header\) [\#498](https://github.com/ably/ably-dotnet/issues/498)
+
+**Closed issues:**
+
+- AuthCallback returned a string which can't be converted to TokenRequest [\#1196](https://github.com/ably/ably-dotnet/issues/1196)
```
owenpearson, line 13: I think we can tidy up the changelog entry a bit here:
- "Errors while creating transport!" should probably not have an exclamation mark
- Stuff like "Validate against 1.0 specification", "Create feature compliance manifest" etc. aren't customer facing so we should remove them.
- Ably-Agent header isn't really an enhancement

owenpearson (changes requested):
Looks generally fine overall, but I don't know what `dotnet fake` does and I don't know about the unity release process so no opinion on that stuff

CHANGELOG.md
sacOO7: Errors while creating transport! is actually the name of the issue! So changelog captured it as is!

QuintinWillison: What I often do is go back to the issue and also modify the title there. Nothing wrong with that. This is a good point to do that kind of housekeeping.

CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+- Unity iOS: Errors while creating transport!. [\#764](https://github.com/ably/ably-dotnet/issues/764)
+- Unity: When building for iOS, an IL2CPP error prevents the app working [\#751](https://github.com/ably/ably-dotnet/issues/751)
+- Unity 2019.3.5 - System.EntryPointNotFoundException: CreateNLSocket [\#408](https://github.com/ably/ably-dotnet/issues/408)
+- Send unhandled exceptions to Ably exception service - RSC20 [\#389](https://github.com/ably/ably-dotnet/issues/389)
+
+**Merged pull requests:**
+
+- Fix authCallback token string [\#1197](https://github.com/ably/ably-dotnet/pull/1197) ([sacOO7](https://github.com/sacOO7))
+- Feature/agent header [\#1195](https://github.com/ably/ably-dotnet/pull/1195) ([sacOO7](https://github.com/sacOO7))
+- Override the default Dependabot configuration for NuGet [\#1188](https://github.com/ably/ably-dotnet/pull/1188) ([QuintinWillison](https://github.com/QuintinWillison))
```
QuintinWillison, line 23: The work done in this PR carries no value for users of the SDK, so I would remove this entry entirely.

QuintinWillison (dismissed):
As far as I can see there are "closed issues" in this log which relate to things that are not in the `main` branch yet (that can be seen from correlating with the merged pull requests list). This is a limitation of the Github changelog generator tool we use, so requires manual intervention to remove them. I think this list should be a lot shorter for it to represent what users are actually getting for new in this SDK release.

Also, while we're here, perhaps worth updating `COPYRIGHT` given this is the first release we're doing in 2023. 😁

CHANGELOG.md
sacOO7: Updated!

owenpearson (approved):
LGTM, thanks

sacOO7:
Merging the PR since it's been approved : )

PR #1210, opened 2023-02-13
Support customContext 

### LINKED ISSUE #1209: Fix synchronization context deprecation
- Synchronization context is marked as deprecated, yet it's necessary to use in multiple cases
- As per discussion with @marto83 , we need it in the .net apps where callbacks need to be executed on the UI thread.
e.g. unity, maui

sacOO7:
Fixed #1209


owenpearson (approved):
LGTM, thanks

marto83 (approved):
👍🏼

PR #1216, opened 2023-03-21
Fix QueueMessages when set to false

### LINKED ISSUE #1219: Fix QueueMessages ClientOption
Customer reported a issue where in some cases messages were getting queued even when `clientOptions.QueueMessages` was set to `false`.

Link to internal discussion
https://ably-real-time.slack.com/archives/C0174D4AP4P/p1679402280230259

sacOO7:
~Fixed queueing that happens for following states~
~CONNECTIONCLOSING, CONNECTIONCLOSED, CONNECTIONSUSPENDED, CONNECTIONFAILED~
~when `QueueMessages=false`~

- Fixed queueing where RealtimeWorkflowScheduler is queuing messages on connected state and sending them after client is disconnected.
- Added missing check to fix the same.
- Implemented https://github.com/ably/specification/issues/134
- Fixed #1219

### DIFF
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -226,17 +226,20 @@ namespace IO.Ably.Transport
                 Logger.Error($"Failed to encode protocol message: {encodingResult.Error.Message}");
             }
 
-            if (State.CanSend == false && State.CanQueue == false)
+            if (State.CanSend == false)
             {
-                throw new AblyException($"The current state [{State.State}] does not allow messages to be sent.");
-            }
+                if (Options.QueueMessages == false)
+                {
+                    throw new AblyException(
+                        $"Not queuing messages for [{State.State}] since Options.QueueMessages is set to False.",
+                        Connection.ConnectionState.DefaultErrorInfo.Code,
+                        HttpStatusCode.ServiceUnavailable);
+                }
 
-            if (State.CanSend == false && State.CanQueue && Options.QueueMessages == false)
-            {
-                throw new AblyException(
-                    $"Current state is [{State.State}] which supports queuing but Options.QueueMessages is set to False.",
-                    Connection.ConnectionState.DefaultErrorInfo.Code,
-                    HttpStatusCode.ServiceUnavailable);
+                if (State.CanQueue == false)
+                {
+                    throw new AblyException($"The current state [{State.State}] does not allow messages to be sent.");
+                }
             }
 
             ExecuteCommand(SendMessageCommand.Create(message, callback).TriggeredBy("ConnectionManager.Send()"));
--- a/src/IO.Ably.Shared/Transport/ConnectionManager.cs
+++ b/src/IO.Ably.Shared/Transport/ConnectionManager.cs
@@ -232,13 +232,16 @@ namespace IO.Ably.Transport
                 {
                     throw new AblyException(
                         $"Not queuing messages for [{State.State}] since Options.QueueMessages is set to False.",
-                        Connection.ConnectionState.DefaultErrorInfo.Code,
-                        HttpStatusCode.ServiceUnavailable);
+                        ErrorInfo.ReasonUnknown.Code,
+                        HttpStatusCode.BadRequest);
                 }
 
                 if (State.CanQueue == false)
                 {
-                    throw new AblyException($"The current state [{State.State}] does not allow messages to be sent.");
+                    throw new AblyException(
+                        $"The current connection state [{State.State}] does not allow messages to be sent.",
+                        ErrorInfo.ReasonUnknown.Code,
+                        HttpStatusCode.BadRequest);
                 }
             }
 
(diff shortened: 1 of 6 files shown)


sacOO7:
Tests need to be added to fix the issue properly

sacOO7:
It seems test can be updated for `send` using `TestTransportFactory`

sacOO7:
Need to mock send response, failing the message and then checking the failed message doesn't get added to the queue.

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
```diff
@@ -1662,6 +1662,58 @@ public async Task ChannelStateCondition_WhenQueueMessagesIsFalse_WhenChannelIsIn
                     client.Close();
                 }
 
+                [Theory]
+                [ProtocolData]
+                [Trait("spec", "RTP16b")]
+                public async Task ChannelStateCondition_WhenQueueMessagesIsFalse_ShouldntQueueMessages_WhenSendFails(Protocol protocol)
```
owenpearson, line 1668: I don't think this is actually testing for the correct behaviour of `ClientOptions.QueueMessages`.

The way I would suggest testing it is:
1. Create a realtime client with `QueueMessages = false`
2. Force the client into the `DISCONNECTED` state.
3. Assert that calling `RealtimeChannel.Publish` (or `PublishAsync`) results in an exception
4. Assert that the message queue is empty

It might be worth also having another test that checks the behaviour when `QueueMessages` is true":
1. Create a realtime client with `QueueMessages = true`
2. Force the client into the `DISCONNECTED` state.
3. Call `RealtimeChannel.Publish` (or `PublishAsync`)
4. Assert that the message is in the message queue
5. Force the client back into the `CONNECTED` state
6. Assert that the message has now been published

owenpearson (changes requested):

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
sacOO7: We already have test for both cases
For first case -
https://github.com/ably/ably-dotnet/blob/4d780978f4aff702b9ab0404bdc78591f518a65a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs#L1629

For second case -
https://github.com/ably/ably-dotnet/blob/4d780978f4aff702b9ab0404bdc78591f518a65a/src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs#L1668

Currently, we are adding a test for a really special case where if send fails, it should check if `QueueMessages` is false, accordingly it should handle message queuing.

sacOO7: I think it was missing before, that's why we had a issue where sometimes, messages were being queued.

src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -226,17 +226,20 @@ internal async Task OnAuthUpdated(TokenDetails tokenDetails, bool wait)
...
+            if (State.CanSend == false)
             {
-                throw new AblyException($"The current state [{State.State}] does not allow messages to be sent.");
-            }
+                if (Options.QueueMessages == false)
+                {
+                    throw new AblyException(
+                        $"Not queuing messages for [{State.State}] since Options.QueueMessages is set to False.",
+                        Connection.ConnectionState.DefaultErrorInfo.Code,
+                        HttpStatusCode.ServiceUnavailable);
```
SimonWoolf, line 236: 503 is the code for a serverside issue, not appropriate for client disconnected. 400 isn't perfect but better

src/IO.Ably.Shared/Transport/ConnectionManager.cs
```diff
@@ -226,17 +226,20 @@ internal async Task OnAuthUpdated(TokenDetails tokenDetails, bool wait)
...
 
-            if (State.CanSend == false && State.CanQueue && Options.QueueMessages == false)
-            {
-                throw new AblyException(
-                    $"Current state is [{State.State}] which supports queuing but Options.QueueMessages is set to False.",
-                    Connection.ConnectionState.DefaultErrorInfo.Code,
-                    HttpStatusCode.ServiceUnavailable);
+                if (State.CanQueue == false)
+                {
+                    throw new AblyException($"The current state [{State.State}] does not allow messages to be sent.");
```
SimonWoolf, line 241: Shouldn't this have a Code and StatusCode?

SimonWoolf (commented):
Couple comments, overall looks plausible to me but I'll leave it to Owen to approve when he's happy

src/IO.Ably.Shared/Transport/ConnectionManager.cs
sacOO7: updated

src/IO.Ably.Tests.Shared/Realtime/PresenceSandboxSpecs.cs
sacOO7: I think we have feature spec updated as a part of https://github.com/ably/specification/issues/134

owenpearson (approved):
LGTM, thanks

owenpearson (approved):
LGTM, thanks

sacOO7:
Thanks for the approval, merging the PR : )

PR #1245, opened 2023-06-27
MAUI config doc

### LINKED ISSUE #1242: Update README for MAUI trimming config
- Update README as per https://github.com/ably/ably-dotnet/issues/1241#issuecomment-1598283469

sacOO7:
- Fixed #1242


README.md
```diff
@@ -13,7 +13,7 @@ This is a .NET client library for Ably. The library currently targets the [Ably
 ## Supported platforms
 
 * [.NET Standard 2.0+](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0)
-* .NET 6.x, 7.x (MAUI supported)
+* .NET 6.x, 7.x, MAUI, check [MAUI config](#maui-configuration).
```
owenpearson, line 16: Out of interest, how come MAUI isn't a separate item in the list?

README.md
```diff
@@ -446,6 +446,17 @@ options.TransportFactory = new MsWebSocketTransport.TransportFactory(websocketOp
 var realtime = new AblyRealtime(options);
 ```
 
+### MAUI configuration
+- Make sure to disable assembly trimming for `IO.Ably`.
+- Add following to the `.csproj` file.
+```
```
owenpearson, line 452:
```suggestion
```xml
```

README.md
```diff
@@ -446,6 +446,17 @@ options.TransportFactory = new MsWebSocketTransport.TransportFactory(websocketOp
 var realtime = new AblyRealtime(options);
 ```
 
+### MAUI configuration
+- Make sure to disable assembly trimming for `IO.Ably`.
```
owenpearson, line 450: I think we should explain why users should disable assembly trimming - I looked at the microsoft docs and It's not clear to me why I would need to do this

sacOO7: It's needed to avoid assembly stripping that happens when MAUI tries to build release build. This is generally to strip unnecessary dependencies that are not referenced directly. 
Since `ably-dotnet` SDK is using reflection API at multiple places, it's hard for MAUI to detect classes that are actually being used. All, we can suggest is not to strip any part of `IO.Ably.dll`, while doing assembly stripping.

README.md
sacOO7: Actually, MAUI is a framework that is supported as a part of `.net6` onwards. It doesn't have a separate runtime, unlike xamarin.

README.md
sacOO7: We are doing the same in `ably-unity` SDK.

README.md
sacOO7: https://learn.microsoft.com/en-us/dotnet/maui/what-is-maui

README.md
owenpearson: OK that's good to know. So how about something like this:
```
### MAUI configuration
Since ably-dotnet makes use of the reflection API, MAUI assembly trimming may cause issues. When using MAUI. we recommend adding the following to your `.csproj` file to disable assembly trimming:
```

README.md
owenpearson: Thanks for explaining!

README.md
sacOO7: Okay, I will update this.

sacOO7: updated!

owenpearson (approved):
LGTM 👍

PR #1247, opened 2023-06-30
Updated connectionTtl to be 120 seconds

### LINKED ISSUE #1243: Update connectionStateTtl
- Update connectionStateTtl to 2 mins as per spec.

sacOO7:
- Fixed #1243

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -85,11 +85,17 @@ namespace IO.Ably.Tests.Realtime
         [ProtocolData]
         public async Task ShouldSaveConnectionStateTtlToConnectionObject(Protocol protocol)
         {
-            var client = await GetRealtimeClient(protocol);
+            var client = await GetRealtimeClient(protocol, (options, _) => options.AutoConnect = false);
+            client.State.Connection.ConnectionStateTtl = TimeSpan.FromSeconds(60);
+            client.State.Connection.ConnectionStateTtl.Should().Be(TimeSpan.FromSeconds(60));
+            client.Connection.ConnectionStateTtl.Should().Be(TimeSpan.FromSeconds(60));
 
+            client.Connect();
             await WaitForState(client);
 
-            client.Connection.ConnectionStateTtl.Should().Be(Defaults.ConnectionStateTtl);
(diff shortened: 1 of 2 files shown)


src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -89,7 +89,7 @@ public async Task ShouldSaveConnectionStateTtlToConnectionObject(Protocol protoc
 
             await WaitForState(client);
 
-            client.Connection.ConnectionStateTtl.Should().NotBe(Defaults.ConnectionStateTtl);
+            client.Connection.ConnectionStateTtl.Should().Be(Defaults.ConnectionStateTtl);
```
owenpearson, line 92: This test change isn't quite right - what this test does is ensure that once a client is connected, the `Connection.ConnectionStateTtl` field is populated with the `connectionStateTtl` from connection details. I'm not sure how easy this is to do with dotnet tests, but to test this properly you need to change the value of `Defaults.ConnectionStateTtl` and assert that the value of `Connection.ConnectionStateTtl` changes when you become connected (as the test was doing before).

owenpearson (changes requested):

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
sacOO7: Then this test should be removed wdyt? I think currently we are just checking if test doesn't fail.

sacOO7: Also. we haven't marked the test with any spec id yet.

owenpearson: I don't think we should remove the test, the behaviour is defined by [`DF1a`](https://sdk.ably.com/builds/ably/specification/main/features/#DF1a).

If we don't test this and someone breaks it then we could end up in a situation where we change the server-side `connectionStateTtl` but ably-dotnet clients just ignore the new value, therefore it's important to test that the `connectionStateTtl` from connection details is used

sacOO7: Updated test as per suggestion.

PR #1251, opened 2023-08-02
[SDK-3770] Fix logging issue across multiple rest/realtime/push instances

### LINKED ISSUE #1250: LogHandler is shared between multiple AblyRealtime instances
I think if I am implementing a custom logging handler, I might want to have AblyRealtime instance-specific things that I would want to log using the custom log handler. 

Currenlty since the AblyRealtime is instantiated with a static DefaultLogger Handler, and even if I set a custom loghandler instance through client option, it does not set the Logger property to the custom instance, but it sets the property Logger.Loggersink to the handler, which makes the LoggerSink to be the last set logger sink for all the AblyRealTime instance.

https://github.com/ably/ably-dotnet/blob/d7abbdea6869dca81d76fa2fa9f82b3323439f57/src/IO.Ably.Shared/AblyRealtime.cs#L63

### LINKED ISSUE #1253: Remove reflection from IOC class
- Remove `static` constructor from `IOC` class.
- Add proper initializer with `new Platform` constructor.
- Add configured logging for the same.

sacOO7:
- Fixed #1250 
- Fixed #1253


sacOO7:
Seems this PR is going to be a bit bigger, because at multiple passes original instance of logger is not being passed : (

Issue #1143, opened 2022-04-07
Address JSON serialisation


AndyNicks:
From conversation with a customer:
the root cause of the problem is that during the serialization of the ProtocolMessage the ably client lib is using a serializer created by this method:
        private static JsonSerializer GetSerializer()
        {
            var settings = GetJsonSettings();

            return JsonSerializer.CreateDefault(settings);
        }
when calling JsonSerializer.CreateDefault the settings passed to the method are merged with the static JsonConvert.DefaultSettings. in our case the later are:
JsonConvert.DefaultSettings = () => new JsonSerializerSettings
{
    Converters = new List<JsonConverter>() { new StringEnumConverter() },
};

the ably client lib should be making a difference between the serialization of the user's messages (stored in ProtocolMessage.Messages property) and other protocol message's properties, so two different instances of JsonSerializerSettings are needed.

the user's message serialization settings should be customizable in the ably client and it's only a matter of contract negotiation between the publisher and the consumer so ably's lib code should not interfere with those settings
the protocol message serialization settings should remain static and not customizable, otherwise you cannot ensure the same json settings apply for serialization on the client and de-serialization on the server. those settings should not depend on JsonConvert.DefaultSettings

Workaround

var customSettings = new JsonSerializerSettings
{
    Converters = new List<JsonConverter>
    {
        new StringEnumConverter() // more converters here
    },
};
var serializedJSON = JsonConvert.SerializeObject(new { payload = $@"published at: {DateTime.UtcNow}" }, customSettings);
// Don't update JsonConvert.DefaultSettings
var result = await channel.PublishAsync(new Message
{
        Data = serializedJSON ,
        Encoding = "json",
        Id = Guid.NewGuid().ToString().ToLower(),
});

see https://ably-real-time.slack.com/archives/CURL4U2FP/p1649240863594569 for details

tomkirbygreen:
The whole area of JSON serialization in the Ably SDK has been under discussion for some time. Since JSON contributes to the public surface area of the SDK this work is waiting for the SDK to be able to take back control of its version number.

We have a number of existing issues that pertain to JSON:

[Allow to send raw jtoken/jobject instead object](https://github.com/ably/ably-dotnet/issues/321)

[Unity: Newtonsoft JSON library conflicts with version that Unity includes by default](https://github.com/ably/ably-dotnet/issues/753)

[...]

Issue #321, opened 2019-04-26
Allow to send raw jtoken/jobject instead object


mcdis:
Allow to send raw jtoken/jobject instead object in IRestChannel.PublishAsync

mattheworiordan:
Hi @mcdis. Can you elaborate on your use case please?

mcdis:
Hi @mattheworiordan!
We have got micro services arch and one type of service is targeting to collecting data in json and sending them to configured channel to configured queue system. One of available target queue system is Ably.
So message aggregation works with not typed json messages and at this moment we are converting them to string because raw jobject publishing doesn't work well.

mattheworiordan:
Thanks for the response, will get @withakay to look at soon.

marto83:
Hi @mcdis. There might be a small issue depending on which version of the library you are using. We currently bundle json.net with the ably binary in .net and xamarin to stop people having issues if they are using older version of the json.net. I don't know if exposing a jtoken class will be the same as you are using. 

Currently there is a way to pass json directly. You can package it an encoded Message object as described in this ticket: https://github.com/ably/ably-dotnet/issues/92#issuecomment-370090243

You can pass a json object to ably at the moment and you will get a JObject back on the other side. I'm sure this will work fine if the services are .net core but need to test it in .Net Framework. 

If you want to send a json string and you don't want ably to try and decode it on the way back then you can create a message with a custom encoding and the library won't touch the json string so you can choose whether to parse it or not. Here is a test showing how that works:

        public async Task SendAndReceiveJson(Protocol protocol)
        {
            var client = await GetRealtimeClient(protocol);

            await client.WaitForState(ConnectionState.Connected);

            var jobject = JObject.Parse("{ \"test\":\"test\"}");

            void PublishJson(IRealtimeChannel c, string name, JToken jsonToken)
            {
                var ablyMessage = new Message(name, jsonToken.ToString()) { Encoding = "custom" };

                c.Publish(ablyMessage);
            }

            var channel = client.Channels.Get("test");
            var awaiter = new TaskCompletionAwaiter();
            channel.Subscribe(message =>
            {
                message.Encoding.Should().Be("custom");
                message.Data.Should().BeOfType<string>();
                awaiter.Tick();
            });

            PublishJson(channel, "json", jobject);

            await awaiter.Task;
        }

Issue #92, opened 2017-02-09
Ability to set the JsonSerializerSettings


drachele:
Hi,

Would it be possible to add a setting or property to the Ably client to allow customization of the Json serializer settings?

For example, we would like to be able to use a different **ContractResolver** in order to set the property names of our model objects automatically using the **CamelCasePropertyNamesContractResolver** when they are processed by Ably.

Thanks,

Dan.

mattheworiordan:
Dan, we've discussed this type of thing in other statically typed libraries as well in fact.  It's a bit tricky because the interface we need to expose needs to be generic, so we have to be careful about implementing features that are specific to a particular trend or need by a user.

Can you explain what the problem is by not having this feature?  I'd like to have a think about the problem domain and how best we can solve it rather than evaluate a single feature request.

drachele:
Hi Matt,

We are in the process of switching to Ably from another provider. We have an existing codebase with a complex domain model that contains a _LOT_ of model classes with a _LOT_ of properties. The front-end code has been written to use JSON objects with camel-case property names. Previously, we defined which JSON resolver we wanted to use and pushed the objects to the API. This took care of the camel-casing for us when serializing and pushing to clients.

In order for it to work with Ably, we would have to iterate over our entire domain model and add the `[JsonProperty("camelCaseName")]` attribute to all of our properties, which would be extremely time-consuming. That, or we would need to rewrite the front-end code to handle the JSON without camel-casing. Again, not ideal.

If there was a way to instruct the Ably API to use camel-casing (or another naming convention) by default when serializing the data, that would be great.

Hope that helps.

Cheers,
Dan.

mattheworiordan:
Ok, thanks for the explanation Dan.

I will get the others to have a look at this, but if the front end client (which I assume is Javascript) relies on camelCase for the data payloads, and your .NET code currently uses another format, why not simply have a middleware encoder / decoder that handles receiving data from Ably and publishing data to Ably, and it itself uses a `CamelCasePropertyNamesContractResolver`.  i.e. pseudo code would be:

```javascript
class DataPortability {
   constructor(data: any) {
     this.data = data
   }
  
   forFrontend() => JsonObject {
     return this.data.changeCase('lowerCamelCase').toJson()
   }

   forBackend() => JsonObject {
     return this.data.changeCase('CamelCase').toJson()
   }
}

func publish(object) {
  channel.publish('event', DataPortability.new(object).forFrontend())
}
```

[...]

PR #1256, opened 2023-08-19
Fix readme

### LINKED ISSUE #1255: JSON Encoding issues
I've had two problems encoding C# objects into JSON with the Ably SDK.

First, when trying to encode a `TokenRequest` from `IRestClient.Auth.CreateTokenRequestObjectAsync`, by calling JSON.net in my HTTP controller. The call to `JsonConvert.SerializeObject(tokenRequest)` was returning JSON with the original field names from the `TokenRequest` class, and ignoring the `JsonProperty` attributes. So the resulting JSON had field names `KeyName`, `Ttl` rather than `keyName`, `ttl`, etc.

The issue seemed to be that version of JSON.net in my app was not able to understand the `JsonProperty` attributes from your SDK.

I solved this problem by calling the (deprecated) `CreateTokenRequestAsync` function, which returned the correct JSON (I can only guess as to who this worked).

The second issue that I hit was converting the other way - I was trying to send domain objects from my app as the payload of Ably messages. This time _my_ `JsonProperty` attributes were being ignored. I solved this problem using this trick:

https://github.com/ably/ably-dotnet/issues/92#issuecomment-370090243

Do you have thoughts on how to address these problems?

If not, it would be nice to include [this code snippet](https://github.com/ably/ably-dotnet/issues/92#issuecomment-370090243) in the docs, since it's very powerful when you want to have fine-grained control over the JSON.

Thanks

sacOO7:
- Related to #1255 
- Related to https://github.com/ably/ably-dotnet/issues/92, https://github.com/ably/ably-dotnet/issues/321 and https://github.com/ably/ably-dotnet/issues/1143

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,8 +235,9 @@ if (result.IsFailure)
 }
 ```
 
-- If publishing message as a dotnet object instance, internal newtonsoft will apply default serialization settings.
-- If received data has unwanted serialization, apply custom serialization settings externally, send it as a serialized string instead.
+- When published message is a dotnet object instance, internal newtonsoft will apply default serialization settings.
+- If received message has unwanted serialization, publish it as a serialized string instead by applying custom serialization settings externally.
+
 ```csharp
 var serializedData = JsonConvert.SerializeObject(message,
     new JsonSerializerSettings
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -235,6 +235,21 @@ if (result.IsFailure)
 }
 ```
 
+- When published message is a dotnet object instance, internal newtonsoft will apply default serialization settings.
```
owenpearson, line 238:
```suggestion
- When calling `channel.Publish` with a dotnet object instance as the message data, the library will use the Newtonsoft Json.NET library to serialize the message with default serialization settings.
```

owenpearson (changes requested):
Some suggestions for wording:

owenpearson (approved):
LGTM, thanks!

PR #1260, opened 2023-09-15
Fix ably request method param

sacOO7:
- Fix http request method as per https://github.com/ably/ably-dotnet/issues/1255#issuecomment-1721001542
- Since the external client `NewtonSoft` version is not guaranteed to be in sync with internal newtonsoft, call to the `Request` method with external JToken param will always fail.
- Updated method with string requestbody as a param instead.
- Since this is an optional param, shouldn't fail for customers upgrading to a new version.


sacOO7:
> This is a breaking change - you are changing the signature of `Ably.Request`, which is public API. I'm not sure what your policy is for such SDK changes.

The policy is releasing a major version for breaking changes as per semvar. 

Generally, we change the name of the public method and mark the old one as deprecated. Though I tried the same approach, seems we can't do it without changing the method name : (
I was pushing for keeping the old name for usability since this method fails while passing param ( JObject ), I don't think there are customers that really use it.

owenpearson (approved):
LGTM 👍

ttypic (approved):
LGTM

sacOO7:
> @sacOO7 can we get this released to Nuget please?

@fdmarc , sure we will do it soon 👍

PR #1261, opened 2023-09-25
Release/1.2.13 - Upgrade to protocol 2.0

sacOO7:
Implement no-connection-serial


ttypic (approved):
LGTM

owenpearson:
@sacOO7 I'm a bit confused by this - the name of the integration/2.0 branch/pr implies that we were releasing this as version 2.0? If there's no breaking changes then at the very least we should update the changelog and PR name to reflect that it's no longer a 2.0 change, but it looks like there are some breaking changes here so I think we need to have a discussion about whether it's appropriate to release this in a patch version

sacOO7:
> @sacOO7 I'm a bit confused by this - the name of the integration/2.0 branch/pr implies that we were releasing this as version 2.0? If there's no breaking changes then at the very least we should update the changelog and PR name to reflect that it's no longer a 2.0 change, but it looks like there are some breaking changes here so I think we need to have a discussion about whether it's appropriate to release this in a patch version

Though, we have changed lot of things internally, I don't think there are any breaking changes to the public API as such.

owenpearson:
OK as long as there are no breaking changes then it's cool to release this as a patch version 👍 I still think the changelog entries here are very unhelpful and misleading, for example I would change "integration/2.0" to "Upgrade library to protocol v2" (it would be good to change the actual PR title too). Also I think we should remove changes to the README and CI tests from the changelog since these aren't relevant to library users.

sacOO7:
> OK as long as there are no breaking changes then it's cool to release this as a patch version 👍 I still think the changelog entries here are very unhelpful and misleading, for example I would change "integration/2.0" to "Upgrade library to protocol v2" (it would be good to change the actual PR title too). Also I think we should remove changes to the README and CI tests from the changelog since these aren't relevant to library users.

Updated!

owenpearson (approved):
LGTM, thanks!

PR #1285, opened 2024-04-19
Release/1.2.15

sacOO7:
- Fixed RTL12 spec
- Fixed broken rest fallbacks
- Fixed flaky tests
- Fixed create token request


unity/Assets/Ably/Plugins/IO.Ably.dll
VeskeR, line 1: was this dll committed by mistake before or why is it removed now?

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Change Log
 
+## [1.2.15](https://github.com/ably/ably-dotnet/tree/1.2.15) (2023-04-19)
+
+[Full Changelog](https://github.com/ably/ably-dotnet/compare/1.2.14...1.2.15)
+
+**Merged pull requests:**
+
+- Fix readme and capabilities by @sacOO7 in https://github.com/ably/ably-dotnet/pull/1274
```
VeskeR, line 9: probably should write here **what** exactly was fixed. after quick look at the PR, it seems it is something related to unity readme inconsistencies?

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+- Fix Create token request by @sacOO7 in https://github.com/ably/ably-dotnet/pull/1282
+- Fix spec RTL12 by @sacOO7 in https://github.com/ably/ably-dotnet/pull/1275
```
VeskeR, line 11: In ably-js we agreed that mentioning "spec" doesn't provide any useful information to our customers, especially including spec points such as `RTL12` here. It refers to the bit of the documentation that they are not (and most likely shouldn't be) familiar with - specification docs. Could we rewrite it somehow to explain what was fixed in relation to no-connection serial?
I understand that the "spec" was mentioned before in this changelog, but it might be a good idea to avoid referencing it in the future.

VeskeR (changes requested):

unity/Assets/Ably/Plugins/IO.Ably.dll
sacOO7: It's not by mistake. It's as per release process!

sacOO7: You can see, it was there in last release too 👍

sacOO7: updated

CHANGELOG.md
sacOO7: updated

CHANGELOG.md
sacOO7: updated

unity/Assets/Ably/Plugins/IO.Ably.dll
sacOO7: It's not removed exactly, it's showing reduction in DLL size

ttypic (approved):
LGTM

unity/Assets/Ably/Plugins/IO.Ably.dll
VeskeR: oh, I see, I misunderstood the change and thought the file was deleted, but it is only changed. my bad, thank you!

VeskeR (approved):

PR #1291, opened 2024-08-13
Delete unnecessary recover error test

lmars:
The test removed in this commit expects that using an invalid recovery key leads to an ERROR protocol message being returned which thus fails the connection, but this isn't part of the spec, we make no guarantee that invalid recovery keys are rejected with an ERROR, and in fact returning a CONNECTED protocol message with the error field indicating the recovery key was invalid is a perfectly good response, which is in fact what we will start to do in the future.

The functionality that an ERROR protocol message is treated as fatal by the SDK and fails the connection is still tested elsewhere, so removing this test doesn't materially reduce the amount of test coverage.

### DIFF
--- a/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
@@ -1121,37 +1121,6 @@ namespace IO.Ably.Tests.Realtime
 
         [Theory]
         [ProtocolData]
-        [Trait("spec", "RTN16l")]
-        [Trait("spec", "RTN15c4")]
-        public async Task WithInvalidRecoverData_ShouldFailAndSetAReasonOnTheConnection(Protocol protocol)
-        {
-            var client = await GetRealtimeClient(protocol, (opts, _) =>
-            {
-                opts.Recover = "{\"connectionKey\":\"random_key\",\"msgSerial\":5,\"channelSerials\":{\"channel1\":\"98\",\"channel2\":\"32\",\"channel3\":\"09\"}}";
-                opts.AutoConnect = false;
-            });
-
-            ErrorInfo err = null;
-            client.Connection.On((args) =>
-            {
-                err = args.Reason;
-                if (args.Current == ConnectionState.Failed)
-                {
-                    ResetEvent.Set();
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
lgtm

sacOO7 (changes requested):
Btw, this test is related to spec https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c4. Are we planning to remove spec since we are deleting related integration tests?

lmars:
> Btw, this test is related to spec https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c4. Are we planning to remove spec since we are deleting related integration tests?

For clarity, here's what that spec point says:

> Any other ERROR ProtocolMessage indicating a fatal error in the connection. The server will close the transport immediately after. The client should transition to the FAILED state triggering all attached channels to transition to the FAILED state as well. Additionally the Connection#errorReason will be set should be set with the error received from Ably

That is still valid, if the server sends an ERROR message then it should be handled as this spec point describes, but the test deleted in this PR is making a connection with an invalid connection key and expecting that to trigger an ERROR which will no longer happen in the future, it will instead get a CONNECTED message with an error field and thus the test won't pass.

sacOO7:
> > Btw, this test is related to spec https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c4. Are we planning to remove spec since we are deleting related integration tests?
> 
> For clarity, here's what that spec point says:
> 
> > Any other ERROR ProtocolMessage indicating a fatal error in the connection. The server will close the transport immediately after. The client should transition to the FAILED state triggering all attached channels to transition to the FAILED state as well. Additionally the Connection#errorReason will be set should be set with the error received from Ably
> 
> That is still valid, if the server sends an ERROR message then it should be handled as this spec point describes, but the test deleted in this PR is making a connection with an invalid connection key and expecting that to trigger an ERROR which will no longer happen in the future, it will instead get a CONNECTED message with an error field and thus the test won't pass.

Then can please you suggest some sort of test for the missing spec. I mean I don't want spec implementation without related test

sacOO7 (approved):
lgtm

lmars:
@sacOO7 yes agreed, I'll follow up with a suggestion but going to merge this to avoid blocking other work, thanks.

PR #1320, opened 2025-11-18
[ECO-5646] Fix behaviour when internet connection is not available

### LINKED ISSUE #1319: Check behaviour of SDK when internet connectivity lost
When internet connectivity is lost the SDK will automatically retry to reconnect

In ably-dotnet as per attached logs it appears to retry to fallbacks even when the internet check does not pass, which potentially results in a connection to a region far away (and potential cross site resume failures) when the client may have been able to connect to their closest region

In other SDKs like ably-js https://github.com/ably/ably-js/blob/9b8de3600d7f7c6815fe260511f32e54cd96b025/src/common/lib/transport/connectionmanager.ts#L1566 

Part of the sequence to decide to use a fallback is to check the internet up url https://internet-up.ably-realtime.com/is-the-internet-up.txt and will try fallbacks if the internet connectivity check passes and doesnt if the internet connectivity check fails

It would be good to check that this process is aligned and behaves as expected compared with other sdks

[dotnet-verbose-logs-2.txt](https://github.com/user-attachments/files/23578730/dotnet-verbose-logs-2.txt)

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5646) by [Unito](https://www.unito.io)

sacOO7:
Fixed https://github.com/ably/ably-dotnet/issues/1319

## Summary by CodeRabbit

* **Bug Fixes**
  * Client now consistently prefers the default realtime host and more explicitly handles instant-retry and connection-key clearing, improving reconnection reliability.

* **Tests**
  * Added tests verifying default-host-first behavior after non-retryable errors and network-down scenarios.

* **Chores**
  * Added a test helper to simulate non-retryable disconnects and made the retryable-disconnect helper optionally wait for the Disconnected state.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
+++ b/src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
@@ -726,7 +726,17 @@ namespace IO.Ably.Realtime.Workflow
                                 State.Connection.ClearKeyAndId();
                             }
 
-                            var connectingHost = AttemptsHelpers.GetHost(State, Client.Options.FullRealtimeHost());
+                            var defaultRealtimeHost = Client.Options.FullRealtimeHost();
+
+                            // Always retry on defaultPrimaryHost first when connecting command triggered by Disconnected/Suspended state timeout.
+                            var connectingHost = defaultRealtimeHost;
+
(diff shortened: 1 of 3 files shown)


ttypic (approved):
LGTM

PR #1323, opened 2025-11-25
[ECO-5650] ClientOptions.HttpClient for better connection pool management

### LINKED ISSUE #1322: Issue: HttpClient Lifecycle Mismanagement in AblyRest and AblyHttpClient
### Summary

The current implementation of AblyRest and AblyHttpClient does not properly manage HttpClient instances, which can lead to socket exhaustion, DNS staleness, or connection leaks depending on how the library is used.

### Relevant source locations:

**AblyRest.InitializeAbly (in AblyRest.cs):**

`HttpClient = new AblyHttpClient(new AblyHttpOptions(Options));`
Does not inject any `HttpMessageHandler `.

**AblyHttpClient constructor (in AblyHttpClient.cs):**

```
internal AblyHttpClient(AblyHttpOptions options, HttpMessageHandler messageHandler = null)
{
    CreateInternalHttpClient(options.HttpRequestTimeout, messageHandler);
}
```
Provides a way to inject a message handler but one is never injected.

**CreateInternalHttpClient (in AblyHttpClient.cs):**

```
internal void CreateInternalHttpClient(TimeSpan timeout, HttpMessageHandler messageHandler)
{
    Client = messageHandler != null ? new HttpClient(messageHandler) : new HttpClient();
    Client.DefaultRequestHeaders.Add("X-Ably-Version", Defaults.ProtocolVersion);
    Client.DefaultRequestHeaders.Add(Agent.AblyAgentHeader, Agent.AblyAgentIdentifier(Options.Agents));
    Client.Timeout = timeout;
}
```

Currently, every new instance of AblyRest constructs a new HttpClient, without any shared HttpMessageHandler or connection management.

### Why This is a Problem
**1. Socket Exhaustion**

When AblyRest is created repeatedly (for example, per request, or per function invocation in cloud environments like Azure Functions or AWS Lambda), each instance creates a new HttpClient and underlying socket connections.

Since HttpClient holds open TCP connections in its internal connection pool, frequent instantiation leads to socket exhaustion and degraded performance.

This behavior can manifest as System.Net.Http.HttpRequestException: No available sockets or timeouts under load.

**2. DNS Staleness & Connection Reuse Issues**

On the other hand, reusing a single AblyRest (and thus one long-lived HttpClient) for the entire application lifetime introduces a different risk:

HttpClient caches DNS results for the lifetime of the connection.

If Ably’s endpoint IPs change (due to load balancing or DNS updates), a long-lived HttpClient may continue using stale DNS information.

This leads to intermittent connection issues or failed requests after infrastructure changes.

**3. Lack of HttpMessageHandler Management**

The constructor AblyHttpClient(AblyHttpOptions options, HttpMessageHandler messageHandler = null) allows injection of a handler, but AblyRest never provides one.

Without a shared or configurable HttpMessageHandler (e.g., via SocketsHttpHandler), the library cannot effectively reuse connections across clients.

### Proposed fix
- Provide a way to inject an application managed HttpClient instance

Rererence: https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5650) by [Unito](https://www.unito.io)

sacOO7:
- Added a way to inject external httpclient for better caching.
- Fixed https://github.com/ably/ably-dotnet/issues/1322

## Summary by CodeRabbit

* **New Features**
  * Added a public HTTP client property in configuration so callers can supply and manage their own HTTP client instance.

* **Breaking Changes**
  * HTTP client construction API updated — callers must use the new configuration-based injection.

* **Chores**
  * CI runner images updated from macOS 13 to macOS 14.

* **Tests**
  * Expanded tests for external HTTP client behavior; adjusted timing in a token-expiry test.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/IO.Ably.Shared/Http/AblyHttpClient.cs
+++ b/src/IO.Ably.Shared/Http/AblyHttpClient.cs
@@ -14,12 +14,12 @@ namespace IO.Ably
         private readonly Random _random = new Random();
         private string _realtimeConnectedFallbackHost;
 
-        internal AblyHttpClient(AblyHttpOptions options, HttpMessageHandler messageHandler = null)
+        internal AblyHttpClient(AblyHttpOptions options)
         {
             Now = options.NowFunc;
             Logger = options.Logger ?? DefaultLogger.LoggerInstance;
             Options = options;
-            CreateInternalHttpClient(options.HttpRequestTimeout, messageHandler);
+            CreateInternalHttpClient(options.HttpRequestTimeout, options.HttpClient);
             SendAsync = InternalSendAsync;
         }
 
@@ -68,11 +68,22 @@ namespace IO.Ably
             }
         }
 
-        internal void CreateInternalHttpClient(TimeSpan timeout, HttpMessageHandler messageHandler)
+        internal void CreateInternalHttpClient(TimeSpan timeout, HttpClient externalHttpClient = null)
         {
-            Client = messageHandler != null ? new HttpClient(messageHandler) : new HttpClient();
-            Client.DefaultRequestHeaders.Add("X-Ably-Version", Defaults.ProtocolVersion); // RSC7a
-            Client.DefaultRequestHeaders.Add(Agent.AblyAgentHeader, Agent.AblyAgentIdentifier(Options.Agents)); // RSC7d
+            Client = externalHttpClient ?? new HttpClient();
+
+            // RSC7a
+            if (!Client.DefaultRequestHeaders.Contains("X-Ably-Version"))
+            {
+                Client.DefaultRequestHeaders.Add("X-Ably-Version", Defaults.ProtocolVersion);
+            }
+
+            // RSC7d
+            if (!Client.DefaultRequestHeaders.Contains(Agent.AblyAgentHeader))
+            {
+                Client.DefaultRequestHeaders.Add(Agent.AblyAgentHeader, Agent.AblyAgentIdentifier(Options.Agents));
+            }
+
             Client.Timeout = timeout;
         }
 
--- a/src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
+++ b/src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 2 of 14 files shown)


src/IO.Ably.Tests.Shared/Rest/AblyHttpClientSpecs.cs
```diff
@@ -190,5 +191,88 @@ public void IsRetryableResponse_WithErrorCode_ShouldReturnExpectedValue(
...
+                // Arrange
+                var externalHttpClient = new HttpClient();
+                var options = new AblyHttpOptions { HttpClient = externalHttpClient };
+
+                // Act
+                var ablyHttpClient = new AblyHttpClient(options);
+
+                // Assert
+                ablyHttpClient.Client.DefaultRequestHeaders.Contains("X-Ably-Version").Should().BeTrue();
+                ablyHttpClient.Client.DefaultRequestHeaders.Contains("Ably-Agent").Should().BeTrue();
```
heikki-heikkinen-sympa, line 224: You could also test this as part of the behavior of the client. There are other ways of adding headers to a request, for example adding headers to an individual HttpRequestMessage. If you test that when a request is made then these headers are also passed with that request, then if the internal way of passing the headers to the request changes the test does not break as it is not assuming how the headers are passed to the request. It just cares that they are passed to the request.

sacOO7: Yes, makes sense 👍 
This is covered as a part of tests 
`WhenCallingUrl_AddsDefaultAblyAgentHeader` and `WhenCallingUrl_AddsCustomizedAblyAgentHeader` under 
the same updated test file

ttypic (approved):
LGTM

src/IO.Ably.Shared/Http/AblyHttpClient.cs
```diff
@@ -68,9 +68,9 @@ internal void SetPreferredHost(string currentHost)
             }
         }
 
-        internal void CreateInternalHttpClient(TimeSpan timeout, HttpMessageHandler messageHandler)
+        internal void CreateInternalHttpClient(TimeSpan timeout, HttpClient externalHttpClient = null)
         {
-            Client = messageHandler != null ? new HttpClient(messageHandler) : new HttpClient();
+            Client = externalHttpClient ?? new HttpClient();
             Client.DefaultRequestHeaders.Add("X-Ably-Version", Defaults.ProtocolVersion); // RSC7a
```
sacOO7: TODO - Check duplicate headers are not added for the same external client.

PR #1324, opened 2025-11-25
Fix/msgpack unity

sacOO7:
- Fix unity build script
- Fix msgpack deps

TODO 

- [ ] Update newtonsoft DLL
- [ ] Update netframework newtonsoft LICENSE


tigran-sargsyan-w (commented):
### Review

Overall this is a solid step towards making the Unity + MsgPack setup more maintainable and IL2CPP‑friendly.

**What looks good**

* New `Deps` and `LicenseHelper` helpers keep build / license logic out of the main Cake scripts and are easy to reuse.
* Explicit dependency checks with clear error messages should make missing DLL issues much easier to diagnose in CI and on dev machines.
* Unity edit/play mode tests being wired into the MsgPack setup helps ensure the Unity path stays in sync with the core library.

**Questions / suggestions**

1. In `DownloadLicenseContent` you create a new `HttpClient` per call and block on `.Result`. For build scripts this is usually fine, but it might be safer to either reuse a single `HttpClient` instance or at least document that the build requires network access to GitHub and these URLs.
2. When a dependency is missing or a license download fails, the thrown `Exception` currently stops the build (which is good), but it could be helpful to hint which script/target was running (e.g. include more context in the message) to simplify debugging.
3. Since this PR is still marked as draft and there are TODOs about updating the Newtonsoft DLL and .NET Framework license, it might be worth adding a short comment in the Cake tasks to make it obvious that those parts are intentionally left for a follow‑up commit.

Once the remaining TODO items are addressed, this looks good to me. 👍

PR #1328, opened 2026-05-26
ci: tighten GitHub Actions workflow security

emptyhammond:
Routine hardening of the workflows under `.github/workflows/` based on a [zizmor](https://docs.zizmor.sh) audit. No CI behaviour changes are intended - pinned action versions match what the floating tags currently resolve to, and the added permissions match what the workflows already do.

## Summary by CodeRabbit

* **Chores**
  * Pinned GitHub Actions steps to specific commit versions across CI workflows for more consistent builds.
  * Strengthened CI security by setting explicit minimal GitHub token permissions and disabling credential persistence in checkout steps.
  * Updated packaging/build workflows to use explicit permissions and forwarding of required secrets for safer, more controlled deployments.

### DIFF
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    uses: ably/features/.github/workflows/sdk-features.yml@main
+    uses: ably/features/.github/workflows/sdk-features.yml@6b3fc7a8ede2ebdd7a6325314f3a96c6466f1453 # main as of 2026-05-26
     with:
       repository-name: ably-dotnet
     secrets: inherit
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -14,4 +14,5 @@ jobs:
     uses: ably/features/.github/workflows/sdk-features.yml@6b3fc7a8ede2ebdd7a6325314f3a96c6466f1453 # main as of 2026-05-26
     with:
       repository-name: ably-dotnet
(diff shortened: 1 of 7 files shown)


.github/workflows/features.yml
```diff
@@ -6,9 +6,14 @@ on:
...
       - main
 
+permissions:
+  contents: read
+
 jobs:
   build:
-    uses: ably/features/.github/workflows/sdk-features.yml@main
+    # Track the live SDK feature spec on main rather than pinning.
+    uses: ably/features/.github/workflows/sdk-features.yml@main # zizmor: ignore[unpinned-uses]
```
VeskeR, line 18: `sdk-features` workflow requires permissions:
```
      deployments: write
      id-token: write
```
, but top level configured to only `contents: read`. This results in an invisible failure during workflow startup, and it's currently broken, see: https://github.com/ably/ably-dotnet/actions/runs/26446846130.

I can see the `features` workflow is now also broken for at least `ably-js` and `ably-python` since the related github actions hardening PRs have landed: [ably-js](https://github.com/ably/ably-js/actions/runs/26302842757), [ably-python](https://github.com/ably/ably-python/actions/runs/26639799574).
More SDK repos might be affected. Happy to review the fixes for them too.

VeskeR (changes requested):
Need to fix permissions `sdk-features` here and in other affected SDKs

ttypic (approved):

PR #511, opened 2021-06-03
Add heartbeat monitor

tomkirbygreen:
This PR monitors the connections heartbeat and queues a disconnected state change if the heartbeat has stopped.

PR #1331, opened 2026-09-02
Implement RTN23 idle-transport detection and the recovery spec points around it

AndyTWF:
Implements RTN23 — dead-transport detection — and the recovery-path spec points that
auditing the constant it had been misusing turned up. Detection latency drops from 120s
to ~25s.

To keep the change set bounded, this is deliberately **not** a general spec update. Nothing outside the connection-recovery paths above is touched, and the protocol version is unchanged — `protocol` stays at 6 upstream and this SDK still sends `v=2`.

Where a clause *on those paths* has since been superseded, I've brought that one clause forward rather than implement the version it replaced. That is the only reason RTN14h appears here: the pre-6.1.0 resume semantics it replaces were the direct cause of several of the bugs above, so preserving them would have meant knowingly implementing behaviour the spec has already retired. Fast-forwarding is scoped to the clauses these fixes already have their hands on, not pursued for its own sake.

[PUB-3824]

### Implemented (not previously present)

- **RTN23a** — disconnect a transport idle longer than `maxIdleInterval` + `realtimeRequestTimeout`. The monitor was measuring against `connectionStateTtl`.
- **RTN23b** — send `heartbeats=true` so protocol HEARTBEATs are contractual rather than an undocumented server default.
- **CD2h** — `maxIdleInterval` parsed from `connectionDetails`; previously never read.
- **TO3l11** — `realtimeRequestTimeout` is now a public, validated `ClientOptions` property; it was `internal`.

### Corrected

- **RTN15c6 / RTN15c7 / RTN16f** — continuation is now detected positively; a connection cleared per RTN15g reconnects with a new `connectionId` and no error, which the old test read as a continuation.
- **RTN19a1 / RTN19a2** — serials kept on a successful resume, renumbered on a failed one, and publish order preserved across the requeue.
- **RTN7b / RTN7d / RTN7e** — RTL6c2 queue failed alongside RTL6c1; failures now report the reason for the state change.
- **RTN24** — a connection update no longer churns channels (was 4 spurious protocol messages per reauth).
- **RTL3d / RTL3d1** — reattach unconditional on entering CONNECTED, and ordered before external listeners.
- **RTL3b / RTL3c / RTN11b / RTP5a** — channels detach when the connection goes away, so presence members no longer survive a close.
- **RTL5l** — a channel detaches immediately whenever the connection is not CONNECTED.
- **RTL11** — queued presence messages fail with an `ErrorInfo` rather than a bare exception.
- **RTL15b2** — `channelSerial` no longer cleared on SUSPENDED.
- **RTN14d** — `retryIn` reports the delay actually waited.
- **RTN14e** — the suspend deadline is evaluated on every path into DISCONNECTED, including the token/auth paths that never suspended.
- **RTB1 / RTB1a / RTB1b** — backoff coefficients, jitter and the `connectionStateTtl` clamp.
- **RTN15a / RTN15h3** — immediate reconnect for a non-token DISCONNECTED.
- **RTN17 / RTN17j** — *which* attempts consult the fallback domains: now all of them, timer-driven ones included. Previously host selection was skipped for those, so once the immediate retry budget was spent every attempt was pinned to the primary and a client whose primary datacenter was down could never reach a fallback at all. The immediate retry is also bounded (was ~65 attempts/sec) and connectivity is checked once per attempt rather than twice.
- **RTN17j** — the connectivity probe is now scoped to what the clause covers, "the use of an alternative host", and taken only when the chosen host is a fallback. It used to gate the immediate retry itself, so a reconnect that was going to stay on the primary waited out a probe nothing asked for, and a probe failing while the realtime endpoint was fine cancelled the reconnect outright — up to `disconnectedRetryTimeout` of downtime where RTN15h3 says reconnect now. The token-renewal path took no probe at all and could open a transport against another datacenter on the strength of an earlier failure; both paths now share one gated decision. ably-js draws the line in the same place.
- **RTN17i** — separately, *what counts as* a fallback-worthy failure has narrowed. A transport dropping out of CONNECTED is no longer grounds to move off the primary domain, so the first reconnect after a transient network blip returns to the primary instead of a fallback datacenter — RTN17i requires the primary be preferred "even if a previous connection attempt to that endpoint has failed", and RTN17f admits an exception only through RSC15l1 ("host unresolvable or unreachable"), which describes an attempt that never landed. A drop out of CONNECTED says the opposite: the host answered a moment ago. The fallbacks are deferred by one attempt, not lost — if the datacenter really has gone, that reconnect fails at connect time and the attempt after it moves off the primary. A server-sent `DISCONNECTED` carrying a 500–504 status is untouched by this, since RTN17f1 makes that fallback-worthy however it arrives.
- **RTN11d** — full reinitialisation on `connect()` from CLOSED, FAILED or CLOSING.
- **RTN12b** — close timeout is `realtimeRequestTimeout`, not a hardcoded 1s.
- **RTN8d / RTN9d** — key and id cleared only in the terminal states the clauses name, cleared *before* the state change is emitted, and cleared even when a transition throws.
- **RTN21** — `connectionDetails` overrides only the attributes it carries.
- **RTN24 / RTN19a** — an `UPDATE` (a reauth, typically) no longer redrives the message queues. RTN19a is scoped to "when a transport is disconnected for any reason", and an update disconnects nothing: the transport that will ACK the in-flight messages is the one they went out on, so resending put a duplicate of each on the wire for Ably to discard by `msgSerial`. Invisible to callers, but this PR reworks exactly that path.
- **RSA4c / RSA4c1** — the auth callback invocation is bounded, and `cause` is set per the clause.

### Adopted from specification 6.1.0

- **RTN14h**, replacing **RTN15g** — the client always attempts a resume and lets the server decide whether continuity survives, rather than discarding its connection state once `connectionStateTtl` has passed. Verified against the live sandbox: a resume is still honoured well beyond the advertised TTL, so the old gate was throwing away resumes the server would have accepted.
- **RTN27c** — a reconnect from `DISCONNECTED` is now genuinely a resume attempt; clearing the key on every failed attempt had made that false from the second attempt onwards.
- **RTN15g1 / RTN15g2 / RTN15g3** and **RTL4j / RTL4j1 / RTL4j2** — deleted at 6.1.0. The TTL freshness check and `ATTACH_RESUME` go with them; `channelSerial` (RTL4c1, kept across a suspend by RTL15b2) already carries the continuity signal.
- **RTN16g3**, replacing **RTN16g2** — `createRecoveryKey()` now returns a key while `SUSPENDED`, where it previously returned empty. RTN8d/RTN9d keep the `connectionKey` through `SUSPENDED` because RTN14h always attempts a resume, so the connection is still recoverable there and the key has to be available to hand over. RTN16g3 comes from [ably/specification#511](https://github.com/ably/specification/pull/511), which tombstones RTN16g2 and drops `SUSPENDED` from the states where `createRecoveryKey()` returns null; that merged while this branch was in review. Written before it merged because it is the direct consequence of RTN14h above, and splitting the two would mean two behaviour changes for callers instead of one. ably-js has behaved this way since 2.27.0.

#### A clause the 6.1.0 edit had missed, since corrected

- **RTN27d** used to call `SUSPENDED` a state whose next connect attempt is "a clean connection (not a resume attempt)", which RTN14h, RTN8d, RTN9d and DF1a contradict between them — spec 6.1.0 moved resumability to the server without touching RTN27. [ably/specification#511](https://github.com/ably/specification/pull/511) has since amended the clause in place and dropped that sentence, so this PR **follows** RTN27d rather than deviating from it. Amending in place is the right shape here: RTN27 declares its states "mutually exclusive and exhaustive", so tombstoning the clause would have left `SUSPENDED` undescribed.

### Behaviour changes

No public API changed, but these are visible to callers:

- **`authCallback` / `authUrl` are now bounded by `realtimeRequestTimeout` (default 10s)** — previously unbounded. A callback that took longer and still succeeded will now fail the auth attempt (RSA4c). The only item here that can break a working deployment.
- **`HeartbeatMonitorDelay` rejects values below 1ms** — previously accepted, but `0` hot-looped and negatives silently killed idle detection, so no working configuration is affected.
- **`Connection.Key` and `Id` read as null during a `CLOSED`/`FAILED` state-change callback** — they are now cleared before the event is emitted (RTN8d/RTN9d), and retained through `SUSPENDED`.
- **`connectionId` no longer changes after a long disconnection** — the client always attempts a resume now (RTN14h). Use `ChannelStateChangedEventArgs.Resumed` to detect a fresh connection.
- **`createRecoveryKey()` returns a key while `SUSPENDED`**, where it previously returned empty (RTN16g3).
- **Some `ErrorInfo` messages and codes changed** — affects code that matches on them.

[PUB-3824]: https://ably.atlassian.net/browse/PUB-3824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added configurable realtime request timeout validation.
  * Added server-provided idle intervals and protocol heartbeats, with caller overrides preserved.
  * Improved connection recovery, channel reattachment, serial continuity, and suspended-state recovery.
  * Added immediate reconnect handling and fallback connectivity checks.

* **Bug Fixes**
  * Auth callbacks now respect request timeouts and report underlying failure details.
  * Queued callbacks continue processing after an application callback error.
  * Retry and close timers now reflect configured timeout values accurately.
  * Improved connection and channel state handling during failures and transitions.


src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -559,6 +632,79 @@ ErrorInfo GetErrorInfoFromTransportException(Exception ex, ErrorInfo @default)
...
+        private bool ProtocolHeartbeatsNotRequestedByCaller()
+        {
+            // Answered once - TransportParams is fixed at client construction.
+            if (_protocolHeartbeatsNotRequested.HasValue)
+            {
+                return _protocolHeartbeatsNotRequested.Value;
+            }
+
+            _protocolHeartbeatsNotRequested = ComputeProtocolHeartbeatsNotRequested();
+            return _protocolHeartbeatsNotRequested.Value;
```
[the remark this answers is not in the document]
AndyTWF: Done

src/IO.Ably.Tests.Shared/Realtime/ConnectionSandBoxSpecs.cs
```diff
@@ -960,12 +960,16 @@ public async Task WhenDisconnectedMessageContainsTokenError_IfTokenRenewFails_Sh
...
-        [Trait("spec", "RTN15g")]
-        [Trait("spec", "RTN15g1")]
-        // "RTN15g2" It can't implement that spec item because RTN23a is not even implemented
-        [Trait("spec", "RTN15g3")]
-        public async Task WhenDisconnectedPastTTL_ShouldNotResume_ShouldClearConnectionStateAndAttemptNewConnection(Protocol protocol)
+        [Trait("spec", "RTN14h")]
+        [Trait("spec", "RTN8d")]
+        [Trait("spec", "RTN9d")]
+        [Trait("spec", "RTN15c6")]
+        public async Task WhenDisconnectedPastTTL_ShouldStillResume_AndReattachChannels(Protocol protocol)
```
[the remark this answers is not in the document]
AndyTWF: Done

src/IO.Ably.Shared/ClientOptions.cs
```diff
@@ -295,6 +295,51 @@ public bool UseBinaryProtocol
...
+                // firing at all. Timeout.InfiniteTimeSpan is -1ms, so the lower bound catches it -
+                // which matters, because both Task.Delay and System.Threading.Timer accept it as
+                // genuine infinity and it would disable RTN14c, RTN12b, RTL4f, RTL5f and RSA4c.
+                //
+                // The upper bound is the tightest limit across the sinks this value reaches on every
+                // framework shipped: Task.Delay allows uint.MaxValue - 1 ms on .NET 6+ but only
+                // Int32.MaxValue on .NET Framework, Mono and Xamarin, and CountdownTimer casts to
+                // int for System.Threading.Timer. So Int32.MaxValue ms - a bound on the arithmetic,
+                // not a supported configuration, which is why the message names the useful range.
+                if (value <= TimeSpan.Zero || value.TotalMilliseconds > int.MaxValue)
```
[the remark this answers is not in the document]
AndyTWF: Done

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -289,6 +360,17 @@ private async Task<RealtimeCommand> ProcessCommandInner(RealtimeCommand command)
             switch (command)
             {
                 case ConnectCommand _:
+
+                    // RTN11d - connect() out of CLOSED or FAILED starts afresh. The channel half,
+                    // back to INITIALIZED with errorReason unset, is done per channel by the command
+                    // queued below; Id and Key are already emptied on entering CLOSED or FAILED.
+                    if (State.Connection.State == ConnectionState.Closed ||
+                        State.Connection.State == ConnectionState.Failed)
```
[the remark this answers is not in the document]
AndyTWF: @sacOO7 I don't think this one is correct.

The spec points are as follows:

```
(RTN11b) If the state is CLOSING, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the CLOSED ProtocolMessage arrives for the old connection, it doesn’t affect the new one. Additionally, the client should ensure that all channels first transition to DETACHED, following RTL3b, and then reinitialize channels per RTN11d.

(RTN11d) If the state is CLOSED or FAILED, transitions all the channels to INITIALIZED and unsets their RealtimeChannel.errorReason, clear all internal connection data (including in particular Connection.errorReason) and resets the msgSerial to 0
```

RTN11b makes no particular reference to resetting the error reason or the message serial. It only references RTN11d insofar as reinitializing the channels. 

As it turns out, we don't lose anything by this exclusion because entering closing empties the connection ID in key for RTN8d and RTN9d. When the next connect arrives, there's no ID that needs to be matched, and the message serial is reset. This happens at `connected`, which is, if you take the spec literally, where it should happen. Ably JS is not a good reference here, as it also doesn't appear to follow the spec in any useful way.

Admittedly, though, the comment in the code is misleading, so I've adjusted the comment to be closer to what the spec actually says.

src/IO.Ably.Shared/Transport/ConnectionAttemptsInfo.cs
```diff
@@ -82,10 +95,16 @@ private void RecordAttemptFailure(ConnectionState state, ErrorInfo error)
 
         private void RecordAttemptFailure(ConnectionState state, Exception ex)
```
sacOO7: **[MAJOR] RTN17i — the first reconnect after a socket-error transport drop now targets a fallback host instead of the primary**

This empty-collection fix is genuinely needed — it seeds `FirstAttempt` so the RTN14e suspend clock starts on a drop out of CONNECTED — but it has a host-selection side effect:

1. A transport that drops with an **Exception** (an ordinary socket error) lands here with the attempts collection empty (entering CONNECTED cleared it), and is now recorded.
2. `AttemptFailedState.ShouldUseFallback()` is true for *any* exception, so `DisconnectedCount()` becomes 1 immediately.
3. On the instant reconnect (RTN15a/RTN15h3), `GetHost` returns `GetFallbackHost(1)` = `FallbackHosts[0]`. The RTN17a early-return doesn't apply (current host is the primary), and the CONNECTING handler accepts the fallback because connectivity was already confirmed.

Net effect: the very first resume after a transient network blip goes to a **fallback datacenter**, although the primary is most likely healthy. [RTN17i](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17i) requires every attempt to "always prefer the primary domain, even if a previous connection attempt to that endpoint has failed", and ably-js always tries the primary first, falling back only when that attempt fails ([connectionmanager.ts](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts)).

Two scoping notes from verification:
- The **RTN23a idle-monitor disconnect does NOT hit this**: `NoActivityFrom` carries HTTP 408 with no exception, and `IsRetryableStatusCode` is strictly 500–504, so that path correctly stays on the primary.
- Partly pre-existing: the `ErrorInfo` overload above already recorded on an empty collection on `main`, so 5xx server-sent DISCONNECTEDs already jumped to a fallback. This change extends the behaviour to exception drops, which the new instant-retry machinery then drives.

**Suggested fix:** keep the drop recorded for the RTN14e clock but stop it advancing RTN17 host selection — e.g. prefer the primary on the first post-drop reconnect (fallbacks only after that attempt fails), or exclude the CONNECTED-drop failure from the `DisconnectedCount()` that `GetHost` consumes while still counting it for `FirstAttempt`. Please also add a test asserting the first reconnect after an exception drop lands on the primary when fallbacks are configured — `WithNonTokenDisconnected_ShouldReconnectImmediately` asserts the `resume` param but never the host, so this isn't currently caught.

src/IO.Ably.Shared/AblyAuth.cs
```diff
@@ -305,11 +305,27 @@ public virtual async Task<TokenDetails> RequestTokenAsync(TokenParams tokenParam
...
+                    // RSA4c bounds an auth attempt by realtimeRequestTimeout. This one is awaited on
+                    // the realtime workflow's single reader thread, so an authCallback that never
+                    // returns would stall channel state, inbound messages and the idle monitor.
+                    //
+                    // Task.Run rather than awaiting the delegate directly: TimeoutAfter extends an
+                    // already-created Task, so a callback whose body runs synchronously - the most
+                    // ordinary shape in C# - would block before there is anything to bound. It
+                    // cannot cancel the callback; an abandoned one runs to completion.
+                    var callbackResult = await Task.Run(() => authOptions.AuthCallback(tokenParams))
+                        .TimeoutAfter(Options.RealtimeRequestTimeout, null);
```
sacOO7, line 317: **[MINOR] RSA4c — only the `authCallback` delegate is bounded by `realtimeRequestTimeout`; the `authUrl` branch and the final token exchange are not**

This wrapper correctly bounds the callback delegate (including a synchronously-blocking one, via `Task.Run`), but two legs of a token acquisition remain bounded only by the HTTP stack:

- the `authUrl` request (`CallAuthUrl` → `_rest.ExecuteRequest`, line 525), and
- the final `TokenRequest` → `TokenDetails` exchange with Ably (line 446) — which runs on **every** branch, including `authCallback` when it returns a `TokenRequest`.

[RSA4c](https://sdk.ably.com/builds/ably/specification/main/features/#RSA4c) bounds the auth *attempt* (authCallback or authUrl) by `realtimeRequestTimeout`; ably-js wraps the whole acquisition — its timeout message is literally "Token request via authCallback/authUrl did not complete within the configured timeout" ([auth.ts#L657](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/client/auth.ts#L657)). With .NET defaults, `HttpMaxRetryDuration` (~15s) can exceed a 10s `RealtimeRequestTimeout`, so a caller's configured bound silently doesn't apply to those legs.

Impact is small (both legs are fully async — no workflow thread is blocked), so this can be a follow-up. **Suggested:** either extend `.TimeoutAfter(Options.RealtimeRequestTimeout, …)` to the authUrl call and the token exchange, or document the intentional HTTP-stack bound. Either way a test pinning the chosen behaviour would help — currently only the callback delegate's timeout is tested.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -977,26 +1201,42 @@ private void UpdateStateAndNotifyConnection(ConnectionStateBase newState)
...
-                foreach (var messageAndCallback in State.WaitingForAck)
+                // RTN19a2 - the same connection is still expecting the serials these messages were
+                // originally given, so resend them unchanged and leave them awaiting their ACK.
+                foreach (var message in State.WaitingForAck.Select(x => x.Message))
                 {
-                    State.PendingMessages.Add(new MessageAndCallback(
-                        messageAndCallback.Message,
-                        messageAndCallback.Callback,
-                        messageAndCallback.Logger));
+                    ConnectionManager.SendToTransport(message);
```
sacOO7: **[MINOR] RTN24 / RTN19a2 — a mid-connection UPDATE re-sends every in-flight message on the live transport**

On an RTN24 update (e.g. reauth via `authorize()`), `connectionContinues` is true via `cmd.IsUpdate`, so this branch re-sends all `WaitingForAck` messages — on the *same live transport* that is still going to ACK the originals. RTN19a is scoped to transport loss; a reauth swaps nothing.

Server-side `msgSerial` dedup makes this invisible to users, and the old code had the same effect, so it's not a regression — but this PR reworked exactly this path, and ably-js never re-drains on a CONNECTED-while-connected (its update path only merges `connectionDetails` and emits `update`, [connectionmanager.ts#L658-L666](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts#L658-L666)).

**Suggested:** skip the resend for updates — e.g. `if (connectionContinues && cmd.IsUpdate == false)` around the resend loop, leaving the serial-preservation semantics for real resumes intact.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -719,22 +813,39 @@ private async Task<RealtimeCommand> HandleSetStateCommand(RealtimeCommand comman
...
+                            // host, so stay on the primary rather than working through fallbacks that
+                            // cannot answer either.
+                            //
+                            // The answer is carried on the command, so it cannot outlive the decision
+                            // it was taken for or be picked up by a CONNECTING another path queued.
+                            // Held on the workflow it would have no bound, because the command loop
+                            // abandons a nested batch once its depth guard trips.
+                            //
+                            // Note HandleConnectingTokenError reaches CreateTransport through
+                            // AttemptANewConnection without a check - an RTN17j hole this does not
```
sacOO7: **[MINOR] RTN17j — the token-renewal reconnect can reach a fallback host with no connectivity check (the hole this comment acknowledges)**

Appreciate the honesty of this comment — filing it so the hole is tracked rather than just noted. `HandleConnectingTokenErrorCommand` → `AttemptANewConnection` calls `GetHost` and then `CreateTransport` directly, bypassing the `CanConnectToAbly` gate this handler applies before accepting an alternative host. If a prior fallback-eligible DISCONNECTED is on record when a token error arrives before the first renewal, the transport is built against a fallback with no [RTN17j](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17j) connectivity check.

Reachability is narrow, so Minor. **Suggested:** route `AttemptANewConnection` through the same connectivity-gated host decision (or have it queue a `SetConnectingStateCommand`), now or as an immediate follow-up.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -841,9 +1003,42 @@ async Task<bool> CheckInstantRetryFlag()
...
+                            var reconnectImmediately = cmd.Exception != null
+                                                       || (cmd.Error != null && cmd.Error.IsRetryableStatusCode())
+                                                       || (State.Connection.State == ConnectionState.Connected
+                                                           && cmd.Error?.IsTokenError != true);
+
+                            if (reconnectImmediately)
                             {
-                                return await Client.RestClient.CanConnectToAbly();
+                                // Remembered so the CONNECTING behind this command does not repeat it.
+                                connectivityAnswer = await Client.RestClient.CanConnectToAbly();
```
sacOO7: **[MINOR] RTN15a/RTN15h3 vs RTN17j scope — the "immediate" reconnect pays an HTTP connectivity round-trip even when the target is the primary**

`reconnectImmediately` gates the instant retry on `CanConnectToAbly()` unconditionally, but [RTN17j](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17j) scopes the connectivity check to *the use of an alternative host*. Two consequences:

1. When online, every "immediate" resume first waits for the connectivity probe (up to `HttpOpenTimeout`) before reconnecting.
2. If the probe transiently fails while the realtime endpoint is fine, the immediate resume is skipped entirely and deferred to the RTB1 timer — up to `disconnectedRetryTimeout` (15s) of avoidable downtime.

ably-js schedules its immediate reconnect on next-tick with no precheck for the primary attempt ([connectionmanager.ts#L1194-L1217](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts#L1194-L1217)).

**Suggested:** grant the immediate retry to the primary unconditionally and require the connectivity check only when the chosen candidate is a fallback. This composes naturally with the RTN17i fix suggested on `ConnectionAttemptsInfo.cs` — after that fix the instant-retry candidate is the primary anyway, and the probe moves to the fallback decision where RTN17j wants it.

src/IO.Ably.Shared/Transport/States/Connection/ConnectionDisconnectedState.cs
```diff
@@ -49,14 +49,74 @@ public override void AbortTimer()
...
+            _timer.Start(retryIn, OnTimeOut);
+        }
+
+        /// <summary>
+        /// RTN14e requires the move to SUSPENDED once the connection state ttl has elapsed, and that
+        /// decision is taken when an attempt fails. Sleeping past the deadline would delay SUSPENDED
+        /// by however long we slept, unbounded because disconnectedRetryTimeout is a client option.
+        /// Waking at the deadline lets the attempt fail there and be converted, keeping one timer.
+        /// </summary>
+        private TimeSpan ClampToStateTtl(TimeSpan retryIn, RealtimeState state)
```
sacOO7, line 86: **[MINOR — tests] RTN14d/RTN14e — the clamp-to-deadline path of `ClampToStateTtl` is never exercised**

The arithmetic here checks out (both operands clamped before subtracting, overflow guarded, clamps only when `0 < remaining < retryIn`), but no test reaches the clamp-return branch: the new `DisconnectedStateSpecs` never seed `AttemptsInfo.FirstAttempt`, so this method always takes the `firstAttempt == null` early return (an integration test seeds it, but never asserts the clamped value). The negative-elapsed (clock step-back) and near-`TimeSpan.MaxValue` guards are also untested.

Since punctual suspension is one of the behaviours this PR fixes, a regression here would pass CI silently. **Suggested:** a `DisconnectedStateSpecs` case seeding `FirstAttempt` close to `ConnectionStateTtl` asserting `RetryIn` equals the remaining time to the deadline, plus a clock-step-back case.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -977,26 +1201,42 @@ private void UpdateStateAndNotifyConnection(ConnectionStateBase newState)
...
+                //
+                // WaitingForAck is cleared because SendMessage re-registers each message as it
+                // goes; stale entries would hold serials of the old sequence that the next ACK also
+                // matches, running their callbacks twice.
+                //
+                // Inserted at the front, not appended: PendingMessages is the RTL6c2 queue and
+                // already holds anything published while disconnected, which happened *after* these.
+                // Appending would give the newer messages the lower serials and reverse publish
+                // order. ably-js prepends for the same reason.
+                State.PendingMessages.InsertRange(
```
sacOO7: **[MINOR — tests] RTN19a1/RTN19a2 — the prepend ordering is never tested with both queues populated**

This `InsertRange(0, …)` is load-bearing: it's what guarantees a renumbered in-flight message keeps a *lower* serial than a message published while disconnected, preserving publish order across a failed resume. But the fixtures only ever populate one queue (`GetClientWithOneUnackedMessage` leaves `PendingMessages` empty), so the relative order is unpinned — an accidental `AddRange` would pass every existing test while reversing publish order.

**Suggested:** one test with an in-flight message plus a message queued while DISCONNECTED, forcing a failed resume, asserting the on-wire serials are `[in-flight = 0, queued = 1]`.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -599,12 +682,31 @@ private void HandleConnectedCommand(SetConnectedStateCommand cmd)
...
+            //    success is judged by the absence of an error.
+            //
+            // Broader than testing for an error on the message: a resume the server refuses is
+            // answered with a new connectionId and, often, no error at all - exactly the case that
+            // most needs a new sequence.
+            //
+            // Must be evaluated before Update below, which overwrites the id being compared, and
+            // before Options.Recover is cleared for RTN16k.
+            var isRecoverAttempt = Client.Options.Recover.IsNotEmpty();
+            var connectionContinues = cmd.IsUpdate ||
```
sacOO7, line 769: **[MINOR — tests] RTN15c6 — "same connectionId WITH an error" resume is correct but unpinned**

The third disjunct treats a resume that returns the *same* `connectionId` together with a non-fatal error as a continuation (serial preserved, error surfaced). That's the right call and matches ably-js (`connIdChanged` false ⇒ preserve, [connectionmanager.ts#L823-L826](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts#L823-L826)) — but no test covers it, and this discriminator is exactly the logic the old code got wrong (it keyed continuity on the error's presence). A regression flipping this back would go unnoticed.

**Suggested:** a resume test with same id + non-fatal error asserting `MessageSerial` is preserved and the error lands in `Connection.ErrorReason`.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -108,7 +110,7 @@ public void Start()
                 {
                     while (true)
                     {
-                        QueueCommand(HeartbeatMonitorCommand.Create(Connection.ConfirmedAliveAt, Connection.ConnectionStateTtl).TriggeredBy("AblyRealtime.HeartbeatMonitor()"));
+                        QueueCommand(HeartbeatMonitorCommand.Create(Now()).TriggeredBy("AblyRealtime.HeartbeatMonitor()"));
```
sacOO7: **[MINOR] RTN23a — `HeartbeatMonitorDelay` is unvalidated and this loop is unguarded; a bad value silently disables idle detection for the client's lifetime**

`HeartbeatMonitorDelay` (`ClientOptions.cs:440`, public `int`, default 1000) now directly drives the RTN23a monitor's granularity, and this loop is a fire-and-forget `Task.Run` with no try/catch:

- a **negative** value makes `Task.Delay` throw `ArgumentOutOfRangeException`, faulting the task unobserved — idle detection is dead for the life of the client with no log line;
- `0` produces a hot loop queueing a command per scheduler tick;
- very large values silently stretch detection latency beyond the RTN23a promise.

The property is pre-existing, but this PR makes it load-bearing for RTN23a — and it validated the sibling knob (`RealtimeRequestTimeout`) against exactly this hazard class, so the same reasoning applies here.

**Suggested:** validate the setter the same way (≥ 1ms, ≤ `int.MaxValue` ms — or clamp), and/or wrap this loop body in try/catch with a `Logger.Error` so a faulted monitor is at least visible. A test pinning "a negative delay does not silently disable idle detection" would close it.

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -664,8 +707,11 @@ private void HandleStateChange(ChannelState state, ErrorInfo error, ProtocolMess
                 Logger.Debug($"HandleStateChange state change from {State} to {state}");
             }
 
-            // RTP5a1
-            if (state == ChannelState.Detached || state == ChannelState.Suspended || state == ChannelState.Failed)
+            // RTL15b2 - "If the channel enters the DETACHED or FAILED state, it must clear its
+            // channelSerial. (Unlike previous spec versions, it must not clear it when entering the
+            // SUSPENDED state)." Retaining it in SUSPENDED lets a channel suspended by an RTL4f
+            // attach timeout carry its serial on the reattach RTL4c1 asks for.
+            if (state == ChannelState.Detached || state == ChannelState.Failed)
```
sacOO7, line 714: **[MINOR — tests] RTL15b2/RTL4c1/RTL4j — `channelSerial` is now the *sole* continuity signal, but none of its three invariants has a direct test**

This line is correct (serial cleared only on DETACHED/FAILED, retained through SUSPENDED — matching ably-js [realtimechannel.ts#L917-L923](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/client/realtimechannel.ts#L917-L923)). But with RTN15g and RTL4j deleted by spec 6.1.0, message continuity across a resume now rides entirely on `channelSerial`, and none of the three invariants is directly asserted:

1. a channel entering SUSPENDED retains its serial, DETACHED/FAILED clear it ([RTL15b2](https://sdk.ably.com/builds/ably/specification/main/features/#RTL15b2));
2. the outgoing ATTACH carries `Properties.ChannelSerial` ([RTL4c1](https://sdk.ably.com/builds/ably/specification/main/features/#RTL4c1));
3. `ATTACH_RESUME` is set on no outgoing message (RTL4j tombstone).

All three were verified correct in this PR by inspection — but a regression in any of them silently loses continuity across resumes and nothing in CI would notice. **Suggested:** three small unit tests pinning the serial lifecycle across channel states, the ATTACH payload, and the absent flag.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -797,11 +925,23 @@ ErrorInfo TransformIfTokenErrorAndNotRetryable()
...
+                        // DISCONNECTED converges. Checking only the two connection-attempt failure
+                        // handlers misses the token and auth retry paths, which do not pass through
+                        // either: a client whose token source keeps failing would loop CONNECTING and
+                        // DISCONNECTED indefinitely without ever suspending.
+                        //
+                        // Ordered before CheckInstantRetryFlag so that suspending beats retrying.
+                        //
+                        // SkipAttach is excluded: the caller has already queued the next command, so
+                        // diverting would emit SUSPENDED and then immediately CONNECTING.
+                        if (cmd.SkipAttach == false && State.ShouldSuspend(Now))
```
sacOO7, line 963: **[QUESTION] RTN14e — the `skipAttach` path bypasses the suspend check; is that intended?**

`RetryAuthCommand` creates a `SetDisconnectedStateCommand` with `skipAttach = (State == Connecting)`, and this guard skips `ShouldSuspend` whenever `SkipAttach` is set. In principle a reauth storm while CONNECTING could keep hopping DISCONNECTED→CONNECTING past the `connectionStateTtl` deadline without ever emitting SUSPENDED.

Risk looks low — these are user/server-driven reauth events, and the next real connection failure re-enters this handler with `SkipAttach == false` and converts — and the comment above explains the mechanical reason (the caller has already queued the next command). Just confirming the deadline-skip is a considered trade-off rather than incidental; if a reauth loop past the TTL is deemed possible, evaluating `ShouldSuspend` even here (converting to SUSPENDED before the queued CONNECTING) would close it.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeState.cs
```diff
@@ -72,19 +90,42 @@ public ConnectionStateChange UpdateState(ConnectionStateBase state, ILogger logg
...
-            public void Update(ConnectionInfo info)
+            public void Update(ConnectionInfo info, bool isUpdate)
             {
+                // Guarded differently on purpose. connectionId is a top-level field and always
+                // meaningful, per RTN8b. connectionKey lives inside connectionDetails, and RTN21
+                // scopes an override to "the attributes within ConnectionDetails" - so a CONNECTED
+                // carrying none overrides no key, and emptying it would leave a live connection with
+                // nothing to resume with under RTN15b. Clearing the key belongs to ClearKey and
+                // ClearKeyAndId, at the points that mean it.
                 Id = info.ConnectionId;
```
sacOO7, line 101: **[OPTIONAL] RTN21 — `Id` is assigned unconditionally on an update while `Key` is guarded; consider mirroring the guard**

On an RTN24 update, `Id = info.ConnectionId` runs unconditionally while `Key` (below) is guarded by `IsNotEmpty()`. RTN8b guarantees the id's presence on a CONNECTED, and ably-js assigns it unconditionally too ([connectionmanager.ts#L841](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts#L841)) — so this is benign and not a deviation. Flagging only because the asymmetry looks accidental to a reader: a one-line non-empty guard (or a comment stating why the id needs none) would make it clearly deliberate.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -251,31 +254,98 @@ internal async Task<IEnumerable<RealtimeCommand>> ProcessCommand(RealtimeCommand
...
+
+            // A window that cannot be represented can never elapse. maxIdleInterval is unbounded off
+            // the wire, and an OverflowException here would be logged and dropped by the command
+            // loop, silently killing detection for the life of the connection.
+            if (maxIdleInterval.Value >= TimeSpan.MaxValue - Client.Options.RealtimeRequestTimeout)
+            {
+                return Enumerable.Empty<RealtimeCommand>();
+            }
+
+            var allowedIdleTime = maxIdleInterval.Value + Client.Options.RealtimeRequestTimeout;
```
sacOO7, line 349: **[OPTIONAL] RTN23a — the `RealtimeRequestTimeout` term is read live while its companions are per-transport snapshots**

`MaxIdleInterval` and `ProtocolHeartbeatsRequested` are deliberately snapshotted per transport (the comments above explain why), but the `+ RealtimeRequestTimeout` slack term is read from `ClientOptions` at every tick — so a caller mutating it mid-connection changes the threshold for the currently-open transport while the heartbeats decision and the server's promise do not move.

Harmless (the setter validates the range, and the term is only additive slack), so no change needed — a one-line comment noting the term is intentionally read live would just remove the ambiguity the other two comments' carefulness creates.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -409,20 +512,10 @@ async Task AttemptANewConnection()
...
-                                cmd.Error ?? ErrorInfo.ReasonDisconnected,
-                                clearConnectionKey: true)
-                            .TriggeredBy(cmd);
-                    }
+
+                    // Suspending is decided in the SetDisconnectedStateCommand handler, for every path.
+                    return SetDisconnectedStateCommand.Create(cmd.Error ?? ErrorInfo.ReasonDisconnected)
+                        .TriggeredBy(cmd);
 
                 case HandleConnectingErrorCommand cmd:
```
sacOO7, line 551: **[NIT] Dead code — `HandleConnectingErrorCommand.ClearConnectionKey` is now never read and never set**

This PR removed the `clearConnectionKey` plumbing from the Set{Disconnected,Suspended}StateCommand paths (correctly, per RTN27c/RTN14h), but `HandleConnectingErrorCommand` in `RealtimeCommands.cs` still carries a `ClearConnectionKey` property, ctor arg and `Create` parameter — no handler reads it, and all call sites use the default `false`. Worth trimming so the retired mechanism doesn't look load-bearing to the next reader.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
```diff
@@ -719,22 +813,39 @@ private async Task<RealtimeCommand> HandleSetStateCommand(RealtimeCommand comman
...
-                            // Always retry on defaultPrimaryHost first when connecting command triggered by Disconnected/Suspended state timeout.
+                            // RTN17 - every attempt considers a fallback, including the timer driven
+                            // ones. Excluding those would lock a client out once the immediate retry
+                            // budget is spent, since every remaining attempt is timer driven and so
+                            // pinned to the primary.
+                            //
+                            // Asked speculatively, which costs nothing: GetHost only reads state, and
+                            // RTN17i is its job - it returns to the primary whenever the last host
+                            // was a fallback.
+                            var candidateHost = AttemptsHelpers.GetHost(State, defaultRealtimeHost);
```
sacOO7: **[NIT] Redundant clear — `SetConnectingStateCommand.ClearConnectionKey` is a no-op now**

The `if (cmd.ClearConnectionKey) State.Connection.ClearKey();` block just above (lines 811–814) can only be reached with `true` from `ConnectionClosingState.Connect()` — but entering CLOSING already ran `ClearKeyAndId()` in the SetClosingStateCommand handler, before that `Connect()` can be processed (single-reader ordering), so the key is always already empty here. Candidate for removal together with the dead `HandleConnectingErrorCommand.ClearConnectionKey` flagged above.

sacOO7 (commented):
## Clause-by-clause review — RTN23 + recovery paths

Full review of the PR against the spec (local 6.1.1 checkout **plus merged [ably/specification#511](https://github.com/ably/specification/pull/511)** for RTN16g3 and the amended RTN27d) and against ably-js v2.27.0 as reference where applicable.

**Verdict: approve with changes requested.** The RTN23 core (RTN23a/b, CD2h, TO3l11) is correct, matches the reference semantics, and is unusually well tested. All five forward-adopted spec points (RTN14h, RTN27c, amended RTN27d, RTN16g3, the RTN15g/RTL4j deletions) were audited as first-class requirements and are **compliant** — with #511 merged, none of them is a deviation any more; the PR description's "not yet in the published spec" / RTN27d paragraphs could be refreshed to cite the merge. Notably, every reconnect path funnels through a single choke point that writes the `resume` param iff a key is held, so RTN14h can't be missed by any path.

**17 inline comments follow**, one per finding, tagged **[MAJOR]** (1) / **[MINOR]** (9, of which 4 are test-coverage asks) / **[QUESTION]** / **[OPTIONAL]** / **[NIT]**. The one Major (fallback host chosen on the first reconnect after a socket-error drop, RTN17i) is the only thing suggested as a pre-merge fix.

src/IO.Ably.Shared/Realtime/RealtimeChannel.cs
```diff
@@ -396,18 +440,17 @@ private void Detach(Action<bool, ErrorInfo> callback, bool force, bool emitUpdat
...
-                }
-                else if (ConnectionState != ConnectionState.Failed)
+                // RTL5l - "if the connection state is anything other than CONNECTED and none of the
+                // preceding channel state conditions apply, the channel transitions immediately to
+                // the DETACHED state". Tested as anything-but-CONNECTED rather than enumerated, so
+                // DISCONNECTED cannot fall through to the RTL6c2 queue.
+                if (ConnectionState != ConnectionState.Connected)
                 {
-                    SendMessage(new ProtocolMessage(ProtocolMessage.MessageAction.Detach, Name));
+                        SetChannelState(ChannelState.Detached);
```
sacOO7: **[NIT] Stray indentation** — `SetChannelState(ChannelState.Detached);` is indented two levels deeper than its block.

src/IO.Ably.Shared/Realtime/Connection.cs
```diff
@@ -181,10 +181,13 @@ private void HandleNetworkStateChange(NetworkState state)
         /// <returns>recoveryKey.</returns>
         public string CreateRecoveryKey()
```
sacOO7, line 188: **[OPTIONAL] RTN16g3 — returns `string.Empty` where the spec (and ably-js) say `Null`; also a note that the CLOSING handling here is deliberately stricter than ably-js**

Two observations, neither blocking:

1. The now-merged [RTN16g3](https://github.com/ably/specification/pull/511) says `createRecoveryKey()` "should return `Null`" in the listed states; this returns `string.Empty` (long-standing convention in this SDK, and consumers gate on `IsNotEmpty`, so it's functionally safe). ably-js returns `null` ([connectionmanager.ts#L855-L866](https://github.com/ably/ably-pubsub-js/blob/2.27.0/src/common/lib/transport/connectionmanager.ts#L855-L866)). Worth either aligning to `null` for literal conformance or stating the empty-string convention in the docstring.

2. For the record from cross-review: excluding CLOSING here is **more** spec-compliant than ably-js, whose `createRecoveryKey` has no state check and whose `closing` state is non-terminal — so js returns a live recovery key while CLOSING, against RTN16g3's letter. Keep this as is; no parity change wanted.

src/IO.Ably.Shared/AblyAuth.cs
AndyTWF: This actually turned out to be more complex - so deferring this to a separate PR

src/IO.Ably.Shared/Realtime/Workflows/RealtimeWorkflow.cs
AndyTWF: Deliberate - we can only outrun the TTL via an unbroken chain of reauth-driven events with no connection failures intervening, any real failures would set SkipAttach = false. If we evaluate ShouldSuspend it emits SUSPENDED and then immediately CONNECTING, as the caller would have already queued the next command.

src/IO.Ably.Shared/Realtime/Workflows/RealtimeState.cs
AndyTWF: Already covered - there's a comment above explaining the rationale.

sacOO7 (approved):
LGTM

PR #1332, opened 2026-09-02
Update repository references for rename to ably-pubsub-dotnet

umair-ably:
**Do not merge until the repo rename happens, in the programme's rename freeze window.** Merging early would point `features.yml` at the `ably-sdk-builds-ably-pubsub-dotnet` IAM role, whose OIDC trust policy is bound to a repository name that does not exist yet — the credential step would fail and the feature-spec report upload would break (a failure mode that is otherwise easy to miss).

This is step 17 of the PDR-091b PubSub split plan for .NET, the sibling of [ably-ruby#457](https://github.com/ably/ably-ruby/pull/457). It only updates strings inside the repo; it does not rename anything.

## What changed

- **`.github/workflows/features.yml`** — `repository-name: ably-dotnet` → `ably-pubsub-dotnet`, with a comment explaining that this input selects the `ably-sdk-builds-ably-pubsub-dotnet` IAM role already provisioned by [ably/infrastructure#13005](https://github.com/ably/infrastructure/pull/13005). YAML validated.
- **`nuget/io.ably.nuspec`, `nuget/io.ably.push.android.nuspec`, `nuget/io.ably.push.ios.nuspec`** — `projectUrl`, `repository url`, `releaseNotes`. Package IDs untouched.
- **`README.md`** — license badge and its link, releases link, GitHub Issues link, the issue #446 link, and the `ably-pubsub-dotnet` SDK mention in the MAUI trimming note. The header image is already a repo-relative path (`images/NETSDK-github.png`), so it needs no change.
- **`CONTRIBUTING.md`** — clone URL and the `ably-dotnet\src\...` local paths, the codespaces link, the Actions/`package.yml`/tags links, and the `github_changelog_generator -u ably -p ably-pubsub-dotnet` invocation.
- **`PushNotifications.md`**, **`unity/README.md`**, **`unity/CONTRIBUTING.md`**, **`cake-build/README.md`**, **`cake-build/build.cake`** — repo links, clone URLs and local-path references.
- **`.devcontainer/devcontainer.json`** — the NuGet cache Docker volume name (`ably-dotnet-nuget-cache` → `ably-pubsub-dotnet-nuget-cache`); derived from the repo name rather than pointing at GitHub, harmless to rename (existing local volumes are just orphaned).
- **`src/IO.Ably.Shared/Agent.cs`** — the PR link in the doc comment only (line 61).
[...]

PR #1333, opened 2026-09-02
[Split 1/3] Restructure into Ably.PubSub.Core (assembly rename, drop Xamarin heads and push satellites, version 2.0.0)

umair-ably:
First of a stack of PRs applying [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) (PubSub device/server package split, major releases) to this repo, targeting the long-lived `integration/v2` branch. Reference implementation: [ably-js#2293](https://github.com/ably/ably-js/pull/2293); sibling port: [ably-java#1232](https://github.com/ably/ably-java/pull/1232).

## What this PR does

* **Assembly rename, namespace unchanged.** The core assembly becomes **`Ably.PubSub.Core.dll`** (`AssemblyName`, output dll, XML doc file). `namespace IO.Ably` and every public type name are untouched — the namespace question is bundled into the ⚠️091d pass so it is decided once across SDKs. The distinct assembly name is the load-bearing half: keeping `IO.Ably.dll` would let an app that transitively pulled both `ably.io` and the new core resolve with only an MSB3277 warning and fail at runtime; a distinct assembly name over a shared namespace turns mixing into loud CS0433 compile errors instead.
* **Layout** (pure `git mv`, so history follows; TFMs unchanged, `netstandard2.0;net6.0;net7.0` plus the `net462` head):

  | was | now |
  |-----|-----|
  | `src/IO.Ably.NETStandard20` | `src/Ably.PubSub.Core` |
  | `src/IO.Ably.NETFramework` | `src/Ably.PubSub.Core.NETFramework` |
  | `src/IO.Ably.Shared`, `src/IO.Ably.Shared.MsgPack` | `src/Ably.PubSub.Shared`, `src/Ably.PubSub.Shared.MsgPack` |
  | `src/IO.Ably.Tests.{Shared,DotNET,NETFramework}` | `src/Ably.PubSub.Tests.{Shared,DotNET,NETFramework}` |
  | `src/IO.Ably.sln`, `.Package.sln`, `.NetStandard.sln`, `.NetFramework.sln` | `src/Ably.PubSub.sln`, `.Package.sln`, `.NetStandard.sln`, `.NetFramework.sln` |

  Every project/solution/shared-project import path, `InternalsVisibleTo`, `AssemblyTitle`, Cake path and test-assembly glob follows. `IO.Ably.ruleset`, `IO.Ably.Tests.ruleset`, `stylecop.json` and `IO.Ably.snk` keep their names — the strong-name key must stay byte-identical and renaming the rulesets buys nothing.
* **Legacy Xamarin heads dropped** (decision 5b): `IO.Ably.Android` (monoandroid) and `IO.Ably.iOS` (Xamarin.iOS) are gone, along with `IO.Ably.Xamarin.sln`, the ILRepack `_Package_Merge_All` task, the `Build.Xamarin` target, and the Android-SDK plus legacy-Xamarin-installer steps in `package.yml`. Classic Xamarin / .NET for Android+iOS / MAUI consumers receive the `netstandard2.0` asset.
* **Push satellites parked** (decision 6): `IO.Ably.Push.Android` / `IO.Ably.Push.iOS`, `nuget/io.ably.push.*.nuspec`, `IO.Ably.PackagePush.sln`, the `PushPackage` Cake target, `package-push.sh`/`.cmd` and the whole `package-push` job are removed, as are the `AndroidSample` and `DotnetPush` examples that target them (both live on in the 1.x maintenance line). Push **admin** is untouched — it stays in the core and is exposed by both doors until ⚠️091d.
* **Packaging**: `nuget/io.ably.nuspec` **deleted** (`ably.io` is never published from this branch again) and replaced by `nuget/ably.pubsub.core.nuspec` — `<id>Ably.PubSub.Core</id>`, same `$version$` token and authors/owners/license/icon/readme/repository metadata, dependency groups for `netstandard2.0`/`net6.0`/`net7.0`/`net46`, and a `<description>` that states plainly that this is an internal implementation package for Ably's own `Ably.PubSub.Device` / `Ably.PubSub.Server` packages and is not intended for direct use. `_Package_Create_NuGet` now packs a **list** of nuspecs so PR 2 can add the door nuspecs to it.
* **Version 2.0.0** in `src/CommonAssemblyInfo.cs` (the single source read back by `Defaults.GetVersion()`) and `unity/Assets/Ably/version.txt`; copyright year to 2026.
* **Unity**: `Update.AblyUnity`, `link.xml`, the three test asmdefs' `precompiledReferences` and `unity/CONTRIBUTING.md` all now name `Ably.PubSub.Core.dll`; the `.unitypackage` is named `ably.pubsub.<version>`.
* **Docs, minimal**: a "2.0 is in development" note at the top of README describing the new package set and the never-mix rule, the Xamarin rows in "Supported platforms" replaced by ".NET for Android, .NET for iOS and MAUI via `netstandard2.0`", the MAUI `TrimmerRootAssembly` snippet fixed to `Ably.PubSub.Core`, plus `CONTRIBUTING.md`, `cake-build/README.md` and a note at the top of `PushNotifications.md`.

### Two real bugs the rename surfaced

Both were pre-existing footguns that the rename tripped, and both are fixed here in their own commits:

1. **`IoC` looked the platform head up by assembly name** (`Assembly.Load(new AssemblyName("IO.Ably"))`). After the rename that throws `FileNotFoundException`, which the static ctor swallows, leaving `Platform` null: the transport factory and `PlatformId` silently fall back (so the agent string loses its `dotnetN`/OS entries) and `IoC.MobileDevice` throws `NullReferenceException`, because only `NotImplementedException` is caught there. 18 `RestInitSpecs` tests failed on it. `Platform.cs` is compiled into the same assembly as the shared code by every head, so it now looks the type up in `typeof(IoC).Assembly` — correct, and immune to any later rename.
2. **`FindTestAssemblies` defaulted to the glob `IO.Ably.Tests.*.dll`.** The Cake task only *warns* when it finds no assemblies and then returns, so the two .NET Framework test legs would have gone green having run zero tests. Now `Ably.PubSub.Tests.*.dll`.

## What this PR deliberately does not do

* **No device/server packages** — `Ably.PubSub.Device` / `Ably.PubSub.Server`, the shared `Side` source and the agent assertions land in PR 2.
* **No agent identifier changes.** `Agent.AblySdkIdentifier` is still `ably-dotnet/<version>`; the family rename to `ably-pubsub-dotnet` and the versionless side flags land in PR 2, because [ably-common#361](https://github.com/ably/ably-common/pull/361) (the registry PR) is still open and no identifier ships unregistered. `Agent.cs` is otherwise only moved.
* **No release tooling** — the pre-flight, `publish.yml`, lockstep ordering and trusted publishing are PR 3.
* **No repo rename.** `features.yml` keeps `repository-name: ably-dotnet` (the rename and its references PR are a separate, later step).
* **No API or type renames, and no namespace change** — ⚠️ PDR-091d is still in review.
* **`ably.io` 1.x is never published from this branch.** Fixes ship from a 1.x maintenance branch cut before the integration merge; `ably.io.push.android` / `ably.io.push.ios` 1.x likewise.

## Notes for reviewers

* Decisions **2** (assembly `Ably.PubSub.Core.dll`, namespace `IO.Ably` kept), **5** (core TFMs unchanged), **5b** (Xamarin heads dropped) and **6** (push satellites parked) are implemented exactly as recorded on 2026-09-02 — this PR does not reopen them.
* **The `net462` head is untested locally.** No Mono/MSBuild in the environment this branch was prepared in, so `Build.NetFramework` and the two `Test.NetFramework.*` legs were not run; CI's `run-tests-macos-mono` and `run-tests-windows-netframework` workflows cover them. The `FindTestAssemblies` fix above is what those legs depend on, so please check they report a real test count and not zero.
* **The Unity plugin dll could not be regenerated.** ILRepack needs Mono or Windows, so the stale `unity/Assets/Ably/Plugins/IO.Ably.dll` is *removed* rather than replaced: someone needs to run `./unity-plugins-updater.sh 2.0.0` (or the `.cmd`) on a machine with Mono and commit the resulting `Ably.PubSub.Core.dll`. The plugin's `.meta` was renamed in place, so the importer settings and asset GUID survive and the regenerated dll drops straight into it. Until then the Unity package is incomplete.
* **`nuget pack` could not be run locally** (needs `nuget.exe`/Mono). Instead every `files` entry was checked against the actual `Release` build output — `src/Ably.PubSub.Core/bin/Release/{netstandard2.0,net6.0,net7.0}/` each contain `Ably.PubSub.Core.{dll,pdb,xml,deps.json}` and `IO.Ably.DeltaCodec.{dll,pdb,xml}`, which the two globs per TFM match.
* **`IO.Ably.DeltaCodec.*` needed explicit `files` entries.** The old nuspec's `IO.Ably.*` glob happened to sweep the delta codec into each `lib/` folder; `Ably.PubSub.Core.*` does not, so the codec would have gone missing silently. Worth a second pair of eyes.
* **The net46 assets are no longer ILRepack-merged**, since `_Package_Merge_All` is gone with the Xamarin heads. `Newtonsoft.Json` is therefore a declared `net46` dependency rather than a merged-in one, and `IO.Ably.DeltaCodec.dll` ships alongside as it already did for the netstandard assets. This is a real change in shape for `net46` consumers and the main thing I would want confirmed before GA.
* **The ILRepack helper in `cake-build/helpers/tools.cake` is kept** even though the plan says to remove it with `_Package_Merge_All`: `Update.AblyUnity` still needs it to merge the Unity plugin assembly, and the Unity package steps are explicitly in scope. Deliberate deviation.
* **Pre-existing, not introduced here:** the `net8.0` and `net9.0` legs of the three `Test.NetStandard.*` workflows run **zero** tests. `Ably.PubSub.Tests.DotNET` targets `net6.0;net7.0`, so `dotnet test --framework net8.0` matches nothing and the leg passes vacuously. Same on `main` today. Worth its own issue; the plan's step 5 already flags the related `Platform.cs` runtime detection gap (.NET 8/9/10 apps report `dotnet-standard`).
* `src/IO.Ably/CustomSerialisers/GeneratedSerializers/**` is dead: three generated MsgPack serializers referenced by no project. Consciously left alone.
* `Import_RootNamespace` in the renamed `.projitems` follows the shared-project names (e.g. `Ably.PubSub.Shared`). It is a Visual-Studio-only hint for new-file namespaces and does not affect compilation; the code namespace is unchanged.
* `git grep` finds no dangling `IO.Ably.NETStandard20` / `IO.Ably.NETFramework` / `IO.Ably.Tests.*` / `IO.Ably.dll` / `io.ably.nuspec` / `PackagePush` / `IO.Ably.Android` / `IO.Ably.iOS` / `IO.Ably.Push.*` / `Build.Xamarin` / `package-push` references outside `lib/`, `common/` and `CHANGELOG.md`. The remaining `ably.io.push.*` mentions are in `PushNotifications.md`, which now opens with a note that it documents the 1.x packages.

## Verification

Run on macOS with .NET SDKs 6.0.428 / 7.0.410 / 8.0.424 / 9.0.317, via the same Cake targets CI uses:

| Command | Result |
|---------|--------|
| `dotnet build src/Ably.PubSub.NetStandard.sln -c Release` | succeeded, 0 warnings, 0 errors (all three TFMs) |
| `./build.sh --target=Test.NetStandard.Unit.WithRetry --framework=net6.0` | **Test Run Successful** — 1112 total, 1093 passed, 19 skipped, 0 failed |
| `./build.sh --target=Test.NetStandard.Unit.WithRetry --framework=net7.0` | **Test Run Successful** — 1112 total, 1093 passed, 19 skipped, 0 failed |
| `./build.sh --target=Test.NetStandard.Unit.WithRetry --framework=net8.0` | ran 0 tests — the pre-existing vacuous leg described above, not a regression |
| `Build.NetFramework`, `Test.NetFramework.*` | **not run** — no Mono/MSBuild locally; CI covers them |
| `./build.sh --target=Package --version=2.0.0` | **not run** — needs `nuget.exe`/Mono; nuspec `files` entries verified against real build output instead |
| Integration tests (sandbox) | **not run** |


maratal (changes requested):
**Verdict: the shape is right, requesting a few changes before merge.** Renaming the artefact while keeping the namespace, bumping to 2.0.0, minimal docs, no CHANGELOG edits, and no agent or door changes is exactly the right scope for the first PR of the stack. The two bugs the rename surfaced are real and correctly fixed. But one packaging bug leaked in from PR 2, the new nuspec is born with stale URLs, and the integration branch itself is well behind `main`.

### Must fix

- **The Unity package task can never succeed on this branch.** The guard added in the last commit checks for `unity/Assets/Ably/Plugins/Ably.PubSub.Device.dll`, but everything else in this PR produces and references `Ably.PubSub.Core.dll` (the merge task, `link.xml`, the three asmdefs, the renamed `.meta`). That is PR 2's name, where the Unity plugin becomes the Device assembly. As merged, `UnityPackage` and the `package-unity` CI job throw unconditionally. Fix the name here, and define the plugin file name once in `paths.cake` so the build task and the packaging guard cannot drift. Since PR 2 renames the plugin again anyway, consider skipping the intermediate Core rename in Unity and going straight to the final name, so the `.meta` is renamed in place once.
- **The new nuspec and README note use the old repo URL.** The repo was renamed on 9 September (#1332 merged to `main`), yet `nuget/ably.pubsub.core.nuspec` is a brand-new file pointing at `github.com/ably/ably-dotnet`. The PR description's claim that `features.yml` "keeps `repository-name: ably-dotnet`" is also stale: the base branch already says `ably-pubsub-dotnet`. New files should not inherit the pre-rename URL.
- **`integration/v2` is behind `main`.** It lacks #1331 (the RTN23 connection fixes), #1332 (rename references), #1338 (1.2.19) and #1339. Merge `main` into `integration/v2` before landing this PR. After 400+ `git mv`s every later sync needs rename-aware conflict resolution, so the cheapest moment is now.

### CI and verification

The Features check is red with `Not authorized to perform sts:AssumeRoleWithWebIdentity`. The pattern across the stack: every pull request into an `integration/*` base fails with `repository-name: ably-pubsub-dotnet`, pull requests into `main` pass, and the one green run in the stack (split-3 at b2c9e771) had flipped the value back to `ably-dotnet`. That points at the new IAM role's trust policy being scoped to base branch `main`, or the old role being looser. It is an infrastructure item (infrastructure#13005), not this PR's code, but we should get a no-op PR into `integration/v2` fully green before stacking on it, and PR 3 should not be reverting the repository name.

Things the description asked reviewers to confirm, which I checked:

- The net46 legs ran real tests, so the `FindTestAssemblies` glob fix works. Both legs also retried a handful of flaky failures and went green.

  | Leg | Unit | Integration |
  |---|---|---|
  | macOS Mono | 1113 | 244 |
  | Windows .NET Framework | 1113 | 244 |

- The `FindTestAssemblies` footgun remains, though: it still warns and returns green on zero assemblies. Make it throw.
- No dangling old-name references outside `lib/`, `common/` and `CHANGELOG.md`. The only hits are pre-existing `IO.Ably.Tests.Shared` namespaces.
- The Unity `.meta` is byte-identical after the rename, so the asset GUID survives.
- The door names in the new XML docs (`PubSubDevice.CreateClient`, `PubSubServer.CreateRealtimeClient`, `PubSubServer.CreateHttpClient`) match #1334's actual API.
- The `IoC` fix is correct: the platform type is compiled into the same assembly in every head, including the ILRepack-merged Unity dll.

### Smaller points

- **Public XML docs promise platform behaviour.** The constructor docs say an unclassified client "will be rejected once MAU-based pricing is live". PDR-091b §5 says that about the *old* package's constructors. For 2.x core constructors, which the doors themselves call, rejection could only be server-side. Verify the wording against the PDR or soften it to the classification and billing dependency. Same for the `PushNotifications.md` note that commits to a `net8.0-android`/`net8.0-ios` push port "post-GA"; decision 6 only says parked.
- **net46 package contents.** The `lib\net46` glob now reads `bin\Release` rather than the old ILRepack `Packaged\` folder, so it will also sweep in `Ably.PubSub.Core.dll.config` (MSBuild copies `app.config` as `<TargetFileName>.config` for libraries) and the pdbs. Harmless but noisy. The netstandard output was verified but not this one; list the net46 files explicitly or check them on the Windows Package run.
- **net46 is strong-named for the first time** under the Package configuration. Fine and consistent with the SDK-style head, but state it in the PR next to the "no longer ILRepack-merged" note, and watch the Package build for CS8002 against `IO.Ably.DeltaCodec` if that submodule is unsigned. I could not check it locally because the submodule is not checked out.
- **Package downgrade.** `System.Threading.Tasks.Extensions` goes 4.5.3 → 4.5.2 to match the net46 `packages.config`. Align upward by bumping the legacy head instead.
- **Root namespace hints.** `Import_RootNamespace` becomes `Ably.PubSub.Shared` and the test project's `RootNamespace` becomes `Ably.PubSub.Tests.NETFramework`. Visual Studio will propose those for new files while the code namespace is intentionally `IO.Ably`. Keep the old values.
- **A 1.6 MB `.trx` is in the history.** It was added in the IoC commit and removed in the last one. Squash-merge, or drop it from that commit.
- **Stack sequencing in durable docs.** README, `cake-build/README.md` and the `package.cake` comment say "added in the next PR in this stack". Acceptable now, but PR 2 must remove the phrasing; docs should carry facts that outlive the rollout.
- **Branch CI.** Workflows only trigger pushes on `main`, so nothing runs on `integration/v2` after merge. Adding `integration/**` to the push triggers is cheap, and PR 3 already touches every workflow.
- **Pre-existing, agree it needs its own issue:** the net8.0 and net9.0 test legs run zero tests because the test project only targets net6.0 and net7.0.

One more thing for PR 3: .NET has two version sites (`src/CommonAssemblyInfo.cs` and `unity/Assets/Ably/version.txt`). The release pre-flight should assert they agree.

maratal:
**Correction to my review above, on the Features check.** I said the base branch already had `repository-name: ably-pubsub-dotnet` and that the description was stale on that point. That was wrong: `integration/v2` and this branch both still say `ably-dotnet`, and `main` is the branch that says `ably-pubsub-dotnet` (from #1332). That value is exactly why Features fails here with `Not authorized to perform sts:AssumeRoleWithWebIdentity`: the old `ably-sdk-builds-ably-dotnet` role's trust policy no longer matches the renamed repository. Every run with `ably-dotnet` fails and every run with `ably-pubsub-dotnet` passes, including the one green run in this stack on #1335, which is the PR that flips it. So this is not an infrastructure item. Please change `features.yml` to `repository-name: ably-pubsub-dotnet` in this PR (or merge `main` into `integration/v2` first, which brings #1332 with it). The rest of the review stands.

src/Ably.PubSub.Core/Ably.PubSub.Core.csproj
```diff
@@ -51,8 +51,8 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="System.Threading.Channels" Version="4.5.0" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
+    <PackageReference Include="System.Threading.Channels" Version="4.6.0" />
```
sacOO7, line 54: Need to check if we really need to modify versions here

sacOO7:
Thanks @maratal — this is a precise read, and your Features self-correction is exactly right: the base branches say `ably-dotnet` and `main` (via #1332) is what carries `ably-pubsub-dotnet`, so the new value is the fix, not a regression. Dispositions:

- **`integration/v2` behind `main`** — agreed, and now done: we've merged `main` in (bringing #1331/#1332/#1338/#1339). It was a clean fast-forward with no conflicts (`integration/v2` had no commits of its own), and we synced the stack upward with merge commits (no rebases), so your SHA-pinned comments stayed anchored. This folded in the `features.yml` value via #1332, so the repo-name question is resolved. Features is now green on the merged head (`14c044dd`) — the AWS trust resolves against the renamed repo.
- **Unity guard vs the `Core`/`Device` name** — correct that #1333 in isolation is inconsistent (the guard is authored at the final `Ably.PubSub.Device.dll` name that #1334 introduces). We deliberately went straight to the final name as you suggested (the plugin has never been committed under a split name — zero migration), so it reconciles the moment #1334 lands on the integration branch. We'll also collapse the plugin path to a single source of truth in `paths.cake` so the build task and guard can't drift.
- **`features.yml`, nuspec URLs, `.trx`, "next PR in this stack" phrasing, version-site pre-flight, net46 test run** — all already addressed at `b2c9e771`: nuspecs point at `ably-pubsub-dotnet`, `.trx` is removed + gitignored, the durable-doc phrasing is gone, and `release.cake` asserts the two version files agree. We'll squash the final merge so the `.trx` blob never reaches `main`.
- **`FindTestAssemblies` throw-on-zero** — agreed, making it throw (follow-up commit on the integration branch).
- **Tasks.Extensions** — good catch; we're reversing our direction and aligning **up** (bumping the net46 head to 4.5.3) rather than down.
- **MAU wording, RootNamespace hints, net46 `lib\net46` contents, strong-naming note** — taking all of these: verifying the MAU claim against PDR-091b §5 (softening if it overstates for 2.x core constructors), restoring `IO.Ably` root-namespace hints, tightening the `lib\net46` file list, and calling out the first-time net46 strong-naming in the description.
- **net8/net9 zero-test legs** — agreed, own issue.

Fix commit: `86be74ee`; the `main` sync has since landed as merge `14c044dd`.

PR #1337, opened 2026-09-09
Provision each sandbox test environment once, atomically

AndyTWF:
Fixes an intermittent `NullReferenceException` that fails sandbox tests on CI, seen most recently as:

```
IO.Ably.Tests.Push.PushAdminSandboxTests+ChannelSubscriptionsTests
  .ShouldSuccessfullySetAndUpdateChannelSubscription(protocol: Json) [FAIL]
System.NullReferenceException : Object reference not set to an instance of an object.
  at IO.Ably.Tests.SandboxSpecs.GetRestClient (SandboxSpecs.cs:85)
```

### Why it happens

`AblySandboxFixture.GetSettings` caches provisioned app settings into a plain `Dictionary` from a static method, with an `await` between the `ContainsKey` check and the write:

```csharp
if (Settings.ContainsKey(environment)) { return Settings[environment]; }
Settings[environment] = await Initialise();
return Settings[environment];
```

`xunit.runner.json` sets `parallelizeTestCollections`, and the sandbox is reached from several collections concurrently — `SandBox Connection`, `AblyRest SandBox Collection`, `SandBox Collection`, `Presence Sandbox`, `Channel SandBox`, plus `ChannelSubscriptionsTests`, which declares no `[Collection]` and so gets an implicit one of its own. Those writes race.

A concurrently written `Dictionary` can be left with its indexer returning `null` for a key `ContainsKey` has just accepted. `SandboxSpecs.GetRestClient` then dereferences the settings it was handed, which is the exception above. The same window also allowed two collections to each provision an app.

### The change

A `Lazy<Task<…>>` over a `ConcurrentDictionary`, so `Initialise` runs once per environment however many callers race and all of them get the same task.

It also passes the environment through to `Initialise`, which the old code dropped — `Initialise` defaults to `"sandbox"`, so requesting any other environment provisioned a sandbox app and cached it under the other environment's name. Nothing requests a non-default environment today, so that half is a latent trap rather than a live bug, but it's a cheap thing to close while here.

### On verification

The exception is intermittent by nature and did not reproduce locally across two full sandbox runs, so this removes the race by construction rather than by a failing test. Those runs did confirm the change is inert otherwise — **5 failed / 230 passed / 12 skipped both with and without it**, four of the five being the same tests, all pre-existing flakes under parallel sandbox load.

### Discounted

Guarding with a lock or `SemaphoreSlim` would also work, but `Lazy` over `ConcurrentDictionary` needs no lock discipline at the call site and gives single-shot initialisation for free.

Found while investigating a CI failure on #1331, which turned out to be unrelated to that PR's changes — `AblySandboxFixture` is untouched there.

## Summary by CodeRabbit

* **Tests**
  * Improved the reliability of test environment setup during concurrent operations.
  * Ensured environment-specific settings are consistently applied when configuring sandbox test runs.


sacOO7 (approved):
LGTM

